PR #1, opened 2026-03-28
feat(mcp): whisper inject — pre-fetch context on initialize, smart get_context cache

YagizEfeGokce:
## What this does

Adds whisper inject support for the MCP adapter — giving Claude Desktop the same automatic context injection that Claude Code users get via `UserPromptSubmit` hooks.

**The problem:** ormah's existing whisper inject relies on Claude Code's hook system. There is no equivalent hook mechanism in Claude Desktop — only the MCP protocol. The MCP server never sees user messages; it only receives tool calls from the LLM. A different approach is required.

---

## Mechanism

Two cooperating layers replace the hook-based flow:

### 1. Eager pre-fetch on initialize

When Claude Desktop connects via MCP, `stdio_server` establishes the transport. `run_mcp_stdio` immediately schedules a background coroutine with `asyncio.create_task` — concurrent with the initialize handshake, before any tool call arrives:

```
[t=0ms]   MCP initialize fires
[t=0ms]   session_id = uuid4()
[t=0ms]   asyncio.create_task(_prefetch_context())  ← background
[t=~200ms] GET /agent/context → result written to _session_cache[session_id]
[t=N seconds] user types first message
[t=N+ms]  LLM calls get_context(task_hint="...") → served from cache
```

By the time the LLM makes its first tool call, the pre-fetch is already complete. Zero added latency on the first `get_context`.

### 2. Fast / slow path in `get_context` dispatch

| Condition | Path | Behaviour |
|-----------|------|-----------|
| No `task_hint`, cache populated | **Fast** | Returns pre-fetched result immediately — no HTTP call |
| `task_hint` provided | **Slow** | Calls `GET /agent/context?task_hint=…` — fresh, task-scoped context |
| No `task_hint`, cache miss or pre-fetch failed | **Slow** | Falls through to API — identical to previous behaviour |
| `session_id` is `None` (non-MCP callers) | **Slow** | Cache is never consulted — fully backward-compatible |

### 3. LLM instruction in tool description

Because the MCP server cannot intercept user messages, the only reliable trigger is the LLM itself. The `get_context` description now includes an explicit directive:

> *IMPORTANT: Call this tool at the very start of every conversation, before doing any other work. Pass the user's first message (or a concise summary of their task) as `task_hint`…*

The LLM reads this on `tools/list` and calls `get_context(task_hint=<user task>)` as its first action — extracting the task hint from conversation context, which is more accurate than any classifier.

---

## Session cache

```python
_session_cache: dict[str, dict] = {}

# One entry per active MCP process:
_session_cache["a1b2c3d4-…"] = {
    "prefetch_result": str | None,   # general context, no task_hint
    "prefetch_done":   bool,
    "prefetch_failed": bool,
    "task_result":     str | None,   # task-scoped context, populated on first slow-path call
}
```

`session_id` flows through the call chain via closure: `run_mcp_stdio → create_mcp_server(session_id) → call_tool closure → _dispatch(session_id)`. Each tool call resolves to the correct session without globals or thread-locals.

The entry is removed in a `finally` block when the stdio stream closes, preventing unbounded growth across reconnections.

---

## Race condition handling

When the MCP session closes, `run_mcp_stdio`'s `finally` block calls `_session_cache.pop(session_id)`. If `_prefetch_context` is still running (or being cancelled as the event loop shuts down), its own `finally` block must not assume the entry still exists. All writes inside `_prefetch_context` use `dict.get()` guards:

```python
finally:
    entry = _session_cache.get(session_id)
    if entry is not None:
        entry["prefetch_done"] = True
```

`CancelledError` is a `BaseException`, not `Exception`, so it bypasses the `except` block and reaches `finally` directly — this is the primary scenario requiring the guard.

---

## Files changed

| File | Change |
|------|--------|
| `src/ormah/adapters/mcp_adapter.py` | `import asyncio, uuid`; `_session_cache` dict; `_prefetch_context()` coroutine; `session_id` parameter on `create_mcp_server` and `_dispatch`; `get_context` fast/slow path; `run_mcp_stdio` rewritten with session lifecycle |
| `src/ormah/adapters/tool_schemas.py` | `get_context` description updated with session-start directive and improved `task_hint` description |

No backend changes. No new API endpoints. No new tools.

---

## Compatibility

- All 14 other tools in `_dispatch` are unchanged.
- `session_id=None` (any caller that doesn't go through `run_mcp_stdio`) produces identical behaviour to the previous version — the cache is never consulted.
- The existing Claude Code `UserPromptSubmit` whisper flow is unaffected.

### DIFF
--- a/src/ormah/adapters/mcp_adapter.py
+++ b/src/ormah/adapters/mcp_adapter.py
@@ -49,7 +49,9 @@ async def _prefetch_context(
             resp = await client.get("/agent/context", params=params)
             if resp.is_success:
                 text = resp.json()["text"]
-                _session_cache[session_id]["prefetch_result"] = text
+                entry = _session_cache.get(session_id)
+                if entry is not None:
+                    entry["prefetch_result"] = text
                 logger.info(
                     "Whisper pre-fetch done for session %s: %d chars",
                     session_id[:8],
@@ -61,14 +63,20 @@ async def _prefetch_context(
                     session_id[:8],
                     resp.status_code,
                 )
-                _session_cache[session_id]["prefetch_failed"] = True
+                entry = _session_cache.get(session_id)
+                if entry is not None:
+                    entry["prefetch_failed"] = True
     except Exception as exc:
         logger.warning(
             "Whisper pre-fetch error for session %s: %s", session_id[:8], exc
         )
-        _session_cache[session_id]["prefetch_failed"] = True
+        entry = _session_cache.get(session_id)
+        if entry is not None:
+            entry["prefetch_failed"] = True
     finally:
-        _session_cache[session_id]["prefetch_done"] = True
+        entry = _session_cache.get(session_id)
+        if entry is not None:
+            entry["prefetch_done"] = True
 
 
 def create_mcp_server(
--- a/src/ormah/adapters/mcp_adapter.py
+++ b/src/ormah/adapters/mcp_adapter.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
+import uuid
 
 import httpx
 from mcp.server import Server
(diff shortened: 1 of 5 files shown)


rishikesh-jentic:
@YagizEfeGokce

This is great stuff. I had missed this, will look into this soon. 
Thanks for the contribution. :raised_hands:

r-spade:
Thanks again for taking this on, and especially for surfacing the Claude Desktop gap. This is a real use case we do want to support.

The main architectural issue is that `get_context` was removed intentionally, not accidentally. We folded the old startup-context behavior into two paths: `get_self` for explicit identity/profile reads, and `whisper` for contextual recall during the session. Because of that, bringing `get_context` back is not the direction I think we should merge.

I think there’s also a practical issue with the current approach: the MCP prefetch/cache would likely go stale across the session, since it snapshots context at startup and doesn’t refresh after memory writes or updates.

This branch would also need a rebase onto current `main`, since it now conflicts with newer MCP adapter changes on our side.

I opened a follow-up PR that tackles the same Claude Desktop problem in the current whisper-first architecture: https://github.com/r-spade/ormah/pull/2

If you’re up for it, I’d genuinely love your review there, since your PR is what highlighted the gap in the first place.

PR #3, opened 2026-04-09
feat(improvements): conflict surfacing + gate tuning + decay alerts + versioning + session summary

YagizEfeGokce:
Five improvements from the `feat/improvements` branch — **791 tests pass**.

---

## 1. Surface tension conflicts as queryable observation nodes

Tension conflicts create a queryable `[observation]` node tagged `conflict` and `needs-review`, surfaced in whisper and linked to both conflicting memories.

---

## 2. Feedback-driven injection gate tuning

Each `submit_feedback` call nudges the global whisper injection gate via a proportional learning rule, persisted in `gate_state`. Accepting whispers raises it (more selective); rejecting lowers it (wider window).

New config: `ORMAH_GATE_TUNING_ENABLED` (true), `ORMAH_GATE_MIN` (0.30), `ORMAH_GATE_MAX` (0.80), `ORMAH_GATE_LEARNING_RATE` (0.02)

---

## 3. Proactive decay alerts before archival demotion

The decay manager writes `decay_alert_log` rows when core/working nodes drop below `decay_alert_threshold` (R < 0.40). First whisper of each session surfaces up to 5 unacknowledged alerts. `recall_node()` auto-acknowledges.

New config: `ORMAH_DECAY_ALERT_ENABLED` (true), `ORMAH_DECAY_ALERT_THRESHOLD` (0.40), `ORMAH_DECAY_ALERT_REFIRE_DAYS` (7)

---

## 4. Memory versioning — field-by-field diffs and recall_history

`update_node` stores both old and new snapshots in `audit_log`. New `recall_history()` formats a human-readable changelog with `field: old → new` diffs. Exposed via `GET /agent/recall/{id}/history` and the `recall_history` MCP tool.

---

## 5. Session summary hook

After `session_watcher` ingests a transcript, it creates a concise `[event]` memory with: space, turn count, memory type breakdown, and top extracted titles. Tagged `session-summary` for easy recall and whisper surfacing. Does not require LLM — built from the already-extracted nodes.

New config: `ORMAH_SESSION_SUMMARY_ENABLED` (true), `ORMAH_SESSION_SUMMARY_MIN_TURNS` (3)

Also fixes `ingest_conversation` to include `type` in returned dicts so breakdown counts are accurate.

---

## Test plan

```bash
uv run pytest tests/   # 791 tests pass
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Before each hooked prompt, Ormah:
 4. Reranks candidates for precision
 5. Avoids repeating recently-whispered context when the topic has not shifted
 6. Applies a relevance gate so silence beats noise
+7. Adjusts that gate over time based on your feedback — accepted whispers raise it, rejected ones lower it
 
 The result is a compact block of context added before the model sees your message.
 
@@ -418,6 +419,12 @@ ORMAH_SIMILARITY_THRESHOLD=0.4
 ORMAH_WHISPER_MAX_NODES=6
 ORMAH_WHISPER_INJECTION_GATE=0.50
 ORMAH_WHISPER_RERANKER_ENABLED=true
+
+# Feedback-driven gate tuning (adjusts injection gate based on submit_feedback signals)
+ORMAH_GATE_TUNING_ENABLED=true
+ORMAH_GATE_MIN=0.30
+ORMAH_GATE_MAX=0.80
+ORMAH_GATE_LEARNING_RATE=0.02
 
 # FSRS decay
 ORMAH_FSRS_INITIAL_STABILITY=1.0
--- a/README.md
+++ b/README.md
@@ -193,13 +193,40 @@ Ormah does not just collect memories. It keeps the graph healthy.
 Background jobs:
 
 - link related memories
-- detect contradictions and belief evolution
+- detect contradictions and belief evolution — surfaced as queryable `[observation]` nodes
 - merge near-duplicates
 - score importance from access, centrality, and recency
 - decay stale memories with FSRS-style retrievability
 - consolidate overlapping working memories
 - assign spaces to orphaned memories
 - refresh indexes incrementally
+
+### Conflict surfacing
+
+When the conflict detector finds a genuine **tension** between two beliefs — both simultaneously held and irreconcilable — it no longer just adds a silent `contradicts` edge. It creates a new `[observation]` node that makes the conflict visible and actionable.
+
+The observation node:
+
+- Is tagged `conflict` and `needs-review` so it surfaces in whisper when related topics come up
+- Includes the explanation of why the two memories contradict each other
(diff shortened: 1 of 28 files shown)


r-spade:
Great thinking here ! 
Thanks for putting this together. There are a lot of thoughtful ideas in here, and I appreciate the ambition behind the PR.

  A few directions I especially like:
  - surfacing genuinely actionable contradictions
  - memory history / `recall_history`
  - session-level summarisation as a way to compress past work

  That said, I think this PR is trying to do too many things at once, which makes it harder to reason about the product and UX implications of each change.

  A few high-level thoughts that I have:

  - I like the contradiction direction, especially for hard project-level inconsistencies like “this project uses Postgres” vs “this project uses SQLite” for the same role. Those feel actionable and worth surfacing.
  - I’m less convinced that all “tensions” should become visible `needs-review` memories. Some contradictions are healthy or nuanced rather than something to resolve, and I don’t want Ormah to become too “in the user’s face.”
  I think being able to hold opposing ideas in your head is a sign of a healthy thinking brain. I could be wrong though. 
  example: , a user might genuinely want Ormah to be proactive and helpful, while also wanting it to stay quiet and non-intrusive unless it’s very confident. That’s a healthy tension to balance, not necessarily something that should
surface as a `needs-review` contradiction.
  - I’m nervous about the global whisper gate tuning. If feedback moves the gate too broadly, it feels like we could end up surfacing more unwanted memories rather than fewer. But happy to be pushed back if you think this is the right move
  - I’m also unsure about decay alerts as a user-facing whisper feature. Internally it makes sense to track fading memories, but I’m not sure the memory system should announce its own decay state to the user unless they explicitly ask.
  - `recall_history` feels like a strong addition because it’s useful when needed without being intrusive.

  Overall: I think there is great work here
  Note: I’d be more comfortable if this were split into smaller pieces so we can evaluate each feature on its own merits. Easier to review and merge smaller chunks too. 
  
  All of what I say here are up for debate and I welcome healthy push back and debate on these. I would be really interested in knowing your thoughts on this.

YagizEfeGokce:
Thank you for the insights. I will create a detailed version of each feature along with clear explanations. I will also keep in mind that Ormah should function as a background system rather than something that presents itself to the user and takes their time. Instead, Ormah should give time back, not take it.

r-spade:
> Thank you for the insights. I will create a detailed version of each feature along with clear explanations. I will also keep in mind that Ormah should function as a background system rather than something that presents itself to the user and takes their time. Instead, Ormah should give time back, not take it.

Ormah should fade away into the background. Never surface itself to the user unless absolutely necessary.
This is an ideology that I have been trying to maintain throughout Ormah.

PR #10, opened 2026-04-15
refactor: align features with fade-into-background vision

YagizEfeGokce:
## Summary

This PR removes three features that conflict with Ormah's core design principle — **"Ormah should fade into the background"** — and fixes two bugs discovered during the review. Net result: **12 files changed, +111 / -405 lines**, all 925 tests pass.

---

### 1. Conflict observation nodes

**Before:** When the conflict detector found a genuine tension between two beliefs, it created a `[observation]` node tagged `needs-review` that surfaced in whisper and demanded user action. This made Ormah announce itself and created actionable to-do items — contradicting the idea that it should operate silently.

**After:** Tension conflicts are recorded as `contradicts` edges only. No observation node is created. The contradiction is still detectable via `recall_history` and graph traversal, but it no longer interrupts the user.

| Aspect | Before | After |
|---|---|---|
| Conflict visibility | `[observation]` node with `needs-review` tag | Silent `contradicts` edge |
| User action required | Yes — must resolve via `mark_outdated` | No — edge is informational |
| Whisper surfacing | Yes — `needs-review` tag pulls it into context | No — only discoverable via `recall_history` |

**Files:** `src/ormah/background/conflict_detector.py`, `README.md`

---

### 2. Feedback-driven gate tuning

**Before:** Each `submit_feedback` call nudged a global `injection_gate` value stored in a `gate_state` DB table. Positive signals raised the gate (more selective), negative signals lowered it (more permissive). This single global knob ignored per-topic context — a signal about one memory's relevance shouldn't change the threshold for all future whispers.

**After:** Feedback adjusts per-node affinity only (the existing affinity boost mechanism). The injection gate stays at its static config value (`ORMAH_WHISPER_INJECTION_GATE`, default 0.50). Removed `get_gate()`, `_update_gate()`, `gate_state` table, and all related config fields.

| Aspect | Before | After |
|---|---|---|
| Gate adjustment | Global, feedback-driven, persisted in DB | Static config value only |
| Config fields | `gate_tuning_enabled`, `gate_min`, `gate_max`, `gate_learning_rate` | Removed |
| DB table | `gate_state` | Removed |
| Tests | `TestGateTuning` (7 tests) | Removed |

**Files:** `src/ormah/engine/memory_engine.py`, `src/ormah/config.py`, `src/ormah/index/db.py`, `src/ormah/index/schema.sql`, `tests/test_engine/test_submit_feedback.py`, `docs/09 - Affinity and Feedback.md`, `README.md`

---

### 3. Decay alert whisper injection

**Before:** On the first message of each session, `context_builder` injected a `## Ormah: memory fading` block listing nodes with low retrievability. This violated the whisper principle by making Ormah visible to the user and framing internal maintenance as a user-facing concern.

**After:** Decay alert logging still runs internally (the `decay_alert_log` table and `run_decay` logic are preserved) — it's just no longer surfaced to the user. Internal logging continues to work for debugging and metrics.

| Aspect | Before | After |
|---|---|---|
| User-visible output | `## Ormah: memory fading` block in whisper | None |
| Internal logging | `decay_alert_log` table | Same — unchanged |
| Config flag | `decay_alert_enabled` (default `True`) | Removed — always runs |
| Tests | `TestDecayAlertWhisper` (2 tests) + disabled-flag test | Removed (3 tests) |

**Files:** `src/ormah/engine/context_builder.py`, `src/ormah/background/decay_manager.py`, `src/ormah/config.py`, `tests/test_engine/test_decay_alerts.py`

---

### 4. mark_outdated now captures snapshots in recall_history

**Before:** `mark_outdated()` wrote an audit log entry with only `{ "reason": "...", "old_valid_until": "..." }` — no snapshot of the node's state before or after the change. `recall_history` rendered this as just "Reason: X" with no before/after context, making it impossible to see what actually changed.

**After:** `mark_outdated()` captures `old_snapshot` (full node before modification) and stores `changed_fields` + `new_snapshot` in the audit detail. `recall_history` now renders field-level diffs for `mark_outdated` entries the same way it does for `update` entries.

**Files:** `src/ormah/engine/memory_engine.py`

---

### 5. Write buffer thread safety

**Before:** `WriteBuffer.append()` wrote to the buffer file without any locking. During a `drain()`, `rewrite()` would read existing entries, then replace the file — but a concurrent `append()` between the read and replace was silently lost.

**After:** Added `threading.Lock` to `WriteBuffer`. `append`, `clear`, and `rewrite` all acquire the lock, preventing concurrent appends from being lost during a drain.

**Files:** `src/ormah/store/write_buffer.py`

---

### 6. session_summary_min_turns raised from 3 to 5

**Before:** `min_turns=3` created `[event]` memories from trivially short sessions, adding noise to the working tier.

**After:** `min_turns=5` ensures only sessions with meaningful interaction produce summaries.

**Files:** `src/ormah/config.py`

---

## Test plan

- [x] All 925 existing tests pass
- [x] Removed 12 tests for deleted features (7 gate tuning + 3 decay alert whisper + 2 conflict observation)
- [x] `mark_outdated` snapshot capture verified via `recall_history` output
- [x] `WriteBuffer` thread safety verified — concurrent `append` during `rewrite` no longer loses entries
- [x] No `gate_state` table in schema migration path (removed from both `schema.sql` and `db.py`)


r-spade (commented):
Thanks for this contribution! There's real engineering effort here and several of these features address genuine gaps: offline buffering, audit history, and session summaries are all things I've wanted. Happy to review it properly.

Before diving into the code, one thing is making this hard to review: the PR description and the actual diff are describing two completely different changes. The description says this removes three features and nets -405 lines, but the diff shows 26 files changed and +2,469 additions. I suspect the description was written for an earlier version of this branch and didn't get updated. Could you rewrite it to describe what's actually here? It doesn't need to be long, just accurate. Without that it's difficult to tell what's intentional and what might be a carry-over from something else.

A few things below ranging from an actual bug to design questions I'd love your thinking on.

r-spade:
**Bug in `promote_node` (routes_agent.py)**

When the core cap is enforced after a promotion, demoted nodes get saved to the file store and the index is rebuilt, but the `tier` column in the `nodes` table is never updated. The DB and file store end up out of sync.

`recall_promoter.py` handles this correctly with `UPDATE nodes SET tier = 'working', updated = ? WHERE id = ?` after each demotion. Could you add the same here?

r-spade:
**Bug in `_drain_write_buffer` (mcp_adapter.py)**

In the `ConnectError` handling inside the drain loop, `pending.index(entry)` returns the position of the **first** occurrence of that entry in the list. If two buffered calls happen to be structurally identical, this re-queues from the wrong position and silently drops one. Since the whole point of the buffer is durability, it's worth fixing:

```python
for i, entry in enumerate(pending):
    ...
    except httpx.ConnectError:
        failed.append(entry)
        failed.extend(pending[i + 1:])
        break
```

r-spade:
**Double `file_store.load()` per node (recall_promoter.py)**

Small one: in the core-cap enforcement block each node is loaded from disk twice, once for the `if` check and once for the list value. A walrus operator or a plain loop with a local variable would halve the disk reads, which matters if the core tier is large.

r-spade:
**Design question: `update_memory` and `connect_memories` moving to agent tools (tool_schemas.py)**

This is the change I'd most like to understand the thinking behind. Previously `update_memory` and `connect_memories` were admin-only tools, so AI agents couldn't call them via MCP. Moving them into `TOOLS` means agents can now self-modify stored memories (content, tier, tags, space) and self-link them.

That's a meaningful trust boundary change. Is it intentional? If so, a short note in the PR explaining the reasoning would help a lot, something like "agents should be able to correct their own mistakes inline" or "the audit log gives enough traceability that this is safe." I just want to make sure it wasn't an accidental grouping change rather than a deliberate design decision.

r-spade:
**Design question: contradiction threshold may be catching duplicates, not contradictions (config.py)**

The pre-storage contradiction check uses a default cosine similarity threshold of 0.88. At that level you're mostly catching near-identical text. Actual contradictions ("I prefer dark mode" vs "I don't like dark mode") are semantically related but typically score around 0.60 to 0.75, not 0.88+. What tends to clear 0.88 is near-duplicate content, and the duplicate-merger job already handles that.

Was the intent here to catch potential duplicates as a quick heads-up, or genuine contradictions? If it's the latter, a lower threshold with a note about the higher false-positive rate might serve better. Happy to discuss what would work well.

r-spade:
**Design question: checkpoint hook runs two Python processes per tool call (setup.py)**

The PostToolUse hook runs two `python3 -c` subprocesses on every single tool call to parse the JSON input, regardless of whether a checkpoint ends up firing. That's a fixed overhead on every Edit, Write, and Bash call the agent makes.

Would a simpler version work just as well? For instance, a single Python invocation that parses the input, checks the tool name, and decides whether to emit the reminder. Or even simpler: always emit the reminder for Write and Edit calls unconditionally. Not a blocker, but hooks run synchronously before the agent continues so it's worth keeping them lean.

r-spade:
**Design question: growing the toolset from 6 to 10 agent tools (tool_schemas.py)**

This PR grows the agent-facing MCP toolset from 6 tools to 10 in one shot. Each tool added is cognitive overhead the agent carries on every call, since the full schema is visible at inference time.

The case for `recall_history` and `promote_memory` isn't immediately obvious to me. `recall_history` feels like a debugging or admin tool rather than something an agent needs during normal operation. And `promote_memory` overlaps with what `update_memory` can already do by setting `tier=core`. Could you walk through the concrete scenarios that motivated these? If they're exploratory, it might be safer to start them in `ADMIN_TOOLS` and move them to agent-facing once there's evidence they're needed, rather than the other way around.

Worth thinking about all four new agent tools together as a group, since the decision on `update_memory` and `connect_memories` is related.

PR #17, opened 2026-06-01
feat(ui): galaxy graph layout with collision, color-by-space, clickable legend

### LINKED ISSUE #16: Graph view: zoom-out cap, overlapping dense clusters, label haze, no legend
## Summary

The graph view (`ui/src/components/GraphView.tsx`) has several interaction and readability problems on large memory graphs.

### 1. Cannot zoom out far enough to see the whole graph
`minZoom: 0.15` is a hard zoom-out floor. On large graphs the viewport never fits the full graph — scrolling/`cy.fit()` stops while most nodes are still off-screen.

### 2. Dense clusters read like a flat image on zoom-in
The single-pass force layout lets nodes pile up on the same coordinates. Zooming into the dense core just scales the pile up — it never reveals the individual nodes, because they physically overlap. (Cytoscape zoom is pure affine scaling and never re-runs layout, so the only way to make zoom-in reveal nodes is to guarantee they don't overlap in the first place — like Obsidian's graph view, whose live force simulation keeps a collision radius around every node.)

### 3. Spaces are scattered too far apart
Same-space nodes belong to the same subject and should read as one "galaxy", but the old ring-placement (`clusterRadius = totalGroups * 200`) spread clusters across a huge canvas, making the graph feel sparse and disconnected.

### 4. Labels dominate the dots
Node labels render at all zoom levels, producing a text haze that hides the dots and the connections between them at medium/low zoom.

### 5. No legend / no way to focus
There is no in-graph legend explaining the colors (link types vs spaces), and no way to highlight one space or one link type while dimming the rest.

## Goal

A graph that reads like "a bit of Obsidian + a bit of the original Ormah": clean colored dots sized by connectivity, clear connection lines, labels only on hover/selection, spaces visually separated as galaxies, orphans on their galaxy's rim, and an in-graph clickable legend for focus.

AndreLFSMartins:
Closes #16.

Redesigns the memory graph view (`ui/src/components/GraphView.tsx`) for readability and interaction on large graphs.

### Zoom-out cap
- `minZoom` lowered `0.15 → 0.01` so `cy.fit()` can frame the whole graph instead of stopping with most nodes off-screen.

### Galaxy layout (spaces as galaxies)
- One **invisible hub node per space** + `fcose` gravity/clustering groups each space into a distinct galaxy. Orphan nodes get a longer hub edge so they settle on the galaxy's outer rim.
- Hubs are ordered by cross-space connectivity (most-connected toward the centre) via a phyllotaxis (sunflower) seed.

### Obsidian-style zoom (the "it looks like a PNG on zoom-in" fix)
- Cytoscape zoom is pure affine scaling and never re-runs layout, so the only way zoom-in can reveal individual nodes is to guarantee they don't overlap. `fcose` is a single-pass layout that allows overlap, so the dense core piled nodes onto the same coordinates.
- Fix: **chain a `cytoscape-cola` pass with `avoidOverlap: true`** (collision, Obsidian-style) seeded from `fcose`'s positions (`randomize: false`, `handleDisconnected: false`). It only pushes overlapping dots apart by their own radius, preserving the galaxy structure.
- Measured result: **0 overlapping pairs**, ~25px min edge-to-edge gap across ~1.7k nodes.

### Two-channel node encoding: space colour + tier shape
- **Fill = space colour** (the project a memory belongs to) — matches the legend and reads the same across users.
- **Tier = shape, never a hue**: archived = hollow ring (dark fill + space-coloured rim), core = solid + white rim, working = solid. Because tier is shape-only it can never collide with a space colour, whatever palette the user picks for their tiers.
- `self` / `identity` keep the space fill but take a teal rim that overrides the tier rim, so "who I am" still reads without painting ~60% of the graph one colour.
- This replaces the earlier approach where the user's tier palette was painted onto the fill — it collided with the space hue and made the legend lie (archived nodes rendered in the custom archival colour, e.g. a wall of red, instead of their space colour).

### Dots, labels, links
- Nodes are **sized by connection degree**.
- Labels render **only on hover/selection** (LOD) instead of an always-on text haze.
- Cross-space edges are aggregated into one gold **"galaxy-link" per connected space pair** (width/opacity scaled by edge count) instead of hundreds of crossing lines. Brightness was toned down and a dedicated rule dims it correctly when another element is focused.

### Floating clickable legend
- In-graph legend (bottom-right): **TIERS / LINKS / SPACES** sections (English headings), each row clickable.
- Click a space, tier, or link type to **highlight matching elements and dim the rest**; click again to clear.
- The legend now sits above the cytoscape canvas (`z-index`) — without it the canvas swallowed every legend click, so space filtering was silently dead.

### Code cleanup (same PR)
- Removed dead `computeClusteredPositions`.
- Merged the two duplicate node-degree maps and fused the `crossDeg`/`pairCount` passes into one loop over cross-space edges.
- Memoized legend data (`edgeLegend`, `spaceLegend`) and extracted a `LegendRow` helper to drop duplicated JSX.
- Removed now-unused `displayNodeSize`/`nodeSize`; dropped a redundant `nodeSize` reset in the appearance effect that reverted degree-based sizing on theme change.
- Centralised the node-style derivation in a single `galaxyNodeStyle()` helper consumed by both node creation and the appearance effect, so the two paths can't drift.
- Cached `spaceColor` (it was re-hashing the space name ~2× per node) and extracted the repeated legend section-title style.

## Verification
- `tsc --noEmit` clean; `vite build` clean.
- Behaviour verified live against the dev server (node overlap count, dim/focus class application, deep-zoom rendering, legend click filtering for space and tier, fill/rim per tier).

### DIFF
--- a/ui/src/components/GraphView.tsx
+++ b/ui/src/components/GraphView.tsx
@@ -2,6 +2,7 @@ import {
   forwardRef,
   useEffect,
   useImperativeHandle,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -129,6 +130,25 @@ function nodeLabel(n: MemoryNode): string {
 
 const OVERVIEW_PADDING = 150;
 const SPACE_FOCUS_PADDING = 110;
+const ZOOM_MIN = 0.03;
+const ZOOM_MAX = 4;
+const ZOOM_SLIDER_MAX = 100;
+const ZOOM_SLIDER_STEP = 8;
+
+function clampZoomSliderValue(value: number): number {
+  return Math.max(0, Math.min(ZOOM_SLIDER_MAX, value));
+}
+
+function zoomToSliderValue(zoom: number): number {
+  const normalized =
+    Math.log(Math.max(ZOOM_MIN, Math.min(ZOOM_MAX, zoom)) / ZOOM_MIN) /
+    Math.log(ZOOM_MAX / ZOOM_MIN);
+  return Math.round(normalized * ZOOM_SLIDER_MAX);
+}
+
+function sliderValueToZoom(value: number): number {
+  return ZOOM_MIN * Math.pow(ZOOM_MAX / ZOOM_MIN, clampZoomSliderValue(value) / ZOOM_SLIDER_MAX);
+}
 
 function buildStyles(appearance: GraphAppearance) {
   const tokens = GRAPH_THEME_TOKENS[appearance.theme];
@@ -145,6 +165,7 @@ function buildStyles(appearance: GraphAppearance) {
         height: "data(nodeSize)",
         label: "data(labelText)",
         "font-size": `${Math.round(12 * GRAPH_DISPLAY_SCALE)}px`,
+        "min-zoomed-font-size": 8,
         color: "data(labelColor)",
         "text-outline-color": "data(labelOutlineColor)",
         "text-outline-opacity": 0.9,
@@ -263,7 +284,7 @@ function computeClusteredPositions(nodes: MemoryNode[]): Map<string, { x: number
   const spaceList = Array.from(spaceGroups.keys()).sort();
   const hasUngrouped = ungrouped.length > 0;
   const totalGroups = spaceList.length + (hasUngrouped ? 1 : 0);
-  const clusterRadius = Math.max(600, totalGroups * 200);
+  const clusterRadius = Math.max(600, Math.sqrt(totalGroups) * 200);
   const positions = new Map<string, { x: number; y: number }>();
 
   function placeGroup(group: string[], centroidAngle: number) {
@@ -304,6 +325,78 @@ const LEGEND_SECTION_TITLE_STYLE: CSSProperties = {
   letterSpacing: 1,
 };
 
+const RIGHT_RAIL_STYLE: CSSProperties = {
+  position: "absolute",
+  right: 12,
+  top: 56,
+  bottom: 12,
+  zIndex: 10,
+  display: "flex",
+  alignItems: "flex-end",
+  gap: 8,
+  minHeight: 0,
+  maxWidth: "calc(100% - 24px)",
+};
+
+const LEGEND_PANEL_STYLE: CSSProperties = {
+  display: "flex",
+  flexDirection: "column",
+  minHeight: 0,
+  maxHeight: "100%",
+  overflow: "hidden",
+  background: "rgba(12,14,18,0.85)",
+  border: "1px solid rgba(255,255,255,0.1)",
+  borderRadius: 8,
+  padding: 10,
+  fontFamily: "monospace",
+  fontSize: 11,
+  color: "#cdd6e0",
+  lineHeight: 1.7,
+  width: 230,
+  maxWidth: "calc(100vw - 78px)",
+};
+
+const SPACE_LEGEND_LIST_STYLE: CSSProperties = {
+  minHeight: 0,
+  overflowY: "auto",
+  paddingRight: 3,
+};
+
+const ZOOM_CONTROL_STYLE: CSSProperties = {
+  display: "flex",
+  flexDirection: "column",
+  alignItems: "center",
+  gap: 8,
+  width: 34,
+  padding: "8px 0",
+  background: "rgba(12,14,18,0.82)",
+  border: "1px solid rgba(255,255,255,0.1)",
+  borderRadius: 8,
+  boxShadow: "0 12px 30px rgba(0,0,0,0.35)",
+};
+
+const ZOOM_BUTTON_STYLE: CSSProperties = {
+  width: 22,
+  height: 22,
+  border: "1px solid rgba(255,255,255,0.14)",
+  borderRadius: 6,
+  background: "rgba(255,255,255,0.04)",
+  color: "#d8dee6",
+  fontFamily: "monospace",
+  fontSize: 14,
+  lineHeight: "18px",
+  cursor: "pointer",
+};
+
+const ZOOM_RANGE_STYLE: CSSProperties = {
+  width: 22,
+  height: 122,
+  writingMode: "vertical-lr",
+  direction: "rtl",
+  accentColor: "#d4a574",
+  cursor: "pointer",
+};
+
 // One clickable legend row: swatch + content as children, dimmed when another
 // row holds the focus. Shared by both the link-type and space sections.
 function LegendRow({
@@ -334,6 +427,18 @@ const GraphView = forwardRef<{ focusNode: (id: string) => void }, Props>(
     onNodeSelectRef.current = onNodeSelect;
     const [layoutReady, setLayoutReady] = useState(false);
     const [legendFocus, setLegendFocus] = useState<{ kind: string; val: string } | null>(null);
+    const [zoomSliderValue, setZoomSliderValue] = useState(() => zoomToSliderValue(0.18));
+
+    const identityNodeIds = useMemo(() => {
+      const ids = new Set<string>();
+      if (!userNodeId) return ids;
+      for (const e of edges) {
+        if (e.edge_type === "defines" && e.source_id === userNodeId) {
+          ids.add(e.target_id);
+        }
+      }
+      return ids;
+    }, [edges, userNodeId]);
 
 
     useImperativeHandle(ref, () => ({
@@ -396,20 +501,11 @@ const GraphView = forwardRef<{ focusNode: (id: string) => void }, Props>(
       },
     }));
 
-    useEffect(() => {
+    useLayoutEffect(() => {
       if (!containerRef.current) return;
       setLayoutReady(false);
 
       const nodeIds = new Set(nodes.map((n) => n.id));
-
-      const identityNodeIds = new Set<string>();
-      if (userNodeId) {
-        for (const e of edges) {
-          if (e.edge_type === "defines" && e.source_id === userNodeId) {
-            identityNodeIds.add(e.target_id);
-          }
-        }
-      }
 
       function selfRole(nodeId: string): string {
         if (nodeId === userNodeId) return "self";
@@ -476,10 +572,15 @@ const GraphView = forwardRef<{ focusNode: (id: string) => void }, Props>(
         elements: [...nodeElements, ...edgeElements],
         style: buildStyles(appearance),
         layout: { name: "preset" },
-        minZoom: 0.03,
-        maxZoom: 4,
+        minZoom: ZOOM_MIN,
+        maxZoom: ZOOM_MAX,
         wheelSensitivity: 0.3,
       });
+
+      const syncZoomSlider = () => {
+        setZoomSliderValue(zoomToSliderValue(cy.zoom()));
+      };
+      cy.on("zoom", syncZoomSlider);
 
       if (positions) {
         cy.nodes().forEach((node) => {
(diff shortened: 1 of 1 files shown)


r-spade:
Thanks again for the contribution here. This PR pointed out something important: the graph was not giving enough of the bigger picture. On larger memory graphs, the user needed a better way to see the whole shape first, then inspect the important clusters.

I pushed a follow-up commit, `ae1cdf4`, that keeps that direction while fitting the current graph behavior and performance constraints:

- removed the extra fCoSE + hidden hub layout pass, which was expensive on a large real graph
- kept the existing organic cola layout
- kept tier colors as the node color language, so the existing core/working/archival color settings stay meaningful
- restored access-count-based node sizing and always-visible labels
- starts the graph farther zoomed out so the full memory map is visible by default
- keeps the legend, but makes it useful for inspection:
  - tier and link rows highlight matching items
  - space rows now navigate/focus that cluster
  - clicking the active space again returns to the full overview
  - clicking legend rows clears any selected graph node first

I think this keeps the key insight from the PR: make the graph more interpretable at a glance, without adding a separate layout mode or making large graphs slower.

I would love to know what you think of this direction, especially whether the overview-first behavior and space navigation preserve the intent you were aiming for.

AndreLFSMartins:
Thanks for taking the time to rework this directly on the branch — I went through `ae1cdf4` and I'm on board with the direction. Dropping the fCoSE + hidden-hub pass for a single cola run is the right call on cost and complexity, and keeping `avoidOverlap: true` preserves the part I cared most about: zoom-in actually resolving into individual nodes. The legend changes are a genuine improvement too — space rows navigating to the cluster is more useful than my highlight-only version, and the stale-focus cleanup is a nice touch.

On tier colors as the color language: fair enough. It's the existing user-facing setting, and I agree the space palette shouldn't silently override what users already configured.

Two small things I'd flag before this merges, both in the spirit of the same perf/readability goals:

1. **Always-on labels at overview zoom.** With labels restored unconditionally, the default zoomed-out view renders every label (~1.7k on my graph) as an unreadable haze — which is problem 4 from #16 — and label rendering is also the most expensive part of cytoscape's draw path, so it works against the perf motivation for removing the second layout pass. There's a one-line middle ground: `"min-zoomed-font-size": 8` (or similar) on the node style keeps labels always-visible whenever they're legible and hides them only when zoom makes them unreadable anyway. Happy to push that if you agree.

2. **Orphans + `handleDisconnected: true`.** Orphan nodes are singleton components, so cola's disconnected-component packing may pull them out of their space cluster even though `computeClusteredPositions` seeds them inside it (in the previous approach the hub edges held them on the galaxy rim). Orphans are common in memory graphs, so this could quietly undo `clusterBySpace` for a meaningful fraction of nodes. Worth a quick check on a real large graph — if they do scatter, `handleDisconnected: false` combined with the preset seed might be all it takes.

Otherwise this looks good to me — overview-first plus space navigation absolutely preserves what I was aiming for.

AndreLFSMartins:
Follow-up with data — I ran `ae1cdf4` (production build, fresh profile, `clusterBySpace` on by default) against my real graph: **~3.5k nodes / ~8.3k edges, 35+ spaces**, and compared it with the PR head on the same data and viewport. Three findings, one confirmation:

**1. The layout doesn't converge at this scale.** `cola` is capped at `maxSimulationTime: 1800` (wall-clock), but `computeClusteredPositions` seeds the clusters on a ring of radius `totalGroups * 200` — with 35+ spaces that's a ~7000-unit ring, and 1.8s isn't remotely enough to contract it. The result (reproducible across 3 runs, dev and prod builds): grid-like artifacts where clusters never untangled, and the graph spread so wide that `cy.fit` clamps at the `minZoom 0.03` floor with most of the viewport empty — so the "overview-first" goal breaks exactly on the graphs that need it. Screenshots 1–2 attached (PR head fits the same graph at zoom 0.28; `ae1cdf4` hits the 0.03 floor).

**2. Orphans get pulled out of their space clusters.** Of 221 orphan nodes, **129 (58%) end up farther than 2× the cluster RMS radius from their space centroid** — `handleDisconnected: true` packs singleton components away from where the preset seeded them, silently undoing `clusterBySpace` for a big chunk of nodes.

**3. Label haze at mid zoom.** At zoom 0.35, labels render at ~4px — an unreadable smudge under every node (screenshot 3; label text replaced with lorem for privacy, density is real). A one-liner fixes the band without losing always-visible labels: `"min-zoomed-font-size": 8`.

**What held up:** `avoidOverlap` — 0 overlapping node pairs out of 3.5k, so the zoom-in fix survives the rewrite. 👍

Concrete suggestions, keeping your direction (single cola pass, tier colors, organic overview):

- scale `maxSimulationTime` with node count, and/or shrink the seed ring (e.g. `Math.sqrt(totalGroups)`-based radius) so cola starts near its equilibrium;
- `handleDisconnected: false` when `clusterBySpace` is on (the preset seed already places orphans sensibly);
- add `"min-zoomed-font-size": 8` to the node style.

Happy to push any of these as a follow-up commit.

r-spade:
> Follow-up with data — I ran `ae1cdf4` (production build, fresh profile, `clusterBySpace` on by default) against my real graph: **~3.5k nodes / ~8.3k edges, 35+ spaces**, and compared it with the PR head on the same data and viewport. Three findings, one confirmation:
> 
> **1. The layout doesn't converge at this scale.** `cola` is capped at `maxSimulationTime: 1800` (wall-clock), but `computeClusteredPositions` seeds the clusters on a ring of radius `totalGroups * 200` — with 35+ spaces that's a ~7000-unit ring, and 1.8s isn't remotely enough to contract it. The result (reproducible across 3 runs, dev and prod builds): grid-like artifacts where clusters never untangled, and the graph spread so wide that `cy.fit` clamps at the `minZoom 0.03` floor with most of the viewport empty — so the "overview-first" goal breaks exactly on the graphs that need it. Screenshots 1–2 attached (PR head fits the same graph at zoom 0.28; `ae1cdf4` hits the 0.03 floor).
> 
> **2. Orphans get pulled out of their space clusters.** Of 221 orphan nodes, **129 (58%) end up farther than 2× the cluster RMS radius from their space centroid** — `handleDisconnected: true` packs singleton components away from where the preset seeded them, silently undoing `clusterBySpace` for a big chunk of nodes.
> 
> **3. Label haze at mid zoom.** At zoom 0.35, labels render at ~4px — an unreadable smudge under every node (screenshot 3; label text replaced with lorem for privacy, density is real). A one-liner fixes the band without losing always-visible labels: `"min-zoomed-font-size": 8`.
> 
> **What held up:** `avoidOverlap` — 0 overlapping node pairs out of 3.5k, so the zoom-in fix survives the rewrite. 👍
> 
> Concrete suggestions, keeping your direction (single cola pass, tier colors, organic overview):
> 
> * scale `maxSimulationTime` with node count, and/or shrink the seed ring (e.g. `Math.sqrt(totalGroups)`-based radius) so cola starts near its equilibrium;
> * `handleDisconnected: false` when `clusterBySpace` is on (the preset seed already places orphans sensibly);
> * add `"min-zoomed-font-size": 8` to the node style.
> 
> Happy to push any of these as a follow-up commit.

@AndreLFSMartins Thanks a ton for checking it and reporting the issues. I had not tested it on a graph that was more than 2k in size. 
Would be great if you can push a follow up commit to fix the issue. I can verify the results and hopefully get the PR merged in later today or tomorrow and issue a new release. 

Also, your feedback and suggestions have been very insightful. Really do appreciate it.

AndreLFSMartins:
Thanks! I pushed on this and have an update — including a finding that changes the plan.

**Two of the three are fixed and verified** on my real graph:
- **Label haze** → `min-zoomed-font-size: 8` on the node style. Labels stay always-visible while legible, hidden only once zoom shrinks them below readability.
- **Orphans** → `handleDisconnected: false` when `clusterBySpace` is on. Orphans-outside-their-cluster dropped from **58% → 1%**.

**The convergence one (#1) is deeper than the tweaks I suggested.** I tested them on my graph (now grown to **5559 nodes / 14k edges / 70 spaces**) and the seed-ring + `maxSimulationTime` changes **don't move it** — fit still clamps at the 0.03 floor, span ~88k, `separationRatio` 0.12 (clusters fully interpenetrated). Crucially, at **25s** of simulation the result is identical, so it isn't a time problem. I also tried **fcose as a single pass** — it would cluster, but its synchronous compute **freezes the main thread >45s** at this scale.

The mechanical cause: no *global* force-directed layout works at 5.5k clustered nodes. cola runs incrementally but dissolves the space clusters (nodes with few intra-space edges have nothing holding them in their space); fcose clusters but freezes. The hidden-hub pass we dropped wasn't only a perf cost — it was *what held clusters cohesive*.

**Proposed fix — and it does NOT bring back the second fCoSE pass:** make the layout tractable by partitioning. Grouping by space already splits 5.5k nodes into ~70 sub-graphs of ~80. So: deterministic **circle-packing of the spaces** (macro) + an isolated small **cola per space** (micro, run headless). Each piece resolves instantly, the compact macro keeps `fit` off the floor, `avoidOverlap` is preserved. Overview-first survives; tier colors and the legend stay as you have them.

Before I push the rewrite: does that direction work for you? Happy to land the two verified fixes first as a small commit if you'd rather review them separately.

r-spade:
Thanks again for pushing this PR forward. It pointed out something important: the graph needed a much better big-picture view to be useful and interpretable.

I pushed a follow-up set of changes to this branch. The current version keeps the graph closer to Ormah's existing visual language while adding the main usability pieces from this discussion:

- wider default overview / zoom-out range
- reduced label haze at overview zoom
- right-side zoom control
- legend focus for tiers, link types, spaces, and self/identity nodes
- scrollable space legend for larger corpora
- tier colors remain the primary node-color encoding for now

While testing larger synthetic graphs around 3.5K nodes, we also found a separate performance issue: Cytoscape element creation plus full-graph layout gets expensive when the entire historical graph is loaded at once. I opened #22 to track that separately, likely as a follow-up PR with a level-of-detail / active-graph-first approach rather than expanding this PR further.

Would love to know what you think of this direction and any feedback you have on the UX and visual tradeoffs.

r-spade:
This is really helpful @AndreLFSMartins , thank you for digging into it with a real 5.5k graph.

Also, quick note: I missed seeing your earlier comment before pushing the latest changes to the branch, so my previous update was not directly responding to this. Sorry about that.

I’ve also been working on this on and off today and pushed a fix that addresses the initial issues: label readability, disconnected/orphan handling, zoom/legend/focus improvements, scrollable spaces, and the overview controls. So I think this PR is now in a much better place for the original graph interpretability problem.

I agree with your mechanical diagnosis too. At 5.5k nodes / 70 spaces, a single global force layout is probably the wrong model. More simulation time is not going to fix it if the layout itself is dissolving the space structure.

The partitioned direction makes sense to me: macro placement for spaces, then small isolated layouts inside each space. I also think progressive rendering by space could fit well with that, so the graph can become useful quickly instead of making the browser ingest everything at once.

The main thing I want to be careful about is preserving Ormah’s visual language. An earlier packed/bubble direction became too artificial, so I’d like the macro/micro work to be evaluated both for performance and for whether it still feels like an organic memory graph.

I opened #22 for the larger 3.5k+ / 5.5k+ performance and level-of-detail work. My preference is to keep this PR focused on the graph interpretability fixes and handle the partitioned/progressive large-graph layout as a follow-up PR. That gives us room to test the approach properly without expanding this PR too far.

Would love your thoughts on that split.

r-spade (approved):

PR #19, opened 2026-06-03
Fix server freeze: run engine work off the event loop + thread-local DB connections

### LINKED ISSUE #18: Server freezes under load: sync engine work on the event loop + global DB lock → 30s MCP timeouts
## Summary

The server intermittently freezes and MCP tool calls fail with:

```
Error: request to Ormah server timed out after 30s while running tool 'remember'
```

The failures are frequent (correlated with indexing) and not specific to payload size or content.

## Root cause

Two compounding defects, confirmed by measurement against a running server:

1. **Synchronous work runs on the asyncio event loop.** Every FastAPI route is declared
   `async def` but calls fully synchronous, CPU/lock/IO-heavy engine methods directly on the
   event-loop thread (e.g. `api/routes_agent.py` → `engine.remember()`). Any wait inside
   (lock acquisition, fastembed encode, file IO) freezes the **entire** server, not just that
   request.

   *Evidence:* a single `recall` takes ~0.18s; **20 concurrent `recall`s take ~3.25s** wall-clock
   (≈ 20 × single) — perfectly serialized on one thread.

2. **A global DB write `RLock` is shared with background threads.** `index/db.py` guards writes
   with a process-wide `threading.RLock()`; nine APScheduler jobs run in worker threads
   (`index_updater` every 1 min). When a background job holds the lock for a long write batch,
   a foreground `remember` — running synchronously **on the event-loop thread** — parks in
   `_lock.acquire()`, freezing the loop. Every queued request then exceeds the 30s MCP client
   timeout (`adapters/mcp_adapter.py`).

   *Evidence:* during a long `remember`, a concurrent `recall` (a read, which does **not** take
   the write lock) also times out — only an event-loop block explains a lock-free read freezing.

Additionally, a single shared `sqlite3.Connection` is used by the event loop, all routes, and all
background threads — a latent thread-safety hazard masked by the GIL and low concurrency.

## Not the cause (ruled out)

- **Embedding model / encode speed:** `embedding_max_content_chars=512` caps encode input; steady
  remember = ~0.2s; a 56 KB payload on an idle server = ~0.2s.
- **Content size:** a 32 KB payload returns in ~0.2s. The occasional multi-second "hang" is backlog
  draining behind a coincidental background job, not the payload.

## Reproduction

With a populated store, fire a heavy `remember` (or wait for a background batch) and concurrently
hit `recall` — the reads stall and MCP calls hit the 30s timeout. Or just fire 20 concurrent
`recall`s and observe ~3.25s serialized wall-clock.

## Proposed fix

Move route work off the event loop (Starlette threadpool) and give each thread its own SQLite
connection (WAL read concurrency), plus chunk the long background write batches. PR follows.

AndreLFSMartins:
Fixes #18.

## Summary

The server froze under load and MCP `remember` calls timed out at 30s because synchronous,
lock/CPU-heavy engine work ran directly on the asyncio event loop, while a single shared SQLite
connection and a global write `RLock` were contended by nine background threads. One long
background write batch parked the event loop in `_lock.acquire()` and stalled every queued request.

This PR removes the freeze with three changes:

1. **Thread-local SQLite connections** (`index/db.py`). Each thread gets its own `sqlite3.Connection`
   (pragmas applied per connection, and **sqlite-vec loaded per connection** — a fresh connection
   without `vec0` cannot query `node_vectors`). WAL gives N concurrent readers + 1 writer; writes
   stay serialized by the existing process-wide `RLock`. The transaction depth counter is now
   thread-local, and the lock is acquired via `with self._lock:` so it can never leak.

2. **Routes off the event loop** (`api/routes_*.py`). Engine-bound routes become plain `def` so
   Starlette dispatches them to its threadpool; routes that must `await` (read the request body,
   websocket) stay `async def` and wrap the blocking engine call in `anyio.to_thread.run_sync`.
   A slow request no longer freezes the whole server.

3. **Chunked background writes** (`background/importance_scorer.py`, `background/auto_cluster.py`).
   These jobs wrapped an all-nodes UPDATE loop in a single transaction; they now commit in chunks of
   100 to bound how long the write lock is held, so a foreground `remember` write isn't stalled
   behind a full-store batch.

## Test plan

New regression tests:

- `tests/test_index/test_db_concurrency.py` — distinct connection per thread; vector search from a
  worker thread (the sqlite-vec landmine); read-during-write does not block.
- `tests/test_api/test_routes_concurrency.py` — engine-bound routes overlap in the threadpool
  (max concurrency ≥ 2, wall-clock ≪ N × single).
- `tests/test_background/test_chunked_writes.py` — the all-nodes write splits into bounded chunks.

Full suite: `python -m pytest tests/` — all green (4 unrelated pre-existing failures in
`test_config.py`/`test_setup.py` are not touched by this PR).

## Verified live (populated store)

- 20 concurrent `/admin/health`: **0.033s** (event loop free, fully parallel).
- 10 concurrent `remember`: **1.75s, all 200** — previously timed out at 30s.
- Freeze test: `recall` stays responsive (~1s) while a heavy `remember` runs — previously returned
  a client timeout (event-loop block).

## Notes / out of scope

- `recall` throughput is still GIL-bound on model inference (concurrent recalls serialize on the
  embedding encode). That is a CPU characteristic, not the freeze bug; true parallelism would need
  process-level inference workers.
- A separate pre-existing crash exists in `engine.memory_engine._spread_activation`
  (`float * None` when an activation input is `None`); not addressed here.

### DIFF
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -98,7 +98,7 @@ def _persist_backup_settings(backup_dir: Path, retention_count: int) -> None:
 
 
 @router.get("/health")
-async def health(request: Request):
+def health(request: Request):
     tracker = getattr(request.app.state, "job_tracker", None)
     result: dict = {"status": "ok"}
     if tracker is not None:
@@ -110,13 +110,13 @@ async def health(request: Request):
 
 
 @router.get("/stats")
-async def stats(request: Request):
+def stats(request: Request):
     engine = request.app.state.engine
     return engine.stats()
 
 
 @router.get("/maintenance-status")
-async def maintenance_status(request: Request):
+def maintenance_status(request: Request):
     manager = getattr(request.app.state, "maintenance_manager", None)
     if manager is None:
         return {"status": "idle"}
@@ -124,14 +124,14 @@ async def maintenance_status(request: Request):
 
 
 @router.get("/backup")
-async def backup_status(request: Request):
+def backup_status(request: Request):
     """Return local backup configuration and latest backup metadata."""
     settings, service = _backup_service_from_request(request)
     return _backup_status_payload(settings, service)
 
 
 @router.post("/backup/create")
-async def create_backup(request: Request):
+def create_backup(request: Request):
     """Create a manual local backup of source-of-truth memory files."""
     from ormah.backup import BackupError
 
@@ -148,7 +148,7 @@ async def create_backup(request: Request):
 
 
 @router.post("/backup/settings")
-async def update_backup_settings(body: BackupSettingsUpdate, request: Request):
+def update_backup_settings(body: BackupSettingsUpdate, request: Request):
     """Update local backup settings for this server and persist them to config."""
     from ormah.backup import service_from_settings
 
@@ -169,14 +169,14 @@ async def update_backup_settings(body: BackupSettingsUpdate, request: Request):
 
 
 @router.post("/rebuild")
-async def rebuild_index(request: Request):
+def rebuild_index(request: Request):
     engine = request.app.state.engine
     count = engine.rebuild_index()
     return {"status": "rebuilt", "nodes_indexed": count}
 
 
 @router.get("/tasks")
-async def list_tasks(request: Request):
+def list_tasks(request: Request):
     """List all registered background tasks and their next run time."""
     scheduler = getattr(request.app.state, "scheduler", None)
     if scheduler is None:
@@ -194,7 +194,7 @@ async def list_tasks(request: Request):
 
 
 @router.post("/tasks/{task_id}/pause")
-async def pause_task(task_id: str, request: Request):
+def pause_task(task_id: str, request: Request):
     """Pause a background task by ID."""
     scheduler = getattr(request.app.state, "scheduler", None)
     if scheduler is None:
@@ -207,7 +207,7 @@ async def pause_task(task_id: str, request: Request):
 
 
 @router.post("/tasks/{task_id}/resume")
-async def resume_task(task_id: str, request: Request):
+def resume_task(task_id: str, request: Request):
     """Resume a paused background task by ID."""
     scheduler = getattr(request.app.state, "scheduler", None)
     if scheduler is None:
@@ -220,7 +220,7 @@ async def resume_task(task_id: str, request: Request):
 
 
 @router.post("/tasks/pause-all")
-async def pause_all_tasks(request: Request):
+def pause_all_tasks(request: Request):
     """Pause all background tasks."""
     scheduler = getattr(request.app.state, "scheduler", None)
     if scheduler is None:
@@ -231,7 +231,7 @@ async def pause_all_tasks(request: Request):
 
 
 @router.post("/tasks/resume-all")
-async def resume_all_tasks(request: Request):
+def resume_all_tasks(request: Request):
     """Resume all background tasks."""
     scheduler = getattr(request.app.state, "scheduler", None)
     if scheduler is None:
@@ -242,7 +242,7 @@ async def resume_all_tasks(request: Request):
 
 
 @router.post("/tasks/{task_id}/run")
-async def run_task(task_id: str, request: Request):
+def run_task(task_id: str, request: Request):
     """Manually trigger a background task by ID."""
     engine = request.app.state.engine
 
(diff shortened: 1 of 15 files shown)


r-spade:
Hi @AndreLFSMartins, this is a really solid PR and the writeup on the issue is excellent. The thread-local connections, moving routes to the threadpool, and the chunked background writes all line up exactly with the root cause. I pulled it locally and the freeze is genuinely gone, recall stays responsive while a heavy remember runs.

One thing surfaced when I stress tested the threaded recall path, and I wanted to flag it for you rather than just patch over it. Now that the routes run in the threadpool, `recall_search` executes on several threads at once, but `GraphIndex` and `HybridSearch` still capture a single `sqlite3.Connection` at construction (`GraphIndex(self.db.conn)` and `self.conn = db.conn`) and reuse it across every thread. That is the same shared-connection hazard the issue describes, just in two spots the thread-local work did not quite reach.

Under concurrent recall it reproduces reliably within a second or two:

- `sqlite3.InterfaceError: bad parameter or other API misuse` (SQLITE_MISUSE)
- `DatabaseError: cannot rollback - no transaction is active`
- occasional corrupted rows (`IndexError: tuple index out of range`)

A single recall passes fine because the window is small and the embedding encode partly serializes the threads, so it only shows up once you push 8 to 12 concurrent recalls through `recall_search` (which `_spread_activation` then drives into the shared graph connection).

The fix is small and in the same spirit as the rest of the PR: have `GraphIndex` and `HybridSearch` hold the `Database` and resolve `.conn` per access through a property, so each thread reads through its own thread-local connection (matching what `VectorStore` already does). I put it on a branch on top of your head commit so you can pull it straight in if you like: `fix/pr19-thread-local-conns` (commit `9d265a2`). It also adds a concurrency regression test that fails on the current code and passes with the change.

Full suite is green (913 passing) merged with current main, no conflicts.

Would love your review on the branch when you get a chance. If the approach looks good to you, feel free to fold the commit into this PR (or I am happy to open it as a separate PR against your branch, whichever you prefer). Keen to hear your thoughts on whether the property approach feels right or if you would rather pass the `Database` through explicitly everywhere.

Really appreciate you taking this on, it is a genuinely valuable fix and the analysis made it easy to follow. Excited to get this one merged with you!

r-spade (approved):

AndreLFSMartins:
Thanks for catching this and for the clean writeup — pulled the branch, reviewed it, and the property approach is exactly right. Resolving `.conn` per access so each thread reads through its own thread-local connection mirrors what `VectorStore` already does, keeping the index layer consistent. Accepting either a `Database` or a raw `sqlite3.Connection` via the duck-typed resolver is the pragmatic call — it keeps the existing `GraphIndex(conn)` test sites working without a churny migration, and `isinstance` would only force the runtime import you deliberately avoided with `TYPE_CHECKING`. Happy to keep it as-is, so I folded your commit straight in (`c3c7bad`) — single PR now.

While verifying I noticed the branch had drifted behind `main`, so I merged it up. That pulls in #23 (the `_spread_activation` NULL weight/score guard for #20) — worth calling out because this PR moves `recall_search` into the threadpool, which makes `_spread_activation` run concurrently, so that guard now sits in a genuinely hot path. Also picked up #24.

Validation: your `test_recall_concurrency.py` deterministically fails pre-fix (`test_graph_conn_is_per_thread` — same connection id across threads) and passes with the change. Full suite green at 908 passing apart from the pre-existing `test_config.py`/`test_setup.py` failures and one flaky watchdog test (`test_hippocampus::test_new_file_triggers_ingestion`) that fails on `main` too.

One minor thing I'll leave as-is unless you feel strongly: `MemoryEngine._get_hybrid_search` is a check-then-set lazy init without a lock, so the first burst of concurrent recalls can construct `HybridSearch` twice. It's benign — `get_encoder` is `@lru_cache` so there's no double model load and both instances are thread-safe — just slightly wasteful on the first hit. Happy to add a lock if you'd prefer.

Appreciate the collaboration on this one!

r-spade:
Thanks Andre, this has been a pleasure. Good call folding the commit in and rebasing, and you are spot on that #23 carries a lot more weight now that `_spread_activation` runs concurrently.

On the lazy `_get_hybrid_search`: agreed, let's leave it lock-free. One small note for the record, `get_encoder` is actually a manual `_adapter_cache` dict rather than `lru_cache`, but it gets warmed at startup via `_warmup_embedder`, so even a double construct never double-loads the model. The cost is just a couple of throwaway objects on the first concurrent hit, which is not worth a lock. I opened #27 to track it properly so we can revisit if we ever want to harden it.

Good to merge from my side.

It has been genuinely great collaborating with you on this one. The whole back and forth was fun and encouraging, and I am really looking forward to building more of ormah together in the future. Thanks again!

Issue #28, opened 2026-06-14
feat(background): bounded forgetting — delete dead-weight `archival` nodes (decay-only accumulates noise indefinitely)


AndreLFSMartins:
# feat(background): bounded forgetting — delete dead-weight `archival` nodes (decay-only accumulates noise indefinitely)

## Summary

Decay is a one-way ratchet that never empties. `decay_manager` demotes stale `working`
nodes to `archival` (`src/ormah/background/decay_manager.py`), but nothing ever removes
them. `archival` is a graveyard that only grows. A demoted node still:

1. **counts toward total node count** — graph node count grows monotonically with usage
   (observed: a real store at ~7K nodes; the graph view already struggles, see #22);
2. **is scanned by every background job** — `auto_linker` is O(n²) over *all* nodes
   (see `UPSTREAM_ISSUE_auto_linker_quadratic.md`), so the cost of the graveyard scales
   with the graveyard, and that work contends with foreground `remember`/`recall`
   (the contention class fixed in #18 / PR #19);
3. **still surfaces in recall** — only down-ranked via `tier_boost_archival = -0.1`, so it
   adds noise rather than disappearing;
4. **is still loaded into the graph UI**.

So decay reduces *priority*, not *cost* or *noise*. Over a long-lived store, decay-only
guarantees an ever-growing tail of dead weight that keeps costing compute and polluting
results. **Deletion is the only lever that attacks the root cause (shrinking the store).**

This is complementary to #22, not a substitute: #22 ("active-graph-first") stops the UI from
*loading* archival; this issue stops the store from *keeping* dead-weight archival at all.
Both are needed — one bounds what the browser ingests, the other bounds what exists.

## Design constraint: deletion is irreversible, and memory lives on trust

A false negative — deleting something that later mattered — is unrecoverable and erodes trust
in the whole system. This drives the central rule:

> Deletion must never depend on a **single** signal. It acts only on the `archival` tail,
> requires the **conjunction** of several independent signals, and keeps a **reversibility
> window** before anything leaves disk.

### 1. Eligibility gates (delete only when ALL hold)

All fields already exist on `nodes` (`tier`, `importance`, `stability`, `last_review`,
`last_accessed`, `access_count`) plus edge degree and affinity/feedback:

1. `tier == archival` — never touches `working`/`core`. Deletion only cleans the graveyard
   decay already filled.
2. **Sustained staleness, not a point reading** — archived ≥ `deletion_min_archival_days`
[...]

PR #30, opened 2026-06-15
perf(background): incremental auto_linker via monotonic seq cursor (#26)

### LINKED ISSUE #26: perf(background): auto_linker is O(n²) — full-table scan × brute-force vector search × LLM-per-candidate every run
# perf(background): auto_linker is O(n²) — full-table scan × brute-force vector search × LLM-per-candidate every run

## Summary

`run_auto_linker` scans **every** node on each scheduled run and, per node, runs a brute-force vector
search (itself O(n)) plus an LLM classification call per surviving candidate. Total cost is **O(n²)**
in compute and **O(candidates) LLM calls** per run, repeated every `auto_link_interval_minutes`. As the
store grows, a single run goes from seconds to minutes/hours.

> Identified by code inspection, not yet measured against a large store.

## Evidence

`src/ormah/background/auto_linker.py` (`run_auto_linker`, around L288–L354):

```python
nodes = conn.execute("SELECT id, content, title, type, space FROM nodes").fetchall()  # O(n) — ALL nodes
...
for node in nodes:                      # O(n)
    query_vec = encoder.encode(text)
    similar = vec_store.search(query_vec, limit=6)   # O(n) brute-force (see vector-search issue)
    for match in similar:
        ...
        llm_result = _llm_classify_link(settings, node, other)   # network/LLM call per candidate
        _apply_edge(engine, node["id"], match["id"], relationship, reason, similarity)
```

`_find_link_candidates` (L131–L220) has the same shape and additionally does
`SELECT ... FROM nodes ORDER BY RANDOM()` (O(n log n) sort) before discarding most rows.

There **is** partial bounding already — `auto_link_max_edges_per_run` caps edges created and
`auto_link_checked` avoids re-checking pairs — but the **scan and per-node vector search still touch the
whole table every run**, so the dominant cost is unbounded in `n`.

## Why it grows

- Compute: O(n²) per run (n nodes × O(n) search).
- LLM calls: one per candidate above threshold — cost and wall-clock scale with the store.
- The work re-runs on a timer, so the system spends an increasing fraction of its time re-linking.

## Relationship to #18 / PR #19

PR #19 added **chunked writes** to `importance_scorer` and `auto_cluster` to bound write-lock hold time,
but did **not** touch `auto_linker`'s O(n²) read loop. Note also that `run_auto_linker` reads via the raw
`engine.db.conn` (bypassing the write `RLock`); PR #19's thread-local connections make that safe by
construction, but the read pattern is worth keeping in mind if the connection model changes again.

- **Incremental cursor:** only consider nodes created/updated since the last run (track a watermark on
  `created_at` / `updated_at`), turning O(n²) per run into O(Δ·n) where Δ = new/changed nodes.
- Optionally cap `similar` search to the incremental set's neighbourhood.
- Benefits compound once the vector search itself is ANN-backed (separate issue).

## Acceptance criteria

- [ ] Auto-linker processes only new/changed nodes per run (incremental watermark), not the full table.
- [ ] A run's cost is bounded by Δ, demonstrated with a test on a populated store.

AndreLFSMartins:
Closes #26.

## Problem
`run_auto_linker` re-scanned the **entire** node table every run (full table scan × brute-force vector search × LLM-per-candidate), so cost grew with total store size — ~70M distance ops + slow local-LLM calls per run on a real ~8k-node store. PR #19 chunked the writes but never touched the read loop.

## Fix
Drive the auto-linker (and the shared `_find_link_candidates` preview) by a **persisted monotonic change-sequence cursor** (`nodes.seq`), processing only nodes newer than the watermark — turning the per-run O(n²) full scan into O(batch·n), draining the historical backlog gradually without missing nodes from reindex/import or transient LLM failures.

Highlights:
- New monotonic `nodes.seq` column + backfill; bumped on every content write in the builder; durable `node_seq_next` counter (monotonic across `INSERT OR REPLACE`).
- Incremental seq-based node selection with an integer watermark; `auto_link_max_nodes_per_run` bound.
- Retry on LLM failure (classify invalid LLM output as `error`, not `none`, so it re-enters the delta).
- Watermark reset on `full_rebuild` so a rebuilt store re-drains correctly.
- Shared seq scan between the link job and the candidate preview.

## Review
Design + plan went through the Dev Council (v1→v3 revisions committed under `docs/superpowers/`). Tests green.


AndreLFSMartins:
Note: #31 (bounded forgetting / #28) is stacked on top of this branch — merge this one first, then #31's diff collapses to just its own changes.

r-spade:
Andre, I found a cursor edge case I wanted your take on before we merge. In `run_auto_linker`, if the vector index is empty or underfilled, `vec_store.search()` returns no matches, `node_resolved` stays true, and the seq watermark advances past those nodes without recording checked pairs or edges.

I reproduced it locally by clearing `node_vectors`, running the PR auto-linker, then re-embedding and running it again: the first run advanced `auto_link_watermark` to max seq with 0 LLM calls and 0 checked pairs; after re-embedding, the second run still skipped the pair because the watermark had already passed it. This can happen around `rebuild_index()`/`full_rebuild()` since vectors are deleted before `_reindex_all_embeddings()` restores them.

What do you think the right guard should be here? Options I see are gating auto-linker while embeddings are rebuilding, requiring vector coverage/current-node vector before advancing the cursor, or treating missing vectors as unresolved. I think we need a regression test for this path before merge.

docs/superpowers/plans/2026-06-15-auto-linker-incremental-cursor-design.html
r-spade, line 1: @AndreLFSMartins , I see that you are using superpowers plugin, I do like it too, but been tinkering with others as well. 
I believe we should remove the plans and spec from the repo. Can you remove them from the PR when you get a chance ?

r-spade:
> Andre, I found a cursor edge case I wanted your take on before we merge. In `run_auto_linker`, if the vector index is empty or underfilled, `vec_store.search()` returns no matches, `node_resolved` stays true, and the seq watermark advances past those nodes without recording checked pairs or edges.
> 
> I reproduced it locally by clearing `node_vectors`, running the PR auto-linker, then re-embedding and running it again: the first run advanced `auto_link_watermark` to max seq with 0 LLM calls and 0 checked pairs; after re-embedding, the second run still skipped the pair because the watermark had already passed it. This can happen around `rebuild_index()`/`full_rebuild()` since vectors are deleted before `_reindex_all_embeddings()` restores them.
> 
> What do you think the right guard should be here? Options I see are gating auto-linker while embeddings are rebuilding, requiring vector coverage/current-node vector before advancing the cursor, or treating missing vectors as unresolved. I think we need a regression test for this path before merge.

@AndreLFSMartins  I see that you have removed the super power docs, just wanted to leave a not here to get your input on this comment, happy to take it on if you want me to

AndreLFSMartins:
Good catch — confirmed and fixed in be9dd83.

**Root cause:** exactly the rebuild-window race you described. When `node_vectors` is empty/underfilled (`full_rebuild()` deletes vectors before `_reindex_all_embeddings()` restores them), `vec_store.search()` returns no candidates, the node loop leaves `node_resolved=True`, and the watermark advances past nodes that were never checked. Clearing the watermark in `full_rebuild` does not help — the run just rescans from 0 against an empty index and burns the cursor to max.

**Guard (your option 3, refined):** I went with *treat a node as unresolved when it has no vector of its own* — `vec_store.get(node["id"]) is None → node_resolved = False` — rather than a blanket "empty `similar` → unresolved". The distinction matters: a node that is embedded but genuinely has no neighbors above threshold is legitimately resolved and must advance, otherwise it becomes a poison node that stalls the cursor forever. Checking the node’s own vector cleanly separates "index not ready for this node" (transient → wait) from "no similar peers" (permanent → advance). Same fail-closed contract as the existing LLM-unavailable path.

I considered option 1 (gate while rebuilding) but it needs a global rebuild flag/lock and still misses incremental-embedding lag, so the per-node vector check is the smaller, more general fix.

**Known ceiling** (noted in a `ponytail:` comment): a node that *never* embeds would block the watermark indefinitely. Acceptable today since embedding failures are already logged; worth revisiting only if a permanently-unembeddable node ever stalls the cursor in practice.

**Regression test** (`test_empty_vector_index_does_not_advance_watermark`) reproduces your exact repro: clear `node_vectors` → run → watermark stays 0, 0 LLM calls, no edge; `_reindex_all_embeddings()` → run → the pair is finally classified and the watermark advances. Full `test_auto_linker.py` suite: 20 passed; `test_migration_seq.py`: 2 passed; ruff clean.

Took it on myself since the watermark design is mine — thanks for the offer though. Should be mergeable now, then #31 rebases cleanly on top.

(Re: the superpowers docs — yes, intentional; my local tooling writes `docs/superpowers/` and `.council/` which should never be versioned, so the PR `.gitignore` excludes them.)

r-spade:
Thanks Andre — I checked `be9dd83` locally and the per-node vector guard resolves the watermark issue. The focused auto-linker/migration suite passes (22 tests), and Ruff is clean.

I added `7e24151` on top of the PR to ignore `/docs/superpowers/`, so the generated planning/spec artifacts stay outside the repo going forward. No remaining blockers from my review.

r-spade (approved):
LGTM!

Issue #29, opened 2026-06-15
engine: mutators write file before the index lock — deletion/eligibility race (root fix for #28)


AndreLFSMartins:
## Summary

`MemoryEngine` mutators write the **markdown file before** acquiring the index write
transaction, and file operations are not covered by `Database.transaction()`'s lock:

- `update_node` — `src/ormah/engine/memory_engine.py:797-799` (`file_store.save(node)` then `builder.index_single(path)`)
- `_touch_access` (recall hot path) — `src/ormah/engine/memory_engine.py:1614-1618` (`file_store.save(node)` then the index `UPDATE` inside a transaction)

Because the file write happens outside any lock, any operation that reads node state and then
acts on the file races with a concurrent mutation. This was surfaced by 4 council rounds while
reviewing **#28 (bounded forgetting)**: even a guard that re-reads the source file inside
`BEGIN IMMEDIATE` and deletes file-first cannot fully close the window — a concurrent
recall/promotion can re-save the file between the guard's read and `soft_delete`'s move.

## Impact

For #28 the race is **accepted and documented** as a known limitation: deletion is opt-in
(OFF by default) and soft (reversible for `deletion_retention_days`), and the precondition (a
node untouched 90+ days being recalled in the exact microsecond of a daily job) makes it
astronomically improbable. But the underlying write-model issue is general — any "read-then-act
on file" path is exposed.

## Proposed root fix

Make eligibility-affecting mutators acquire `db.transaction()` **before** `file_store.save()`
(index/lock authoritative), so the `BEGIN IMMEDIATE` of any reader/deleter serializes against
them. Scope: `update_node`, `_touch_access`, and any other path that saves the file then writes
the index.

**Rejected alternative:** a global lock around `recall` — it reintroduces the #18/#19 contention
that PR #19 fixed.

## Notes

- This is the **root fix** deferred out of #28 by explicit decision; #28 ships with the race
  documented. See `docs/superpowers/specs/2026-06-15-bounded-forgetting-design.md` ("Known
  limitation").
- Care needed on the recall hot path: holding the transaction across the file write adds
  contention; measure before/after (the #18 concern, milder than a dedicated recall lock).

AndreLFSMartins:
Related: #27 (also a concurrency race exposed by concurrent recall after #19, but a different mechanism — `_get_hybrid_search` lazy double-construction, benign). This issue (#29) is the file-before-index-lock **write-ordering** race; distinct fix surface.

Issue #22, opened 2026-06-13
perf(ui): make graph view scale for large number of nodes


r-spade:
## Problem

The graph view becomes slow or temporarily unresponsive at larger memory sizes. This was first reproduced locally around the 3.5K node / 8.3K edge synthetic case, and Andre later tested the same class of issue on a real 5.5K node / 14K edge / 70-space graph in PR #17.

This is not primarily a SQLite/backend problem: the graph API returns data quickly. The bottleneck is on the frontend.

There are two stacked costs:

1. Cytoscape element creation is synchronous. Passing thousands of nodes plus thousands of edges into the Cytoscape constructor blocks the browser main thread while Cytoscape creates internal element/state structures, resolves styles, builds adjacency data, and initializes rendering.
2. Full-graph layout then adds another expensive pass. Shortening or deferring cola helps, but does not solve the initial synchronous element creation cost if Cytoscape still receives the full graph at once.

## Andre's Finding From PR #17

Credit to @AndreLFSMartins for digging into this with a real large graph and changing the shape of the problem.

On a 5.5K node / 14K edge / 70-space graph, the issue is not just that the simulation time is too short. A single global force-directed layout appears to be the wrong model at that scale:

- more cola simulation time did not fix convergence;
- cola can dissolve space structure because weakly connected nodes have little force keeping them in their space;
- fCoSE can cluster, but its synchronous compute can freeze the main thread for too long;
- the removed hidden-hub approach was expensive, but it was also providing cluster cohesion.

Andre's proposed direction is to make the layout tractable by partitioning:

- macro layout: deterministic placement of spaces;
- micro layout: isolated small layouts inside each space;
- progressive rendering by space so the graph becomes useful quickly instead of forcing the browser to ingest the whole corpus at once.

This should be explored carefully because prior packed/bubble-style experiments were technically faster but did not preserve Ormah's visual language.

## Product Direction

The large-graph view should avoid giving Cytoscape the entire historical graph by default.

A likely Ormah-native default is an active graph first:

- self / identity nodes
- core nodes
- working nodes

Archival nodes should remain reachable, but should appear on demand through search, node focus, or space drill-down rather than all being part of the initial overview.

This is preferable to a generic hard cap because the tier model already encodes graph importance and keeps the behavior explainable.

## Possible Technical Direction

Explore a combination of:

- active-graph-first initial payload/rendering;
- partitioned macro/micro layout by space;
- progressive rendering by space or importance tier;
[...]

PR #31, opened 2026-06-15
feat(background): bounded forgetting — delete dead-weight archival nodes (#28)

### LINKED ISSUE #28: feat(background): bounded forgetting — delete dead-weight `archival` nodes (decay-only accumulates noise indefinitely)

AndreLFSMartins:
Closes #28.

> **Stacked on #30 (#26).** This branch is cut from `perf/auto-linker-incremental`, so until #30 merges the diff here also shows #26's commits. Review/merge #30 first; the diff then collapses to the #28 changes (commits `34fde4e`..`706d29e`, all `(#28)`).

## What
Decay only demotes stale `working` nodes to `archival` — nothing ever removes them, so `archival` is a graveyard that only grows (real store: ~80% archival). This adds an **opt-in** background `forgetting_manager` that deletes dead-weight archival nodes through a conjunction of safety gates, with a reversible soft-delete → hard-purge lifecycle and a cap backstop.

## Safety model (deletion is irreversible — design centers on trust)
- **Opt-in, OFF by default** (`ORMAH_DELETION_ENABLED`); the job is a verified no-op until armed. All tuning is env-var driven (no UI needed).
- **Conjunction protections** (Phase A *and* the cap share one `_evaluate_protection`): never deletes the self node, a node with `archived_at IS NULL`, importance ≥ threshold, any positive feedback, or a hub (degree/strong-edge). Phase A additionally requires sustained staleness.
- **Atomic delete-if-eligible**: re-checks inside the `BEGIN IMMEDIATE` deletion txn; the guard reads volatile fields from the **source file** (not the lagging index); the file is moved **first**, then the index row removed.
- **Reversible**: soft-delete stamps `deleted_at` atomically into `deleted/`; hard-purge only after `deletion_retention_days`. New durable `nodes.archived_at` (frontmatter → index, survives `full_rebuild`); legacy files backfilled atomically once.

## Review
Design went through **4 Dev Council rounds** (revisions committed under `docs/superpowers/`): cap-vs-conjunction, TOCTOU, `archived_at` durability across rebuild, hybrid guard, atomic `soft_delete`, move-first ordering, lazy disk backfill. Final code review APPROVE_WITH_NITS (nits addressed). 54 feature tests green (7-gate conjunction matrix, cross-path race, durability, write-failure preserves file).

## Accepted limitation → #29
A microscopic deletion/mutation race remains because the engine writes the file before the index lock — accepted here (opt-in, reversible 30d, astronomically improbable) and tracked as the separate engine issue #29. See the spec's "Known limitation".

## Note on first run
On a real store the legacy backfill rewrites archival files once to stamp `archived_at`. At default thresholds on a young/dense store it may soft-delete 0 nodes (correct, conservative); lower `ORMAH_DELETION_MIN_ARCHIVAL_DAYS` to clean sooner.


AndreLFSMartins:
Pushed an additional commit wiring `forgetting_manager` into the sleep-cycle pass (`c9d54e1`).

Until now `forgetting_manager` was only registered on the in-process APScheduler timer (`forgetting_interval_hours`), which **restarts reset** — on a dev/laptop setup that restarts often, bounded forgetting effectively never fired, and the nightly `run-all` (sleep-cycle) skipped it entirely.

This commit adds it to `_TASK_RUNNERS` (manually runnable), `_TASK_DESCRIPTIONS`, and `_SLEEP_CYCLE_ORDER` — placed **after `decay_manager`** (so it acts on freshly-demoted archival) and **before `memory_backup`** (so the nightly snapshot reflects the post-forgetting state). No new flag: `run_forgetting` already self-gates on `deletion_enabled`, so when deletion is off the sleep-cycle entry is a safe no-op. Tests added in `tests/test_api/test_routes.py`.

AndreLFSMartins:
Force-pushed a cleanup commit: removed the 26 tracked `docs/superpowers/` files (local agent-tooling artifacts, not for upstream) from this branch and added `docs/superpowers/`, `.superpowers/`, `graphify-out/` to `.gitignore` so they can't creep back in — same fix you applied on #30 via `7e24151`. Diff vs `main` is now clean of those artifacts. No source changes.

r-spade:
Thanks @AndreLFSMartins . I re-reviewed this and I agree the core problem is real: decay lowers priority, but it does not bound the archival tail. Archival nodes still live in storage, search/index state, graph/admin payloads, and background-job scan paths, so a long-lived store keeps accumulating cost and noise.

I do not think this is ready to merge yet. Main findings:

**Branch shape: stale/conflicting diff**

The PR currently conflicts with current `main` and still includes stale duplicate work from already-landed PRs. That makes the diff much broader than bounded forgetting and makes merge/review risky. I think this needs a clean rebase/split so the remaining diff is only the #28 lifecycle.

**`deleted/` semantics: this affects more than forgetting**

Today `deleted/` is already used for merge tombstones and other soft-delete paths, not just forgetting. This PR stamps `deleted_at` in the shared `FileStore.soft_delete()` path, so future merge/manual-delete tombstones can also become purge candidates once deletion is enabled. That makes the feature broader than “delete dead archival nodes”.

**Recovery model: reversible on disk is not enough**

The PR adds internal `list_deleted()`/`purge()` helpers, but there is no user or agent surface for searching, recovering, or promoting cold memories. For a memory system, reversibility needs to be practical. I do not think users should inspect and restore tombstones one by one; that creates too much cognitive load.

**Cap behavior: the soft cap bypasses staleness**

The normal gate path is conservative: old archival age, no recent access, low retrievability, low importance, no positive feedback, and low connectivity. The `archival_soft_cap` path skips the sustained-staleness gates and can evict recently accessed/recently archived archival nodes to hit a count. That may be too aggressive, or at least it needs much clearer semantics.

**Docs/config: deletion needs explicit user-facing documentation**

Because this can eventually hard-purge memory files, the config and lifecycle need to be documented before merge: `ORMAH_DELETION_ENABLED`, retention, cap behavior, recovery expectations, and what happens to merge/manual-delete tombstones.

**Design direction: Deep Recall**

I think the natural product shape here is not “trash plus purge”; it is a cold memory layer with an explicit **Deep Recall** mode.

The model I have in mind:

- normal recall stays clean and searches active memory;
- bounded forgetting moves very low-value archival memories out of the hot path;
- those cold/tombstoned memories are excluded from normal recall and maintenance scans;
- when a user or agent explicitly needs old context, Ormah can run Deep Recall over that cold layer and surface a small set of useful old insights with provenance;
- restore/promotion happens only when a surfaced memory proves useful again.

That keeps the operational benefit of bounded forgetting without throwing away the long-tail value. Old memories are sparse and noisy, but occasionally they are exactly where the best insight is. Deep Recall lets Ormah dig there deliberately without making every normal recall pay for it.

I would be interested in your thoughts on shaping this PR toward that lifecycle: active memory → archival → cold/deep layer → optional explicit purge, rather than treating `deleted/` mainly as a purge queue.

The implementation work here is substantial, and the focused gate tests are strong. I ran the focused bounded-forgetting tests and the full PR-head suite locally, and they passed. The main work now is getting the lifecycle, UX, and merge shape right.

AndreLFSMartins:
Thanks for the thorough review — the objections all land, and I agree the branch shape needs to be fixed regardless.

I'm looking into the lifecycle direction more deeply before replying properly. Early digging suggests the archival tier itself may be the thing to re-examine first, which would change what bounded forgetting and Deep Recall should even be doing. I'd rather bring you data than a rushed answer.

Will come back with a considered response in the next few days.

r-spade:
> Thanks for the thorough review — the objections all land, and I agree the branch shape needs to be fixed regardless.
> 
> I'm looking into the lifecycle direction more deeply before replying properly. Early digging suggests the archival tier itself may be the thing to re-examine first, which would change what bounded forgetting and Deep Recall should even be doing. I'd rather bring you data than a rushed answer.
> 
> Will come back with a considered response in the next few days.

@AndreLFSMartins : Looking forward to your thoughts ! :raised_hands:

AndreLFSMartins:
@r-spade — here's the considered response I owed you.

Before answering I went digging into the archival tier, as promised, and what I found reshaped the plan: the lifecycle clock underneath decay *and* this PR's gates is effectively uncalibrated (details in §1 below — every gate keyed to retrievability or importance is currently selecting on noise). So the honest sequence starts with calibration, not with any forgetting policy.

On the design direction: I'm on board with Deep Recall, with one amendment from the digging — cold should be an **indexed state**, not a tombstone, because today's soft-delete removes the node from every index, so there would be nothing for a deep search to find. The proposal below decomposes your lifecycle into small PRs, addresses each of your five findings, and adds an honest erasure contract I found we need along the way.

Two notes on rigor: this went through a multi-perspective design review before reaching you (maintainer, local-LLM/Ollama user, non-Claude-agent user, data-integrity, privacy, fresh-install, memory-systems lenses), and every mechanism claim in §1–§2 is backed by runnable tests on my side — including a failing-test reproduction of #123, which I'm happy to post on that issue.

---

# Proposal — Memory lifecycle: clock calibration, cold layer, and Deep Recall

## 1. Problem

Two independent roots, plus one confession:

1. **The lifecycle clock is uncalibrated.** Every node is created with the default `stability = 1.0` — the `fsrs_initial_stability` setting exists but is never wired into node creation. Stability only grows on *explicit* recall (whisper injections don't count as access), which is rare by design. With `fsrs_decay_threshold = 0.3`, `R = exp(-t/1.0)` crosses the threshold after **~29 hours**: a node not explicitly recalled within a day of creation becomes a decay candidate, by construction. "Archival" therefore means *older than a day or two*, not *long-tail stale* — and every forgetting gate keyed to retrievability or importance is currently selecting on noise (any node past `deletion_min_archival_days` has R ≈ 0; importance scores in practice don't reach the protection thresholds). Fixing the knob alone is not enough: thousands of already-persisted nodes carry the broken `stability = 1.0` and would keep the broken clock forever — the stock needs a one-shot migration, not just the flow.
2. **`deleted/` conflates three different things.** Merge tombstones (undo semantics), manual deletes, and forgetting output all land in the same directory, and soft-delete removes the node from the index entirely — so a future "cold layer" built on it would be invisible to any search, and the purge job is origin-blind.
3. **Part of the original push for deletion was #22**: the graph view becomes unusable on a large store, so "fewer nodes" looked like the fix. The right fix is the renderer (see §6) — which decouples the UI pain from the lifecycle decision.

## 2. Current lifecycle

```mermaid
flowchart TD
  classDef issue fill:#fde8e8,stroke:#c0392b,color:#7b241c
  classDef hot fill:#e8f6f0,stroke:#1e8449,color:#145a32
  classDef surface fill:#fef9e7,stroke:#b7950b,color:#7d6608
  classDef gone fill:#f2f3f4,stroke:#7f8c8d,color:#515a5a

  ING["Ingest — sessions + remember<br/>no write gate: every extraction becomes a new node"]:::issue
  ING --> W["WORKING<br/>stability = 1.0 — initial-stability knob not wired"]:::hot
  W -- "promote (rare)" --> C["CORE (capped)"]:::hot
  W -- "decay: R &lt; 0.3 after ~29h without explicit recall<br/>'old' ≈ one day, by construction" --> A["ARCHIVAL<br/>out of whisper · still scanned by maintenance"]:::issue

  WH["Whisper (automatic)<br/>core + working only<br/>injection does NOT reinforce stability"]:::surface -. reads .-> W
  RC["Explicit recall<br/>all tiers, archival penalized"]:::surface -. reads .-> A
  MT["Maintenance — dedup / conflict / linker<br/>scans ALL tiers"]:::issue -. scans .-> A

  CD["conflict_detector"] -- "evolution → evolved_from edge · tension → contradicts edge<br/>either way BOTH memories stay live and retrievable — nothing resolves" --> A
  MO["mark_outdated (agent-initiated only)<br/>valid_until = now"] -- "out of every search,<br/>file kept as history" --> A

  A -- "forgetting gates — armed by ORMAH_DELETION_ENABLED<br/>keyed to R + importance: currently degenerate" --> D["deleted/ tombstones<br/>removed from EVERY index — invisible to any search<br/>merge + manual + forgetting mixed together"]:::issue
  MG["duplicate merge"] -- tombstone --> D
  MAN["manual / API delete"] --> D
  D -- "purge after retention — origin-blind" --> G[("hard delete<br/>irreversible")]:::gone
```

One more honesty note about today's "hard delete": it is not complete. Full-node snapshots persist indefinitely in `audit_log.node_snapshot` and `merge_history.removed_node_snapshot`, and injected `whisper_log` rows retain the raw prompt text — no purge job touches any of these. §4 addresses this.

## 3. Proposed lifecycle

```mermaid
flowchart TD
  classDef new fill:#eaf2fb,stroke:#2874a6,color:#1a5276
  classDef hot fill:#e8f6f0,stroke:#1e8449,color:#145a32
  classDef gone fill:#f2f3f4,stroke:#7f8c8d,color:#515a5a
  classDef aux fill:#fef9e7,stroke:#b7950b,color:#7d6608,stroke-dasharray:5 5

  SRC["sessions + remember"] --> GATE["WRITE GATE — later phase<br/>ADD / UPDATE / NOOP by embedding similarity<br/>no LLM call by default (optional flag)"]:::new
  GATE -- ADD --> W["WORKING — window of weeks, not days<br/>initial stability wired (~7d ⇒ demotion after ~8–9 unsignaled days;<br/>a 2–3 week window maps to ~12–17d — the eval harness decides)<br/>reinforcement: explicit recall + positive feedback (strong),<br/>whisper injection (weak, capped 1×/day/node); negative feedback lowers stability"]:::hot
  GATE -- "UPDATE: new fact supersedes old" --> VU
  W -- promote --> C["CORE"]:::hot
  W -- "decay: weeks without any signal" --> A["ARCHIVAL — genuinely cold<br/>out of whisper AND out of maintenance scans"]
  A -- "forgetting gates (real signals)<br/>INVARIANT: this transition only arms once the cold layer exists" --> COLD["COLD — new state<br/>STAYS INDEXED, flagged out of normal queries<br/>provenance in FRONTMATTER: when/why it went cold"]:::new
  COLD -- "DEEP RECALL — MCP tool + HTTP route<br/>explicit, instrumented (hits by age)" --> S["resurfaced memory<br/>with provenance"]:::new
  S -- "proves useful: feedback+" --> W
  COLD -- "purge: explicit / opt-in only<br/>criterion declared up front, values from telemetry (§4)" --> G[("hard delete<br/>+ satellite-table redaction")]:::gone

  SLEEP["sleep-cycle distillation — later phase<br/>consolidate old episodes into durable gist per space<br/>inherits per-run caps from day one"]:::aux -. curation .-> A

  CD["conflict_detector"] -- "evolution conflicts: newer/older already computed today" --> P["proposal: mark_outdated(older)<br/>ALWAYS through proposal review — never auto-applied"]:::new
  P -- "approved" --> VU["valid_until = now<br/>invalidated: out of search, kept as history/provenance"]:::new

  MG["merge tombstone — deleted_reason = merged"] -- "retention window (undo semantics)" --> G
  PRIV["explicit user intent — deleted_reason = privacy<br/>immediate hard delete + immediate redaction"] --> G
```

## 4. Design points

**Reinforcement hierarchy (clock calibration).** Stability grows from a hierarchy of signals rather than explicit recall alone: explicit recall and positive feedback are strong reinforcement; whisper injection is weak reinforcement, capped per `(node_id)` per day to avoid a rich-get-richer loop; negative feedback *lowers* stability, giving the clock its missing failure signal. `fsrs_initial_stability` gets wired, and — critically — a **one-shot migration recalculates `stability` for already-persisted nodes** (same pattern as the existing `archived_at` legacy backfill), so the fix reaches the stock, not just future writes. Two honest caveats, telemetry-tracked rather than hand-waved: the cap limits reinforcement *magnitude* but not the exposure-bias loop (injected nodes rank better, get injected more) — the eval harness measures stability concentration in the top-k to verify the cap suffices; and feedback arrival is agent-dependent (agents must call `submit_feedback`; the transcript-mining LLM judge is the only agent-independent signal and stays off by default for local-LLM users) — per-agent cooling rates are part of the calibration telemetry before the knob value is final. The importance-scorer revisit has a declared shape as well: measure the per-type score distribution, then rescale the scorer's output so the existing protection thresholds fall inside the observed range — thresholds keep their semantic meaning; the scorer moves.

**Cold is a state, not a tombstone.** Forgetting transitions archival → cold: the node stays in the index, flagged out of normal recall and out of maintenance scans, with provenance (when/why) stamped **in frontmatter — the file remains the source of truth**, following the `archived_at` pattern point-by-point (index column + migration + `full_rebuild` insert list + backfill task), since each of those steps missed means the field silently reverts on rebuild. Excluding cold and archival from maintenance scans is where the accumulating operational cost actually bites; it shrinks the maintenance working set to the active tiers. `Tier.cold` is a schema-visible change (backend enum + UI type union + external MCP clients). The compatibility contract is concrete: the enum change is **additive** — API/MCP responses may carry a tier value older clients don't know, and the documented behavior is to treat an unknown tier as archival-like rather than fail validation; MCP tool schemas update in the same PR, and the UI sits behind a feature flag until the frontend supports the new state.

**Deep Recall is a surface with a declared trigger, not a hope.** It ships as an MCP tool **and** an HTTP route (mirroring `recall`), agent-agnostic from day one. Its tool description tells the model *when* to reach for it — normal recall returned weak/empty results and the question references past context — and a server-side auto-escalation (deep recall suggested when normal recall confidence is low) is evaluated in the same PR. Deep-recall usage telemetry (calls, hits, hit age) is part of the PR, because a cold layer nobody queries is dead code — the same criticism this proposal makes of today's invisible tombstones. Search runs on embedding similarity over the cold partition; no LLM call by default.

**`deleted_reason` splits the tombstone semantics.** Merge and manual tombstones remain a pure undo/purge queue (retention window, then hard delete — as today). Forgetting no longer writes there at all. Pre-existing tombstones without the field get `deleted_reason = "legacy_unknown"` and the most conservative policy: never auto-purged. The `archival_soft_cap` backstop is dropped — bypassing the staleness gates to hit a count is indefensible as-is, and no count-based eviction returns without explicit staleness semantics.

**Conflicts resolve instead of accumulating.** Today the conflict detector already distinguishes *evolution* (writes an `evolved_from` edge, computing which node is newer) from *tension* (writes `contradicts`) — but in both cases the memories stay live and compete in ranking. For evolution conflicts the resolution data already exists: the detector emits a *proposal* to `mark_outdated` the older node (`valid_until = now`) — zero new LLM cost, and **always through proposal review, never auto-applied** (a write with destructive-adjacent semantics stays gated; #82's proposals UI gives it a surface). Invalidated memories leave every search surface but remain on disk as history.

**An honest erasure contract.** Hard delete has two doors — expired merge/manual tombstones, and explicit user intent (`deleted_reason = "privacy"`, immediate, skipping retention) — but the doors are only real if the satellite copies go too. Today they don't: full-node snapshots persist in `audit_log` and `merge_history`, and injected `whisper_log`/`retrieval_events` rows keep prompt text, with no purge job touching any of them. The purge step therefore gains **redaction**: `UPDATE ... SET snapshot/prompt_text = NULL WHERE node_id = ?` across the three satellite tables — audit structure (operation, node id, timestamps) survives; content doesn't. Redaction runs inside a single index transaction, and its PR carries a concurrency test — privacy purge racing a concurrent file save on the same node — so the one erasure path that promises "immediate and hard" does not silently inherit the FileStore lost-update class (#125). Backups are documented as the remaining honest caveat (a purged node can survive in up to N rotating snapshots). Distillation (later phase) must never absorb a node already marked for privacy deletion; privacy runs before the sleep cycle. Until the redaction lands, docs state the gap explicitly rather than claiming complete erasure.

| Origin of dead weight | Treatment | Rationale |
|---|---|---|
| Conflict / superseded | `mark_outdated` via proposal — no delete | Out of every search, kept as history/provenance |
| Merge tombstone | Soft → hard after retention (as today) | Content absorbed by survivor; tombstone is undo only |
| Cooled by forgetting | Cold state — Deep Recall searchable | Deleting destroys the evidence needed to justify deleting |
| Explicit user intent | Hard + immediate satellite redaction | The one case where hard delete is a requirement |
| Legacy tombstone (no reason) | Never auto-purged | Origin unknown → most conservative policy |

**Purge policy has a declared shape, not a "later".** Whether organically-created memories are ever purged is decided with deep-recall telemetry, but the criterion is declared now so it can actually mature: a cold node becomes purge-eligible only after ≥ N days cold, with the cold layer having served ≥ M deep-recall queries in that window and the node never surfacing (values for N/M set from telemetry, not opinion). If cold-partition growth measurably pressures the vector index before the criterion matures, the cold partition moves out of the hot vector index — growth has a named ceiling either way.

**Human-memory sanity check.** Forgetting in humans is loss of *access*, not erasure — effortful, cued retrieval still reaches it, and consolidation (distilling episodes into durable gist) is how the long tail pays rent. Purge has no biological analogue; it's an engineering decision and is justified here as one (hygiene + explicit intent), not as "memory decay".

## 5. Delivery plan (small PRs, replacing the current branch shape)

1. **(a) Clock calibration** — wire `fsrs_initial_stability`; reinforcement hierarchy (incl. negative-feedback signal); **one-shot stability migration for the existing stock**; importance-scorer revisit. Merge gates: a decay-curve regression test (days-to-archival by access pattern) so the knob value is testable, not aspirational, **and** the exposure-bias measurement — stability concentration in the top-k over simulated reinforcement cycles — because an undetected rich-get-richer loop would propagate silently into every later phase. The stock migration is idempotent (per-node guard, crash-safe re-run — the same guarantee the `archived_at` backfill already has). Ships with telemetry on whisper-pool size and per-agent cooling rates, and a documented rollback (revert the knob) if whisper precision degrades before (c) lands.
2. **(b) PR #31 rebuilt clean** from current `main`, slimmed to the transition machinery: gates + atomic guarded soft-delete + retention, `deleted_reason` split (incl. `legacy_unknown` + `privacy` + satellite redaction), cap dropped. **Blocker: #123** (reindex drops incoming edges — the degree-based protection gate is paper armor until fixed). Merge gate: user-facing lifecycle docs are a merge criterion, not a checklist item. **Invariant: the forgetting archival→cold transition ships dark and only arms once (c) exists** — no half-shipped window where forgetting still produces invisible tombstones. `deletion_enabled=False` stays the default (written invariant). (b) also takes the **audit-durability decision as a merge gate**: hard-delete snapshots mirrored into tombstone frontmatter before unlink (recommended), or `audit_log` joins the backup set — decided and implemented here, because the audit table lives only in the index DB, which backups exclude and `full_rebuild` cannot reconstruct.
3. **(c) Cold layer + Deep Recall** — cold state in frontmatter + index (full archived_at-pattern wiring + backfill), exclusion from normal queries and maintenance scans, MCP+HTTP deep-recall surface with declared trigger, restore/promotion on positive feedback, telemetry. Preconditions: **#29** (file-before-index ordering) and **#125** (FileStore write serialization) land first — (c) adds a new concurrent frontmatter writer (promotion), which is exactly the pattern both issues attack; #29 additionally simplifies or removes the TOCTOU guard machinery. Merge gate: deep-recall precision/recall metrics exist in the eval harness (they don't today) over a synthetic corpus with known-cold nodes.
4. **(d) WebGL graph view for #22** — sigma.js + graphology (FA2 layout), active-first loading, per-space drill-down; already running in our fork, offered upstream as a separate PR.
5. **(e) Evolution-conflict resolution proposals** — the conflict detector emits `mark_outdated` proposals for evolution conflicts (the newer/older determination it already computes today); always through proposal review, never auto-applied (#82's proposals UI gives it a surface; API/CLI approval works meanwhile). Small and independent of (a)–(d).

Every PR carries its explicit settings delta (added/removed knobs) in the PR body. Coordination note: **#81/#87/#90** (maintenance throughput) are relieved substantially by (c) shrinking the scan set.

## 6. Open questions

1. **Normal recall boundary:** keep explicit recall searching archival (penalized, as today) with whisper active-only and deep recall adding cold — or make normal recall active-only? Preference: keep archival in explicit recall; but this is a product call.
2. **Working-window target:** initial stability ~7d ⇒ unsignaled demotion at ~8–9 days; ~12–17d ⇒ 2–3 weeks. The eval harness decides; the decay-curve regression test in (a) makes whichever value testable.
3. **Distillation (sleep-cycle consolidation)** is sketched as a later phase — worth an issue of its own? (If pursued: inherits per-run caps from day one, and never absorbs privacy-deleted nodes.)

r-spade:
> @r-spade — here's the considered response I owed you.
> 
> Before answering I went digging into the archival tier, as promised, and what I found reshaped the plan: the lifecycle clock underneath decay _and_ this PR's gates is effectively uncalibrated (details in §1 below — every gate keyed to retrievability or importance is currently selecting on noise). So the honest sequence starts with calibration, not with any forgetting policy.
> 
> On the design direction: I'm on board with Deep Recall, with one amendment from the digging — cold should be an **indexed state**, not a tombstone, because today's soft-delete removes the node from every index, so there would be nothing for a deep search to find. The proposal below decomposes your lifecycle into small PRs, addresses each of your five findings, and adds an honest erasure contract I found we need along the way.
> 
> Two notes on rigor: this went through a multi-perspective design review before reaching you (maintainer, local-LLM/Ollama user, non-Claude-agent user, data-integrity, privacy, fresh-install, memory-systems lenses), and every mechanism claim in §1–§2 is backed by runnable tests on my side — including a failing-test reproduction of #123, which I'm happy to post on that issue.
> 
> # Proposal — Memory lifecycle: clock calibration, cold layer, and Deep Recall
> ## 1. Problem
> Two independent roots, plus one confession:
> 
> 1. **The lifecycle clock is uncalibrated.** Every node is created with the default `stability = 1.0` — the `fsrs_initial_stability` setting exists but is never wired into node creation. Stability only grows on _explicit_ recall (whisper injections don't count as access), which is rare by design. With `fsrs_decay_threshold = 0.3`, `R = exp(-t/1.0)` crosses the threshold after **~29 hours**: a node not explicitly recalled within a day of creation becomes a decay candidate, by construction. "Archival" therefore means _older than a day or two_, not _long-tail stale_ — and every forgetting gate keyed to retrievability or importance is currently selecting on noise (any node past `deletion_min_archival_days` has R ≈ 0; importance scores in practice don't reach the protection thresholds). Fixing the knob alone is not enough: thousands of already-persisted nodes carry the broken `stability = 1.0` and would keep the broken clock forever — the stock needs a one-shot migration, not just the flow.
> 2. **`deleted/` conflates three different things.** Merge tombstones (undo semantics), manual deletes, and forgetting output all land in the same directory, and soft-delete removes the node from the index entirely — so a future "cold layer" built on it would be invisible to any search, and the purge job is origin-blind.
> 3. **Part of the original push for deletion was [perf(ui): make graph view scale for large number of nodes #22](https://github.com/r-spade/ormah/issues/22)**: the graph view becomes unusable on a large store, so "fewer nodes" looked like the fix. The right fix is the renderer (see §6) — which decouples the UI pain from the lifecycle decision.
> 
> ## 2. Current lifecycle
> ```mermaid
> flowchart TD
>   classDef issue fill:#fde8e8,stroke:#c0392b,color:#7b241c
>   classDef hot fill:#e8f6f0,stroke:#1e8449,color:#145a32
>   classDef surface fill:#fef9e7,stroke:#b7950b,color:#7d6608
>   classDef gone fill:#f2f3f4,stroke:#7f8c8d,color:#515a5a
> 
>   ING["Ingest — sessions + remember<br/>no write gate: every extraction becomes a new node"]:::issue
>   ING --> W["WORKING<br/>stability = 1.0 — initial-stability knob not wired"]:::hot
>   W -- "promote (rare)" --> C["CORE (capped)"]:::hot
>   W -- "decay: R &lt; 0.3 after ~29h without explicit recall<br/>'old' ≈ one day, by construction" --> A["ARCHIVAL<br/>out of whisper · still scanned by maintenance"]:::issue
> 
>   WH["Whisper (automatic)<br/>core + working only<br/>injection does NOT reinforce stability"]:::surface -. reads .-> W
>   RC["Explicit recall<br/>all tiers, archival penalized"]:::surface -. reads .-> A
>   MT["Maintenance — dedup / conflict / linker<br/>scans ALL tiers"]:::issue -. scans .-> A
> 
>   CD["conflict_detector"] -- "evolution → evolved_from edge · tension → contradicts edge<br/>either way BOTH memories stay live and retrievable — nothing resolves" --> A
>   MO["mark_outdated (agent-initiated only)<br/>valid_until = now"] -- "out of every search,<br/>file kept as history" --> A
> 
>   A -- "forgetting gates — armed by ORMAH_DELETION_ENABLED<br/>keyed to R + importance: currently degenerate" --> D["deleted/ tombstones<br/>removed from EVERY index — invisible to any search<br/>merge + manual + forgetting mixed together"]:::issue
>   MG["duplicate merge"] -- tombstone --> D
>   MAN["manual / API delete"] --> D
>   D -- "purge after retention — origin-blind" --> G[("hard delete<br/>irreversible")]:::gone
> ```
> One more honesty note about today's "hard delete": it is not complete. Full-node snapshots persist indefinitely in `audit_log.node_snapshot` and `merge_history.removed_node_snapshot`, and injected `whisper_log` rows retain the raw prompt text — no purge job touches any of these. §4 addresses this.
> 
> ## 3. Proposed lifecycle
> ```mermaid
> flowchart TD
>   classDef new fill:#eaf2fb,stroke:#2874a6,color:#1a5276
>   classDef hot fill:#e8f6f0,stroke:#1e8449,color:#145a32
>   classDef gone fill:#f2f3f4,stroke:#7f8c8d,color:#515a5a
>   classDef aux fill:#fef9e7,stroke:#b7950b,color:#7d6608,stroke-dasharray:5 5
> 
>   SRC["sessions + remember"] --> GATE["WRITE GATE — later phase<br/>ADD / UPDATE / NOOP by embedding similarity<br/>no LLM call by default (optional flag)"]:::new
>   GATE -- ADD --> W["WORKING — window of weeks, not days<br/>initial stability wired (~7d ⇒ demotion after ~8–9 unsignaled days;<br/>a 2–3 week window maps to ~12–17d — the eval harness decides)<br/>reinforcement: explicit recall + positive feedback (strong),<br/>whisper injection (weak, capped 1×/day/node); negative feedback lowers stability"]:::hot
>   GATE -- "UPDATE: new fact supersedes old" --> VU
>   W -- promote --> C["CORE"]:::hot
>   W -- "decay: weeks without any signal" --> A["ARCHIVAL — genuinely cold<br/>out of whisper AND out of maintenance scans"]
>   A -- "forgetting gates (real signals)<br/>INVARIANT: this transition only arms once the cold layer exists" --> COLD["COLD — new state<br/>STAYS INDEXED, flagged out of normal queries<br/>provenance in FRONTMATTER: when/why it went cold"]:::new
>   COLD -- "DEEP RECALL — MCP tool + HTTP route<br/>explicit, instrumented (hits by age)" --> S["resurfaced memory<br/>with provenance"]:::new
>   S -- "proves useful: feedback+" --> W
>   COLD -- "purge: explicit / opt-in only<br/>criterion declared up front, values from telemetry (§4)" --> G[("hard delete<br/>+ satellite-table redaction")]:::gone
> 
>   SLEEP["sleep-cycle distillation — later phase<br/>consolidate old episodes into durable gist per space<br/>inherits per-run caps from day one"]:::aux -. curation .-> A
> 
>   CD["conflict_detector"] -- "evolution conflicts: newer/older already computed today" --> P["proposal: mark_outdated(older)<br/>ALWAYS through proposal review — never auto-applied"]:::new
>   P -- "approved" --> VU["valid_until = now<br/>invalidated: out of search, kept as history/provenance"]:::new
> 
>   MG["merge tombstone — deleted_reason = merged"] -- "retention window (undo semantics)" --> G
>   PRIV["explicit user intent — deleted_reason = privacy<br/>immediate hard delete + immediate redaction"] --> G
> ```
> ## 4. Design points
> **Reinforcement hierarchy (clock calibration).** Stability grows from a hierarchy of signals rather than explicit recall alone: explicit recall and positive feedback are strong reinforcement; whisper injection is weak reinforcement, capped per `(node_id)` per day to avoid a rich-get-richer loop; negative feedback _lowers_ stability, giving the clock its missing failure signal. `fsrs_initial_stability` gets wired, and — critically — a **one-shot migration recalculates `stability` for already-persisted nodes** (same pattern as the existing `archived_at` legacy backfill), so the fix reaches the stock, not just future writes. Two honest caveats, telemetry-tracked rather than hand-waved: the cap limits reinforcement _magnitude_ but not the exposure-bias loop (injected nodes rank better, get injected more) — the eval harness measures stability concentration in the top-k to verify the cap suffices; and feedback arrival is agent-dependent (agents must call `submit_feedback`; the transcript-mining LLM judge is the only agent-independent signal and stays off by default for local-LLM users) — per-agent cooling rates are part of the calibration telemetry before the knob value is final. The importance-scorer revisit has a declared shape as well: measure the per-type score distribution, then rescale the scorer's output so the existing protection thresholds fall inside the observed range — thresholds keep their semantic meaning; the scorer moves.
> 
> **Cold is a state, not a tombstone.** Forgetting transitions archival → cold: the node stays in the index, flagged out of normal recall and out of maintenance scans, with provenance (when/why) stamped **in frontmatter — the file remains the source of truth**, following the `archived_at` pattern point-by-point (index column + migration + `full_rebuild` insert list + backfill task), since each of those steps missed means the field silently reverts on rebuild. Excluding cold and archival from maintenance scans is where the accumulating operational cost actually bites; it shrinks the maintenance working set to the active tiers. `Tier.cold` is a schema-visible change (backend enum + UI type union + external MCP clients). The compatibility contract is concrete: the enum change is **additive** — API/MCP responses may carry a tier value older clients don't know, and the documented behavior is to treat an unknown tier as archival-like rather than fail validation; MCP tool schemas update in the same PR, and the UI sits behind a feature flag until the frontend supports the new state.
> 
> **Deep Recall is a surface with a declared trigger, not a hope.** It ships as an MCP tool **and** an HTTP route (mirroring `recall`), agent-agnostic from day one. Its tool description tells the model _when_ to reach for it — normal recall returned weak/empty results and the question references past context — and a server-side auto-escalation (deep recall suggested when normal recall confidence is low) is evaluated in the same PR. Deep-recall usage telemetry (calls, hits, hit age) is part of the PR, because a cold layer nobody queries is dead code — the same criticism this proposal makes of today's invisible tombstones. Search runs on embedding similarity over the cold partition; no LLM call by default.
> 
> **`deleted_reason` splits the tombstone semantics.** Merge and manual tombstones remain a pure undo/purge queue (retention window, then hard delete — as today). Forgetting no longer writes there at all. Pre-existing tombstones without the field get `deleted_reason = "legacy_unknown"` and the most conservative policy: never auto-purged. The `archival_soft_cap` backstop is dropped — bypassing the staleness gates to hit a count is indefensible as-is, and no count-based eviction returns without explicit staleness semantics.
> 
> **Conflicts resolve instead of accumulating.** Today the conflict detector already distinguishes _evolution_ (writes an `evolved_from` edge, computing which node is newer) from _tension_ (writes `contradicts`) — but in both cases the memories stay live and compete in ranking. For evolution conflicts the resolution data already exists: the detector emits a _proposal_ to `mark_outdated` the older node (`valid_until = now`) — zero new LLM cost, and **always through proposal review, never auto-applied** (a write with destructive-adjacent semantics stays gated; #82's proposals UI gives it a surface). Invalidated memories leave every search surface but remain on disk as history.
> 
> **An honest erasure contract.** Hard delete has two doors — expired merge/manual tombstones, and explicit user intent (`deleted_reason = "privacy"`, immediate, skipping retention) — but the doors are only real if the satellite copies go too. Today they don't: full-node snapshots persist in `audit_log` and `merge_history`, and injected `whisper_log`/`retrieval_events` rows keep prompt text, with no purge job touching any of them. The purge step therefore gains **redaction**: `UPDATE ... SET snapshot/prompt_text = NULL WHERE node_id = ?` across the three satellite tables — audit structure (operation, node id, timestamps) survives; content doesn't. Redaction runs inside a single index transaction, and its PR carries a concurrency test — privacy purge racing a concurrent file save on the same node — so the one erasure path that promises "immediate and hard" does not silently inherit the FileStore lost-update class (#125). Backups are documented as the remaining honest caveat (a purged node can survive in up to N rotating snapshots). Distillation (later phase) must never absorb a node already marked for privacy deletion; privacy runs before the sleep cycle. Until the redaction lands, docs state the gap explicitly rather than claiming complete erasure.
> 
> Origin of dead weight	Treatment	Rationale
> Conflict / superseded	`mark_outdated` via proposal — no delete	Out of every search, kept as history/provenance
> Merge tombstone	Soft → hard after retention (as today)	Content absorbed by survivor; tombstone is undo only
> Cooled by forgetting	Cold state — Deep Recall searchable	Deleting destroys the evidence needed to justify deleting
> Explicit user intent	Hard + immediate satellite redaction	The one case where hard delete is a requirement
> Legacy tombstone (no reason)	Never auto-purged	Origin unknown → most conservative policy
> **Purge policy has a declared shape, not a "later".** Whether organically-created memories are ever purged is decided with deep-recall telemetry, but the criterion is declared now so it can actually mature: a cold node becomes purge-eligible only after ≥ N days cold, with the cold layer having served ≥ M deep-recall queries in that window and the node never surfacing (values for N/M set from telemetry, not opinion). If cold-partition growth measurably pressures the vector index before the criterion matures, the cold partition moves out of the hot vector index — growth has a named ceiling either way.
> 
> **Human-memory sanity check.** Forgetting in humans is loss of _access_, not erasure — effortful, cued retrieval still reaches it, and consolidation (distilling episodes into durable gist) is how the long tail pays rent. Purge has no biological analogue; it's an engineering decision and is justified here as one (hygiene + explicit intent), not as "memory decay".
> 
> ## 5. Delivery plan (small PRs, replacing the current branch shape)
> 1. **(a) Clock calibration** — wire `fsrs_initial_stability`; reinforcement hierarchy (incl. negative-feedback signal); **one-shot stability migration for the existing stock**; importance-scorer revisit. Merge gates: a decay-curve regression test (days-to-archival by access pattern) so the knob value is testable, not aspirational, **and** the exposure-bias measurement — stability concentration in the top-k over simulated reinforcement cycles — because an undetected rich-get-richer loop would propagate silently into every later phase. The stock migration is idempotent (per-node guard, crash-safe re-run — the same guarantee the `archived_at` backfill already has). Ships with telemetry on whisper-pool size and per-agent cooling rates, and a documented rollback (revert the knob) if whisper precision degrades before (c) lands.
> 2. **(b) PR [feat(background): bounded forgetting — delete dead-weight archival nodes (#28) #31](https://github.com/r-spade/ormah/pull/31) rebuilt clean** from current `main`, slimmed to the transition machinery: gates + atomic guarded soft-delete + retention, `deleted_reason` split (incl. `legacy_unknown` + `privacy` + satellite redaction), cap dropped. **Blocker: [bug(index): reindexing a node deletes its incoming edges and cannot restore them — the connection lives in the source node's markdown #123](https://github.com/r-spade/ormah/issues/123)** (reindex drops incoming edges — the degree-based protection gate is paper armor until fixed). Merge gate: user-facing lifecycle docs are a merge criterion, not a checklist item. **Invariant: the forgetting archival→cold transition ships dark and only arms once (c) exists** — no half-shipped window where forgetting still produces invisible tombstones. `deletion_enabled=False` stays the default (written invariant). (b) also takes the **audit-durability decision as a merge gate**: hard-delete snapshots mirrored into tombstone frontmatter before unlink (recommended), or `audit_log` joins the backup set — decided and implemented here, because the audit table lives only in the index DB, which backups exclude and `full_rebuild` cannot reconstruct.
> 3. **(c) Cold layer + Deep Recall** — cold state in frontmatter + index (full archived_at-pattern wiring + backfill), exclusion from normal queries and maintenance scans, MCP+HTTP deep-recall surface with declared trigger, restore/promotion on positive feedback, telemetry. Preconditions: **[engine: mutators write file before the index lock — deletion/eligibility race (root fix for #28) #29](https://github.com/r-spade/ormah/issues/29)** (file-before-index ordering) and **[bug(store): the load -> modify -> save cycle on a node's markdown is not serialized — FileStore has no lock, so a lost update is possible #125](https://github.com/r-spade/ormah/issues/125)** (FileStore write serialization) land first — (c) adds a new concurrent frontmatter writer (promotion), which is exactly the pattern both issues attack; [engine: mutators write file before the index lock — deletion/eligibility race (root fix for #28) #29](https://github.com/r-spade/ormah/issues/29) additionally simplifies or removes the TOCTOU guard machinery. Merge gate: deep-recall precision/recall metrics exist in the eval harness (they don't today) over a synthetic corpus with known-cold nodes.
> 4. **(d) WebGL graph view for [perf(ui): make graph view scale for large number of nodes #22](https://github.com/r-spade/ormah/issues/22)** — sigma.js + graphology (FA2 layout), active-first loading, per-space drill-down; already running in our fork, offered upstream as a separate PR.
> 5. **(e) Evolution-conflict resolution proposals** — the conflict detector emits `mark_outdated` proposals for evolution conflicts (the newer/older determination it already computes today); always through proposal review, never auto-applied ([feat(ui): add a Proposals review screen — merge/conflict proposals have backend endpoints but no UI #82](https://github.com/r-spade/ormah/issues/82)'s proposals UI gives it a surface; API/CLI approval works meanwhile). Small and independent of (a)–(d).
> 
> Every PR carries its explicit settings delta (added/removed knobs) in the PR body. Coordination note: **#81/#87/#90** (maintenance throughput) are relieved substantially by (c) shrinking the scan set.
> 
> ## 6. Open questions
> 1. **Normal recall boundary:** keep explicit recall searching archival (penalized, as today) with whisper active-only and deep recall adding cold — or make normal recall active-only? Preference: keep archival in explicit recall; but this is a product call.
> 2. **Working-window target:** initial stability ~7d ⇒ unsignaled demotion at ~8–9 days; ~12–17d ⇒ 2–3 weeks. The eval harness decides; the decay-curve regression test in (a) makes whichever value testable.
> 3. **Distillation (sleep-cycle consolidation)** is sketched as a later phase — worth an issue of its own? (If pursued: inherits per-run caps from day one, and never absorbs privacy-deleted nodes.)

@AndreLFSMartins This is unreal! Thanks for the detailed review. I had a quick glance and it looks like the right direction. Because this is a core functionality and will affect the behaviour and user and agent experience I will have a detailed read through and analysis and get back to you on this with me thoughts. But once again a million thanks.

I have been a bit busy getting cross machine sync working for ormah so Ormah can go from cross agent (need to figure out how to add more integrations for new agents @YagizEfeGokce checked PI out of the list) to cross machine, encrypted memory protection and backup. I have it in decent shape now, need to get a few more things over the line and I should be able to release it. 

Will notarize the Mac app soon too. So that becomes the distribution channel for everything ormah. 

I know I have a few PRs and issues to get through. Will do soon, but if any is urgent if you can tag me and mark it urgent I will get to it ASAP.

AndreLFSMartins:
Moving this to **draft** — not stalled, deliberately gated.

The [#191 design decision](https://github.com/r-spade/ormah/issues/191#issuecomment-5287036815) states it directly:

> Bounded forgetting in #28/#31 should wait until these lifecycle signals are corrected.

That is the same conclusion I reached in [my July response above](https://github.com/r-spade/ormah/pull/31#issuecomment-4973151833) — the honest sequence starts with calibration, not with a forgetting policy. This PR's protection gates read retrievability, importance, access counts and edge degree; every one of those is currently selecting on noise, so enabling deletion now would protect and delete on bad data.

Blocked on the four lifecycle issues, in their agreed landing order:

- #220 — separate surfaced results from confirmed use
- #222 — retrievability alone controls demotion
- #221 — bounded reinforcement + daily cooldown
- #223 — reversible promotion + the seven-day initial lease

Draft rather than closed because the work is intact and still wanted: 7 protection gates, cap backstop, soft-delete with retention window, hard purge, plus the gate #6 `edge_type` fix pushed today (`bc2ca03`, `6a4027a` — a lone `contradicts` edge at weight 0.9 no longer immunizes a node the way `supports` does).

Still owed before this comes back for review, from @r-spade's [2026-07-10 review](https://github.com/r-spade/ormah/pull/31#issuecomment-4939076370): the rebase/split so the diff is only the #28 lifecycle, and the `deleted/` semantics question. Both are better done against corrected signals anyway.

I'll mark it ready once #223 lands.

PR #34, opened 2026-06-15
fix(session-watcher): ingest only appended turns via byte cursor (#33)

### LINKED ISSUE #33: bug(session-watcher): active transcripts re-ingested in full on every append — no turn cursor, node_ids accumulate without supersede (~6x store inflation)
## Summary

The session watcher re-ingests an **entire** transcript every time the file changes. For an **active** conversation (a session still being written to), every appended turn changes the file hash, so the watcher re-parses the whole `.jsonl` and re-extracts memories from **all** turns again — creating a fresh batch of nodes each time. Old nodes from prior ingests of the same file are never superseded. The result is severe store inflation: on a real store, ~69% of nodes (≈6.3k of 9.1k) came from session-watcher re-extraction, with single conversations producing an order of magnitude more nodes than they have turns.

## Root cause

`_ingest_session` (`src/ormah/background/session_watcher.py`) only short-circuits on a **whole-file hash match**:

```python
existing = state.get(rel)
if existing and existing.get("hash") == h:
    return False          # unchanged file → skip
...
result = parse_transcript(path)            # parses the ENTIRE file
ingested = engine.ingest_conversation(
    content=result.conversation,           # the FULL conversation, every time
    agent_id="session-watcher", ...
)
new_node_ids = [m["node_id"] for m in ingested]
prev_node_ids = existing.get("node_ids", []) if existing else []
state[rel] = { "hash": h, ...
    "node_ids": prev_node_ids + new_node_ids,   # accumulates, never used to clean up
}
```

There is **no turn/byte cursor**: re-ingestion is all-or-nothing on the whole transcript. An active session's hash changes on every appended turn (and `SessionHandler.on_modified` schedules a debounced re-ingest on each write), so the full conversation is re-extracted repeatedly. The state already records `node_ids` per file but **never uses them to supersede/replace** the prior extraction — they only accumulate.

The cursor that processes "only new content" (`test_cursor_processes_only_new_content`) lives in the **whisper-store** path, not in the watcher — the watcher has no equivalent.

## Evidence (real store, ~9.1k nodes)

Ingestion events vs unique files in one log window: **51 ingestion events / 12 unique files**. Worst offenders re-ingested 19×, 13×, 7× in minutes (the active sessions).

Accumulated `node_ids` per file from `.session_watcher_state` (turns → nodes):

| transcript | user turns | accumulated node_ids |
|------------|-----------:|---------------------:|
| 15c93a00…  | 98 | 320 |
| 1946e505…  | 22 | 229 |
| 3b6503eb…  | 33 | 222 |
| e4d0ed43…  | 25 | 157 |

A 22-turn conversation produced 229 nodes (~10 nodes/turn). Across 127 tracked files the state holds **6149** node_ids, matching the `agent:session-watcher` node count in the index (6309). A healthy ratio (~0.3–0.5 memories/turn) would put session-watcher's contribution near ~700–1000 nodes, so the store is inflated roughly 6×.

No **exact** duplicates exist (`COUNT(*) == COUNT(DISTINCT title||content)`), because each LLM extraction of the same material differs slightly — which is exactly why simple dedup and the (often-disabled) duplicate detector don't catch it.

## Proposed fix

1. **Turn-offset cursor (primary).** Persist the last-ingested user-turn index (or byte offset) per file in the state; on a changed file, feed only turns *after* the cursor to `ingest_conversation`. This makes re-ingestion incremental and idempotent.
2. **Idle-gating (complement).** Skip files still being actively written (e.g., mtime within the debounce/idle window) and ingest once the session goes quiet, so an in-progress conversation isn't re-extracted on every turn.
3. **Supersede-on-reingest (fallback if full re-extract is kept).** Use the already-stored `node_ids` to soft-delete/replace the prior extraction for that file before inserting the new one, instead of accumulating.

(1)+(2) is the correct design; (3) bounds the damage if incremental extraction isn't feasible.

## Impact

- Primary driver of unbounded store growth (~1–1.5k nodes/day observed), which compounds #25 (brute-force vector scan), #26 (auto_linker O(n²)), and #32 (full re-embed on startup) — all scale with `n`.
- Pollutes recall/whisper with many near-duplicate memories of the same conversation.
- A cleanup migration is advisable for already-inflated stores (collapse per-file node sets, keeping the latest/most-complete extraction).

Found bringing a large real store online (≈9.1k nodes, ~80% archival).

AndreLFSMartins:
Fixes #33.

## Problem

The session watcher re-ingests an **entire** transcript every time the file changes. For an **active** conversation (still being written), every appended turn changes the whole-file hash, so `_ingest_session` re-parses the full `.jsonl` and re-extracts memories from **all** turns again — piling up near-duplicate nodes. On a real ~9.1k-node store this accounted for ~69% of nodes (≈6.3k), with single 22-turn conversations producing 229 nodes. No **exact** duplicates appear (each LLM extraction differs slightly), so naive dedup misses it.

## Fix

`parse_transcript` already supports incremental reads (`start_offset` arg + `TranscriptResult.end_offset`) — the watcher just never used them. This wires them in:

- Persist `end_offset` per file in `.session_watcher_state`; on a changed file, re-open from the stored offset so only the **appended** turns are parsed and extracted.
- The `min_turns` gate now applies to the **new** slice — small appends defer until enough new signal accumulates, instead of re-extracting the whole conversation.
- Reset the cursor to `0` when a file shrinks below the stored offset (compaction / rewrite, e.g. `/compact`) so the whole transcript is re-ingested.
- `user_turns` is now cumulative across the file; `node_ids` accumulation is now non-overlapping.

`parser.py` is unchanged.

## Known limitation

An in-place prefix rewrite that does **not** shrink the file (rare; `/compact` shrinks) is not detected by the size-based reset and would ingest from the stale offset. Documented in the commit; a prefix-hash reset can be added later if it ever surfaces.

## Tests

3 new tests, written first (red → green):
- `test_incremental_only_new_turns` — a later change feeds only the appended turns to ingest.
- `test_incremental_defers_small_append` — an append below `min_turns` triggers no second extraction and leaves state untouched.
- `test_shrink_resets_cursor` — a shrunken file is re-ingested from the start.

Full watcher suite green (18/18, 10 pre-existing unchanged). `ruff check` clean on the changed source.

### DIFF
--- a/docs/10 - Hippocampus and Session Watcher.md
+++ b/docs/10 - Hippocampus and Session Watcher.md
@@ -63,6 +63,7 @@ Current defaults:
 - `session_watcher_debounce_seconds = 60`
 - `session_watcher_min_turns = 5`
 - `session_watcher_lookback_hours = 72`
+- `session_watcher_idle_threshold = 30.0`
 - `feedback_llm_judge_enabled = false`
 - `feedback_llm_judge_min_confidence = 0.75`
 
--- a/docs/12 - Configuration Reference.md
+++ b/docs/12 - Configuration Reference.md
@@ -102,6 +102,7 @@ Operational note: default-enabled does not mean active without configured watch 
 | `session_watcher_debounce_seconds` | `60.0` |
 | `session_watcher_min_turns` | `5` |
 | `session_watcher_lookback_hours` | `72` |
+| `session_watcher_idle_threshold` | `30.0` |
 
 `session_watcher_dir` is the primary watch directory and remains the historical Claude Code
 default. When it is left at the default, Ormah also watches `~/.codex/sessions` if that
--- a/src/ormah/adapters/cli_adapter.py
+++ b/src/ormah/adapters/cli_adapter.py
@@ -439,19 +439,28 @@ def cmd_whisper_store(args):
 
     try:
         result = parse_transcript(path, start_offset=start_offset)
+        if result.leading_orphan:
+            # Cursor left mid-response by an older version: re-parse from the start to
+            # recover the dropped tail with its prompt (one-time full re-extract).
+            start_offset = 0
+            result = parse_transcript(path, start_offset=0)
     except Exception:
         sys.exit(0)
 
+    # Commit only the closed ("safe") payload — content proven complete by a terminal
+    # stop_reason or a following user turn — and advance the cursor to its boundary. Like
+    # the session watcher, this never splits a multi-record response from its prompt if
+    # the hook fires while a response is still being written.
     min_turns = settings.whisper_out_min_turns
-    if result.user_turn_count < min_turns:
+    if result.safe_user_turn_count < min_turns:
         sys.exit(0)
 
-    if not result.conversation.strip():
+    if not result.safe_conversation.strip():
         sys.exit(0)
 
     space = detect_space_from_dir(cwd) if cwd else None
 
-    body: dict = {"content": result.conversation}
+    body: dict = {"content": result.safe_conversation}
     params: dict = {"extra_tags": "whisper-out"}
     if space:
         params["default_space"] = space
@@ -464,8 +473,9 @@ def cmd_whisper_store(args):
         # Server down, timeout, or any error — exit silently, never block compaction
         sys.exit(0)
 
-    # Update cursor only after successful extraction
-    cursors[cursor_key] = result.end_offset
+    # Update cursor only after successful extraction, to the closed boundary so a
+    # still-in-flight trailing response is re-read (with its prompt) on the next run.
+    cursors[cursor_key] = result.safe_end_offset
     _save_cursors(cursors)
 
     sys.exit(0)
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -785,7 +785,7 @@ def _ingest_session(
 
     logger.info(
         "Session watcher ingested %s (%d new turns, %d memories extracted, %d signals recorded)",
-        rel, result.user_turn_count, count, signals_recorded,
+        rel, result.safe_user_turn_count, count, signals_recorded,
     )
     return True
 
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -766,9 +766,11 @@ def _ingest_session(
         return False
 
     new_node_ids = [m["node_id"] for m in ingested] if isinstance(ingested, list) else []
-    prev_node_ids = existing.get("node_ids", []) if existing else []
-    # prev_offset == 0 means a fresh/whole re-ingest; don't carry stale cumulative turns.
-    prev_turns = existing.get("user_turns", 0) if (existing and prev_offset > 0) else 0
+    # prev_offset == 0 means a fresh/whole re-ingest; don't carry stale cumulative
+    # turns or node_ids forward (the new ingest re-covers them).
+    carry = existing and prev_offset > 0
+    prev_node_ids = existing.get("node_ids", []) if carry else []
+    prev_turns = existing.get("user_turns", 0) if carry else 0
 
     state[rel] = {
         "hash": h,
@@ -854,6 +856,7 @@ class SessionHandler(FileSystemEventHandler):
         self._state = _load_state(watch_dir)
         self._timers: dict[str, Timer] = {}
         self._ingesting: set[str] = set()
+        self._pending: set[str] = set()
         self._lock = Lock()
 
     def _schedule_ingest(self, path: Path) -> None:
@@ -888,7 +891,11 @@ class SessionHandler(FileSystemEventHandler):
         with self._lock:
             self._timers.pop(key, None)
             if key in self._ingesting:
-                return  # an ingest for this path is already running; skip
+                # An ingest for this path is already running and has already parsed
+                # its slice; mark the path so the new content is re-ingested once it
+                # finishes, instead of dropping this event.
+                self._pending.add(key)
+                return
             self._ingesting.add(key)
         try:
             _ingest_session(
@@ -899,6 +906,10 @@ class SessionHandler(FileSystemEventHandler):
         finally:
             with self._lock:
                 self._ingesting.discard(key)
+                rerun = key in self._pending
+                self._pending.discard(key)
+        if rerun:
+            self._schedule_ingest(path)  # re-process content that arrived mid-ingest
 
     def on_created(self, event):
         if not event.is_directory and event.src_path.endswith(".jsonl"):
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -694,6 +694,8 @@ def _ingest_session(
     state: dict,
     watch_dir: Path,
     min_turns: int,
+    idle_threshold: float = 30.0,
+    on_defer_active=None,
 ) -> bool:
     """Ingest a single JSONL session transcript if changed. Returns True if ingested."""
     rel = str(path.relative_to(watch_dir))
@@ -724,8 +726,20 @@ def _ingest_session(
         logger.warning("Session transcript parse error for %s: %s", path, e)
         return False
 
-    if result.user_turn_count < min_turns:
-        return False  # too few NEW turns; offset unchanged so they're reconsidered later
+    # Bug 1: no complete new user+assistant pair — cursor stays put.
+    if result.safe_end_offset == prev_offset:
+        return False
+
+    # Bug 2: short tail — defer unless the session looks idle/finished.
+    if result.safe_user_turn_count < min_turns:
+        try:
+            age = time.time() - path.stat().st_mtime
+        except OSError:
+            age = idle_threshold + 1  # treat unstatable file as idle
+        if age <= idle_threshold:
+            if on_defer_active is not None:
+                on_defer_active()  # schedule a retry so the tail is not lost
+            return False
 
     result.session_id = _resolve_transcript_session_id(
         engine,
@@ -738,7 +752,7 @@ def _ingest_session(
 
     try:
         ingested = engine.ingest_conversation(
-            content=result.conversation,
+            content=result.safe_conversation,
             space=space,
             agent_id=result.source,
             extra_tags=["session-transcript"],
@@ -758,12 +772,12 @@ def _ingest_session(
 
     state[rel] = {
         "hash": h,
-        "end_offset": result.end_offset,
+        "end_offset": result.safe_end_offset,
         "last_ingested": datetime.now(timezone.utc).isoformat(),
         "session_id": result.session_id,
         "source": result.source,
         "space": space,
-        "user_turns": prev_turns + result.user_turn_count,
+        "user_turns": prev_turns + result.safe_user_turn_count,
         "node_ids": prev_node_ids + new_node_ids,
         "signals_recorded": signals_recorded,
     }
@@ -830,13 +844,16 @@ class SessionHandler(FileSystemEventHandler):
         watch_dir: Path,
         debounce_seconds: float,
         min_turns: int,
+        idle_threshold: float = 30.0,
     ) -> None:
         self.engine = engine
         self.watch_dir = watch_dir
         self.debounce_seconds = debounce_seconds
         self.min_turns = min_turns
+        self.idle_threshold = idle_threshold
         self._state = _load_state(watch_dir)
         self._timers: dict[str, Timer] = {}
+        self._ingesting: set[str] = set()
         self._lock = Lock()
 
     def _schedule_ingest(self, path: Path) -> None:
@@ -854,11 +871,34 @@ class SessionHandler(FileSystemEventHandler):
             self._timers[key] = timer
             timer.start()
 
-    def _do_ingest(self, path: Path) -> None:
-        """Actually ingest the session (called after debounce)."""
+    def _schedule_retry(self, path: Path) -> None:
+        """Re-attempt ingestion after idle_threshold so an active short tail is not lost."""
+        key = str(path)
         with self._lock:
-            self._timers.pop(str(path), None)
-        _ingest_session(self.engine, path, self._state, self.watch_dir, self.min_turns)
+            if key in self._timers:
+                self._timers[key].cancel()
+            timer = Timer(self.idle_threshold, self._do_ingest, args=(path,))
+            timer.daemon = True
+            self._timers[key] = timer
+            timer.start()
+
+    def _do_ingest(self, path: Path) -> None:
+        """Actually ingest the session (called after debounce or retry)."""
+        key = str(path)
+        with self._lock:
+            self._timers.pop(key, None)
+            if key in self._ingesting:
+                return  # an ingest for this path is already running; skip
+            self._ingesting.add(key)
+        try:
+            _ingest_session(
+                self.engine, path, self._state, self.watch_dir, self.min_turns,
+                idle_threshold=self.idle_threshold,
+                on_defer_active=lambda: self._schedule_retry(path),
+            )
+        finally:
+            with self._lock:
+                self._ingesting.discard(key)
 
     def on_created(self, event):
         if not event.is_directory and event.src_path.endswith(".jsonl"):
@@ -896,6 +936,7 @@ def start_session_watcher(engine: MemoryEngine) -> list[Observer]:
         # Start real-time watcher
         handler = SessionHandler(
             engine, watch_dir, s.session_watcher_debounce_seconds, s.session_watcher_min_turns,
+            s.session_watcher_idle_threshold,
         )
         observer = Observer()
         observer.schedule(handler, str(watch_dir), recursive=True)
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -391,8 +391,16 @@ def _insert_affinity(
 def _record_whisper_usage_signals(
     engine: MemoryEngine,
     transcript: TranscriptResult,
+    turns: list | None = None,
 ) -> int:
-    """Mine transcript responses for clear usage of injected whisper memories."""
+    """Mine transcript responses for clear usage of injected whisper memories.
+
+    *turns* restricts mining to a subset of ``transcript.turns`` (e.g. only the
+    closed/safe blocks of an active session, so a still-growing assistant response
+    is not judged from a partial body). Defaults to all turns.
+    """
+    if turns is None:
+        turns = transcript.turns
     llm_judge_enabled = _feedback_llm_judge_enabled(engine)
     rows = engine.db.conn.execute(
         """
@@ -432,7 +440,7 @@ def _record_whisper_usage_signals(
         prompt_text = row["prompt_text"] or ""
         if prompt_text not in response_cache:
             response_cache[prompt_text] = _assistant_response_after_prompt(
-                transcript.turns,
+                turns,
                 prompt_text,
             )
         response = response_cache[prompt_text]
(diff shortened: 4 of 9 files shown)


r-spade:
I like the direction here. Using the parser's `start_offset` / `end_offset` support seems like the right way to stop repeated whole-transcript extraction.

A couple of edge cases I'm worried about:

First, the watcher can advance the cursor while the transcript is still mid-turn.

Example:

- watcher has already ingested up to byte 1000
- then 5 new user turns are appended, so the new slice passes `min_turns`
- but the 5th assistant response has not been written yet
- the watcher ingests the slice and saves `end_offset` after `User: question 5`
- later `Assistant: answer 5` is appended
- now the next incremental parse starts after the user prompt, sees only the assistant message, has 0 user turns, and skips

That means the final prompt/response pair is never seen together. With the newer feedback-mining work, this also means we may miss whisper usage signals because those need to match a prompt with the following assistant response.

Second, applying `min_turns` to only the appended slice can permanently skip the last few turns of a session.

Example with the default `min_turns = 5`:

- first 10 turns are ingested successfully and the cursor is saved
- the user then has 3 more complete turns
- the session ends
- the appended slice has only 3 user turns, so `_ingest_session` returns `False`
- because the session is over, no later append pushes that slice over the threshold

So those last 3 turns never get memory extraction. After the feedback-loop work, they also never get feedback mining.

One smaller housekeeping note: the branch looks out of date with current `main`, so this probably needs a rebase before these interactions are easy to reason about.

What do you think about holding back the cursor to the last complete user+assistant boundary, keeping a small uncommitted tail buffer, or having a "session looks idle/finished" path where a short final tail is processed?

AndreLFSMartins:
Review in progress — analysing the cursor edge cases before replying.

AndreLFSMartins:
Thanks for the detailed edge-case analysis — both cases were real. Rebased onto current `main` and reworked the cursor so the two failure modes you described can't happen.

## What changed

**Rebased onto `main`** (picks up the feedback-mining work), so the interactions are easier to reason about now.

**1. Mid-turn race → safe boundary in the parser.** `parse_transcript` now exposes `safe_end_offset` / `safe_conversation` / `safe_user_turn_count`. A pair only *completes* at an assistant turn **that has text**. The watcher ingests the safe payload and saves `safe_end_offset` as its cursor, so the ingested content and the cursor always share the same completion boundary. Concretely, for your example: after `User: question 5` the cursor does **not** advance (no completed pair), so the next incremental parse re-reads that user turn and pairs it with `Assistant: answer 5` when it arrives. No prompt/response pair is ever split, which also keeps the whisper-feedback prompt↔response matching intact.

A nice side effect: a `tool_use`-only assistant line does not complete a pair, so `user → tool_use → text` is read as one pair rather than being fragmented.

**2. Session tail loss → idle flush + scheduled retry.** A short tail below `min_turns` is no longer skipped forever. New `session_watcher_idle_threshold` (default 30s, < the 60s debounce): once the file has been idle past the threshold the tail is flushed; while the session still looks active the handler schedules a retry so the final turns aren't stranded waiting for an event that never comes.

## Known limitations (deliberate, documented)

- **Non-append rewrites**: the byte cursor assumes append-only transcripts (it already resets on shrink). A rewrite/compaction that keeps size ≥ the stored offset would seek into changed content. Real Claude Code transcripts are append-only in practice; flagging it as a follow-up rather than adding prefix-hash verification here.
- **Startup catch-up of an *active* short tail**: the catch-up scan already ingests finished sessions (old mtime). The narrow remaining gap is a session written within the idle window at the exact moment of a watcher restart that then never receives another write — it isn't re-ingested until the next write or restart. Closing it fully means unifying the catch-up/handler state and scheduling retries from the scan, which is orthogonal to the cursor fix; tracking as a follow-up.

## Tests

New coverage: mid-turn race, idle tail with a trailing dangling user (no duplication), idle flush, retry fires & ingests after idle, concurrent-ingest in-flight guard + reschedule, shrink resets provenance. Full suite green except 4 pre-existing environmental failures unrelated to this change.

r-spade:
Thanks for the update. The two cases from the previous review look much better now: the dangling-user cursor case is handled by `safe_end_offset`, and the short final tail has the idle flush/retry path.

I think there may still be one related boundary edge case, though. Right now the parser advances `safe_end_offset` after any text-bearing assistant record. That protects against `User: Q5` being committed before `Assistant: A5`, but it can still split one assistant response if the assistant side arrives as multiple text-bearing records.

Example:

- `User: question`
- `Assistant: partial answer`

The watcher can treat that as safe and save the cursor after `partial answer`.

Later the same assistant response continues:

- `Assistant: final answer with the actual memory usage`

Now the next incremental slice starts after the user prompt and sees only the second assistant record. Memory extraction might still ingest the text, but feedback mining cannot match it back to the original user prompt, so any whisper usage evidence in that later assistant text is missed.

Would it be safer to define the active-session safe boundary as "the previous user+assistant block becomes safe when the next real user turn starts," and then use the idle flush path to commit the final block when the file looks finished? That would make the active path more conservative, but it keeps the full assistant response attached to its prompt.

AndreLFSMartins:
You're exactly right, and that multi-record split was the real bug — thanks for catching it. I pushed a969ea1 implementing the boundary you suggested, with one deviation I want to flag.

**What it does.** The active-session safe boundary now advances *only* on a real completion signal, never just because an assistant record bore text:

- **Claude:** a terminal `message.stop_reason` (`end_turn` / `stop_sequence` / `max_tokens` / `refusal`). `tool_use` / `pause_turn` keep the turn open.
- **Codex:** the `event_msg` record whose `payload.type == "task_complete"` (Codex `response_item` records carry no `stop_reason`).
- **Any source:** the next real user turn — so an interrupted/abandoned turn still closes at the following prompt.

The full prompt+response block stays attached on every incremental slice.

**The one deviation: I dropped the idle/mtime flush for the final block.** I started with your "commit the final block on idle" idea but it has the same failure mode it's meant to fix — idle means "no writes for N seconds," not "the response finished." A response that's slow between its first and second text record can cross the idle threshold mid-stream and split at exactly the boundary we're trying to protect. So instead the *completion signal itself* closes the final block: a finished turn always carries a terminal `stop_reason` (or `task_complete`), so there's no need to guess from timing. The only liveness cost is a turn whose process was killed mid-response — that stays in-flight and uningested rather than being split, which I think is the right trade.

This deviation matters most for Codex: ~71% of sampled Codex turns emit ≥2 text-bearing records, and with no `stop_reason` the next-user-turn rule alone would hold every Codex turn until the *following* turn started. `task_complete` closes it at the correct boundary.

**Two more things in the same commit:**

- *Both* incremental ingest paths now use the safe boundary — the session-watcher (`_ingest_session`) and the whisper-out hook (`cli_adapter.cmd_whisper_store`). The latter was still advancing the raw cursor.
- A legacy mid-response cursor (a slice that starts with assistant content before any user turn, persisted by the old code) is detected (`leading_orphan`) and recovered by re-parsing from offset 0, so the stranded tail gets re-paired with its prompt. One-time: after recovery the cursor sits on a safe boundary. The hash short-circuit was relaxed to `hash == h and prev_offset >= size` so an unchanged-but-not-fully-consumed file still gets this recovery pass.

Tests cover terminal vs non-terminal `stop_reason`, Codex `task_complete` closing a multi-record turn, in-flight turns held back on both paths, and the legacy-cursor recovery.

r-spade (approved):
Thanks for the careful iteration here. The completion-signal boundary is the right shape: it keeps multi-record assistant responses attached to their prompt, handles Codex via task_complete, and applies the same safe cursor idea to whisper store as well as the session watcher.

I added one small docs commit for the new idle-threshold setting. Focused verification passed on the updated branch:

- ..................................................                       [100%]
50 passed in 2.89s -> 93 passed
- F401 [*] `datetime.datetime` imported but unused
 --> src/ormah/background/consolidator.py:7:22
  |
5 | import json
6 | import logging
7 | from datetime import datetime, timezone
  |                      ^^^^^^^^
8 |
9 | logger = logging.getLogger(__name__)
  |
help: Remove unused import

F401 [*] `datetime.timezone` imported but unused
 --> src/ormah/background/consolidator.py:7:32
  |
5 | import json
6 | import logging
7 | from datetime import datetime, timezone
  |                                ^^^^^^^^
8 |
9 | logger = logging.getLogger(__name__)
  |
help: Remove unused import

F401 `ormah.embeddings.encoder.get_encoder` imported but unused; consider using `importlib.util.find_spec` to test for availability
  --> src/ormah/background/consolidator.py:28:46
   |
26 |     """
27 |     try:
28 |         from ormah.embeddings.encoder import get_encoder
   |                                              ^^^^^^^^^^^
29 |         from ormah.embeddings.vector_store import VectorStore
30 |     except ImportError:
   |
help: Remove unused import: `ormah.embeddings.encoder.get_encoder`

F401 [*] `time` imported but unused
  --> src/ormah/background/hippocampus.py:10:8
   |
 8 | import logging
 9 | import subprocess
10 | import time
   |        ^^^^
11 | from datetime import datetime, timezone
12 | from pathlib import Path
   |
help: Remove unused import: `time`

F401 [*] `dataclasses.field` imported but unused
  --> src/ormah/background/job_tracker.py:8:36
   |
 6 | import threading
 7 | import time
 8 | from dataclasses import dataclass, field
   |                                    ^^^^^
 9 | from datetime import datetime, timezone
10 | from typing import Any, Callable
   |
help: Remove unused import: `dataclasses.field`

F401 [*] `functools.lru_cache` imported but unused
 --> src/ormah/embeddings/encoder.py:6:23
  |
5 | import logging
6 | from functools import lru_cache
  |                       ^^^^^^^^^
7 |
8 | from ormah.embeddings.base import EmbeddingAdapter
  |
help: Remove unused import: `functools.lru_cache`

E402 Module level import not at top of file
  --> src/ormah/embeddings/hybrid_search.py:36:1
   |
34 |     """Detect whether a query is a natural language question."""
35 |     return _QUESTION_PATTERN.search(query) is not None
36 | from ormah.embeddings.encoder import get_encoder
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
37 | from ormah.embeddings.vector_store import VectorStore
38 | from ormah.index.graph import GraphIndex
   |

E402 Module level import not at top of file
  --> src/ormah/embeddings/hybrid_search.py:37:1
   |
35 |     return _QUESTION_PATTERN.search(query) is not None
36 | from ormah.embeddings.encoder import get_encoder
37 | from ormah.embeddings.vector_store import VectorStore
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
38 | from ormah.index.graph import GraphIndex
   |

E402 Module level import not at top of file
  --> src/ormah/embeddings/hybrid_search.py:38:1
   |
36 | from ormah.embeddings.encoder import get_encoder
37 | from ormah.embeddings.vector_store import VectorStore
38 | from ormah.index.graph import GraphIndex
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |

F401 [*] `pathlib.Path` imported but unused
 --> src/ormah/embeddings/local_adapter.py:6:21
  |
5 | import logging
6 | from pathlib import Path
  |                     ^^^^
7 |
8 | import numpy as np
  |
help: Remove unused import: `pathlib.Path`

F401 [*] `json` imported but unused
 --> src/ormah/index/builder.py:5:8
  |
3 | from __future__ import annotations
4 |
5 | import json
  |        ^^^^
6 | import logging
7 | from pathlib import Path
  |
help: Remove unused import: `json`

F401 [*] `watchdog.events.FileModifiedEvent` imported but unused
 --> src/ormah/store/watcher.py:8:53
  |
6 | from pathlib import Path
7 |
8 | from watchdog.events import FileSystemEventHandler, FileModifiedEvent, FileCreatedEvent, FileDeletedEvent
  |                                                     ^^^^^^^^^^^^^^^^^
9 | from watchdog.observers import Observer
  |
help: Remove unused import

F401 [*] `watchdog.events.FileCreatedEvent` imported but unused
 --> src/ormah/store/watcher.py:8:72
  |
6 | from pathlib import Path
7 |
8 | from watchdog.events import FileSystemEventHandler, FileModifiedEvent, FileCreatedEvent, FileDeletedEvent
  |                                                                        ^^^^^^^^^^^^^^^^
9 | from watchdog.observers import Observer
  |
help: Remove unused import

F401 [*] `watchdog.events.FileDeletedEvent` imported but unused
 --> src/ormah/store/watcher.py:8:90
  |
6 | from pathlib import Path
7 |
8 | from watchdog.events import FileSystemEventHandler, FileModifiedEvent, FileCreatedEvent, FileDeletedEvent
  |                                                                                          ^^^^^^^^^^^^^^^^
9 | from watchdog.observers import Observer
  |
help: Remove unused import

F401 [*] `tempfile` imported but unused
 --> tests/conftest.py:5:8
  |
3 | from __future__ import annotations
4 |
5 | import tempfile
  |        ^^^^^^^^
6 | from pathlib import Path
  |
help: Remove unused import: `tempfile`

F401 [*] `pathlib.Path` imported but unused
 --> tests/conftest.py:6:21
  |
5 | import tempfile
6 | from pathlib import Path
  |                     ^^^^
7 |
8 | import pytest
  |
help: Remove unused import: `pathlib.Path`

F401 [*] `pytest` imported but unused
  --> tests/test_background/test_decay_manager.py:8:8
   |
 6 | from datetime import datetime, timedelta, timezone
 7 |
 8 | import pytest
   |        ^^^^^^
 9 |
10 | from ormah.background.decay_manager import run_decay
   |
help: Remove unused import: `pytest`

E402 Module level import not at top of file
  --> tests/test_background/test_duplicate_merger.py:10:1
   |
 8 | _LLM_PATCH = "ormah.background.llm_client.llm_generate"
 9 |
10 | from ormah.models.node import CreateNodeRequest, NodeType
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |

F401 [*] `pathlib.Path` imported but unused
 --> tests/test_background/test_hippocampus.py:7:21
  |
5 | import json
6 | import time
7 | from pathlib import Path
  |                     ^^^^
8 | from unittest.mock import patch
  |
help: Remove unused import: `pathlib.Path`

F401 [*] `pytest` imported but unused
  --> tests/test_background/test_hippocampus.py:10:8
   |
 8 | from unittest.mock import patch
 9 |
10 | import pytest
   |        ^^^^^^
11 |
12 | from ormah.background.hippocampus import (
   |
help: Remove unused import: `pytest`

F401 [*] `ormah.config.Settings` imported but unused
  --> tests/test_background/test_hippocampus.py:23:26
   |
21 |     stop_hippocampus,
22 | )
23 | from ormah.config import Settings
   |                          ^^^^^^^^
24 | from ormah.engine.memory_engine import MemoryEngine
   |
help: Remove unused import: `ormah.config.Settings`

F401 [*] `ormah.config.Settings` imported but unused
  --> tests/test_background/test_session_watcher.py:22:26
   |
20 |     stop_session_watcher,
21 | )
22 | from ormah.config import Settings
   |                          ^^^^^^^^
23 | from ormah.engine.memory_engine import MemoryEngine
   |
help: Remove unused import: `ormah.config.Settings`

F401 [*] `ormah.engine.memory_engine.MemoryEngine` imported but unused
  --> tests/test_background/test_session_watcher.py:23:40
   |
21 | )
22 | from ormah.config import Settings
23 | from ormah.engine.memory_engine import MemoryEngine
   |                                        ^^^^^^^^^^^^
24 |
25 | _LLM_PATCH = "ormah.background.llm_client.llm_generate"
   |
help: Remove unused import: `ormah.engine.memory_engine.MemoryEngine`

F401 [*] `pytest` imported but unused
  --> tests/test_embeddings/test_hybrid_search_title_cap.py:13:8
   |
11 | from unittest.mock import MagicMock
12 |
13 | import pytest
   |        ^^^^^^
14 |
15 | from ormah.config import Settings
   |
help: Remove unused import: `pytest`

F401 [*] `ormah.models.node.Tier` imported but unused
 --> tests/test_engine/test_feedback.py:7:60
  |
5 | import pytest
6 |
7 | from ormah.models.node import CreateNodeRequest, NodeType, Tier
  |                                                            ^^^^
  |
help: Remove unused import: `ormah.models.node.Tier`

E731 Do not assign a `lambda` expression, use a `def`
   --> tests/test_engine/test_hybrid_search.py:93:9
    |
 91 |           hs = HybridSearch(db, settings)
 92 |           # Stub get_node to return minimal node dicts
 93 | /         _node_dict = lambda nid: {
 94 | |             "id": nid,
 95 | |             "type": "fact",
 96 | |             "tier": "working",
 97 | |             "content": f"content of {nid}",
 98 | |         }
    | |_________^
 99 |           MockGraph.return_value.get_node.side_effect = _node_dict
100 |           MockGraph.return_value.get_nodes_batch.side_effect = lambda ids: {
    |
help: Rewrite `_node_dict` as a `def`

E731 Do not assign a `lambda` expression, use a `def`
   --> tests/test_engine/test_hybrid_search.py:249:9
    |
247 |           db.conn = conn
248 |           hs = HybridSearch(db, settings)
249 | /         _node_dict = lambda nid: {
250 | |             "id": nid,
251 | |             "type": "fact",
252 | |             "tier": "working",
253 | |             "content": f"content of {nid}",
254 | |         }
    | |_________^
255 |           MockGraph.return_value.get_node.side_effect = _node_dict
256 |           MockGraph.return_value.get_nodes_batch.side_effect = lambda ids: {
    |
help: Rewrite `_node_dict` as a `def`

E731 Do not assign a `lambda` expression, use a `def`
   --> tests/test_engine/test_hybrid_search.py:304:9
    |
302 |           db.conn = conn
303 |           hs = HybridSearch(db, settings)
304 | /         _node_dict = lambda nid: {
305 | |             "id": nid,
306 | |             "type": "fact",
307 | |             "tier": "working",
308 | |             "content": f"content of {nid}",
309 | |         }
    | |_________^
310 |           MockGraph.return_value.get_nodes_batch.side_effect = lambda ids: {
311 |               nid: _node_dict(nid) for nid in ids
    |
help: Rewrite `_node_dict` as a `def`

E731 Do not assign a `lambda` expression, use a `def`
   --> tests/test_engine/test_hybrid_search.py:355:9
    |
353 |           db.conn = conn
354 |           hs = HybridSearch(db, settings)
355 | /         _node_dict = lambda nid: {
356 | |             "id": nid,
357 | |             "type": "fact",
358 | |             "tier": "working",
359 | |             "content": f"content of {nid}",
360 | |         }
    | |_________^
361 |           MockGraph.return_value.get_nodes_batch.side_effect = lambda ids: {
362 |               nid: _node_dict(nid) for nid in ids
    |
help: Rewrite `_node_dict` as a `def`

E731 Do not assign a `lambda` expression, use a `def`
   --> tests/test_engine/test_hybrid_search.py:404:9
    |
402 |           db.conn = conn
403 |           hs = HybridSearch(db, settings)
404 | /         _node_dict = lambda nid: {
405 | |             "id": nid,
406 | |             "type": "fact",
407 | |             "tier": "working",
408 | |             "content": f"content of {nid}",
409 | |         }
    | |_________^
410 |           MockGraph.return_value.get_nodes_batch.side_effect = lambda ids: {
411 |               nid: _node_dict(nid) for nid in ids
    |
help: Rewrite `_node_dict` as a `def`

F401 [*] `ormah.models.node.CreateNodeRequest` imported but unused
  --> tests/test_engine/test_ingest.py:8:31
   |
 6 | from unittest.mock import patch
 7 |
 8 | from ormah.models.node import CreateNodeRequest
   |                               ^^^^^^^^^^^^^^^^^
 9 |
10 | _LLM_PATCH = "ormah.background.llm_client.llm_generate"
   |
help: Remove unused import: `ormah.models.node.CreateNodeRequest`

F401 [*] `datetime.datetime` imported but unused
 --> tests/test_engine/test_tier_manager.py:5:22
  |
3 | from __future__ import annotations
4 |
5 | from datetime import datetime, timezone
  |                      ^^^^^^^^
6 |
7 | import pytest
  |
help: Remove unused import

F401 [*] `datetime.timezone` imported but unused
 --> tests/test_engine/test_tier_manager.py:5:32
  |
3 | from __future__ import annotations
4 |
5 | from datetime import datetime, timezone
  |                                ^^^^^^^^
6 |
7 | import pytest
  |
help: Remove unused import

F401 [*] `pytest` imported but unused
 --> tests/test_engine/test_tier_manager.py:7:8
  |
5 | from datetime import datetime, timezone
6 |
7 | import pytest
  |        ^^^^^^
8 |
9 | from ormah.engine.tier_manager import TierManager
  |
help: Remove unused import: `pytest`

F401 [*] `tempfile` imported but unused
 --> tests/test_index/test_feedback_schema.py:6:8
  |
5 | import sqlite3
6 | import tempfile
  |        ^^^^^^^^
7 | from pathlib import Path
  |
help: Remove unused import: `tempfile`

F401 [*] `ormah.models.node.MemoryNode` imported but unused
 --> tests/test_index/test_graph.py:8:31
  |
7 | from ormah.index.graph import GraphIndex
8 | from ormah.models.node import MemoryNode, NodeType
  |                               ^^^^^^^^^^
  |
help: Remove unused import

F401 [*] `ormah.models.node.NodeType` imported but unused
 --> tests/test_index/test_graph.py:8:43
  |
7 | from ormah.index.graph import GraphIndex
8 | from ormah.models.node import MemoryNode, NodeType
  |                                           ^^^^^^^^
  |
help: Remove unused import

F841 Local variable `path` is assigned to but never used
  --> tests/test_store/test_file_cache.py:47:5
   |
45 |     store = FileStore(tmp_path / "nodes")
46 |     node = _make_node()
47 |     path = store.save(node)
   |     ^^^^
48 |
49 |     # Corrupt the cache entry to point to a nonexistent path
   |
help: Remove assignment to unused variable `path`

E741 Ambiguous variable name: `l`
  --> tests/test_store/test_markdown_enriched.py:31:24
   |
29 |         text = serialize_node(node)
30 |         # Simulate a legacy file by stripping confidence line
31 |         lines = [l for l in text.split("\n") if not l.startswith("confidence:")]
   |                        ^
32 |         legacy_text = "\n".join(lines)
33 |         parsed = parse_node(legacy_text)
   |

E741 Ambiguous variable name: `l`
  --> tests/test_store/test_markdown_enriched.py:39:24
   |
37 |         node = _make_node()
38 |         text = serialize_node(node)
39 |         lines = [l for l in text.split("\n") if not l.startswith("importance:")]
   |                        ^
40 |         legacy_text = "\n".join(lines)
41 |         parsed = parse_node(legacy_text)
   |

Found 40 errors.
[*] 27 fixable with the `--fix` option (6 hidden fixes can be enabled with the `--unsafe-fixes` option). on the touched source/test files -> all checks passed

This looks good to merge. Thanks again for sticking with the edge cases; this ended up much more robust.

PR #35, opened 2026-06-16
fix(llm): disable Ollama thinking so it doesn't starve JSON extraction

AndreLFSMartins:
## Problem

qwen3-family models (e.g. `qwen3.5:4b`) are thinking-on by default. Ollama counts the reasoning tokens against `num_predict`, so on large transcripts the model spends the budget thinking and returns an **empty or truncated** response. The structured-JSON extraction in `OllamaAdapter.generate(json_mode=True)` then fails with `Expecting value: line 1 column 1` / `Extra data` / truncated JSON, and the memory is silently dropped.

## Fix

Send `think: false` on the `/api/generate` payload (top-level, alongside `format: json`).

## Evidence

Measured on a real ~2,000-file transcript ingest with `qwen3.5:4b-mlx`:

| | LLM attempts | failures | rate |
|---|---|---|---|
| think on (default) | 80 | 21 | **26%** |
| think off | 22 | 3 | **14%** |

Re-running over the previously-failing files, **19 of 22 were recovered** with thinking disabled.

## Safety

Non-thinking models ignore the flag — verified against `gemma3:4b` (clean `{"ok": true}` response, no error). So this is safe for every Ollama LLM backend, not just thinking-capable ones.

## Tests

`tests/test_background/test_llm_adapters.py` asserts `think: false` is present in the payload regardless of `json_mode`. Full adapter suite green (9 passed).

### DIFF
--- a/src/ormah/background/llm/ollama_adapter.py
+++ b/src/ormah/background/llm/ollama_adapter.py
@@ -29,6 +29,10 @@ class OllamaAdapter(LLMAdapter):
             "model": self.model,
             "prompt": prompt,
             "stream": False,
+            # Disable thinking: reasoning tokens consume the num_predict budget
+            # and on large transcripts starve the JSON, yielding empty/truncated
+            # extractions. Non-thinking models ignore this flag.
+            "think": False,
             "options": {"num_predict": self.num_predict},
         }
         if json_mode:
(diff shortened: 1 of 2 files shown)


r-spade (approved):
Approved. This is a really thoughtful fix, Andre.

The problem write-up made the failure mode easy to understand, and the change itself is nicely contained. I especially appreciate that you backed it with real ingest evidence instead of just a theoretical explanation.

Thanks for putting the time into this and for continuing to make Ormah better. Happy to merge.

PR #36, opened 2026-06-16
fix(server): exit cleanly on port conflict instead of KeepAlive respawn loop

AndreLFSMartins:
## Problem

The autostart server has no defense against its port being unavailable. On macOS the launchd plist uses unconditional `<key>KeepAlive</key><true/>` with **no `ThrottleInterval`**, and `ormah server start` lets uvicorn crash on bind. So whenever `127.0.0.1:8787` is already taken, the job enters an **infinite hot respawn loop**:

1. launchd starts `ormah server start`
2. uvicorn fails to bind → process exits non-zero
3. `KeepAlive` respawns it **immediately** (no backoff)
4. → back to 1, forever — pegging CPU and re-running startup work each time

This fires for **any** occupant of the port, not just a duplicate ormah instance:

- another application is already listening on 8787;
- the user started `ormah server start` in a terminal and the launchd job also comes up;
- `ormah setup` is re-run while the previous server still holds the socket (or it's in `TIME_WAIT`).

In one observed case a leftover autostart job crash-looped **2968 times**, holding the CPU at ~30% continuously while another process owned the port — but the underlying defect is the missing throttle + unconditional respawn + crash-on-bind, which any of the scenarios above can trigger.

## Fix

Defensive hardening of the autostart lifecycle, both ends of the loop:

- **`_cmd_server_start`** pre-checks the port via a new `is_port_in_use()` before launching uvicorn. If something already listens there, it logs and **exits 0** instead of crashing on bind.
- **launchd plist** now uses `KeepAlive={SuccessfulExit: false}` + a `ThrottleInterval` of 30s, so a clean exit does **not** respawn and any genuine crash loop is throttled. (The systemd unit already uses `Restart=on-failure` + `RestartSec=5`, so no change needed there.)

The two changes compose: the port check makes a redundant instance exit cleanly, and `SuccessfulExit:false` ensures that clean exit isn't immediately respawned.

## Tests

`tests/test_server_manager.py` (new):
- `is_port_in_use` returns True for a listening socket, False for a free port
- `ormah server start` does **not** launch uvicorn when the port is already in use
- uvicorn **is** launched when the port is free
- plist renders `SuccessfulExit:false` + `ThrottleInterval`, and no longer the unconditional `KeepAlive=true`

`tests/test_setup.py::TestPlistTemplate` updated to assert the new KeepAlive shape.

```
8 passed
```

### DIFF
--- a/src/ormah/cli.py
+++ b/src/ormah/cli.py
@@ -61,6 +61,19 @@ def _cmd_server_start(args):
     else:
         import uvicorn
         from ormah.config import settings
+        from ormah.console import info
+        from ormah.server_manager import is_port_in_use
+
+        # Pre-flight: if the port is already taken (typically by an existing
+        # ormah server), exit cleanly instead of letting uvicorn crash on bind.
+        # Under launchd/systemd KeepAlive this prevents a respawn loop that would
+        # re-run expensive startup work on every restart.
+        if is_port_in_use(settings.host, settings.port):
+            info(
+                f"A process is already listening on {settings.host}:{settings.port}; "
+                "assuming an ormah server is already running. Exiting."
+            )
+            return
 
         uvicorn.run(
             "ormah.main:app",
--- a/src/ormah/server_manager.py
+++ b/src/ormah/server_manager.py
@@ -9,6 +9,7 @@ import re
 import shlex
 import shutil
 import signal
+import socket
 import subprocess
 import sys
 import threading
@@ -59,7 +60,11 @@ PLIST_TEMPLATE = """\
     <key>PATH</key><string>{bin_dir}:/usr/local/bin:/usr/bin:/bin</string>
   </dict>
   <key>RunAtLoad</key><true/>
-  <key>KeepAlive</key><true/>
+  <key>KeepAlive</key>
+  <dict>
+    <key>SuccessfulExit</key><false/>
+  </dict>
+  <key>ThrottleInterval</key><integer>30</integer>
 </dict>
 </plist>
 """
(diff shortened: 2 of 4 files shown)


r-spade:
Thanks @AndreLFSMartins . I pushed one small follow-up commit on top of this PR: `62b7fb1 fix(server): handle IPv6 port preflight`.

The original fix is good for the default `127.0.0.1` path, and to be explicit: Ormah does default to IPv4 today. This came up only during review because the new preflight helper sits in front of uvicorn and uses the configured `ORMAH_HOST`. If a user ever sets `ORMAH_HOST=::1` or another IPv6 literal, uvicorn can normally handle that, but the IPv4-only preflight would fail first. Since the fix is tiny and keeps the guard from being narrower than uvicorn itself, I added it as a small robustness follow-up rather than changing the main approach.

The follow-up switches the probe to `socket.create_connection((host, port), timeout=0.5)`, which lets Python choose the right address family, and adds focused IPv6 regression coverage.

Targeted checks passed locally:

```text
uv run pytest tests/test_server_manager.py tests/test_setup.py::TestPlistTemplate -q
# 10 passed

uv run ruff check src/ormah/server_manager.py tests/test_server_manager.py
# All checks passed
```

Could you take a look and confirm you are happy with this small addition on top of your PR? The main approach and launchd lifecycle fix are unchanged.

AndreLFSMartins:
Reviewed and tested the follow-up commit `62b7fb1` locally.

`socket.create_connection` is the right call here — delegates address-family selection to `getaddrinfo` instead of hardcoding `AF_INET`. All 10 tests pass, including the two new IPv6 cases. Happy with the addition.

r-spade (approved):
LGTM !

PR #37, opened 2026-06-16
fix(background): tolerate markdown-fenced JSON in LLM maintenance jobs

AndreLFSMartins:
## Problem

Thinking-capable local models (e.g. `qwen3.5`) frequently wrap their output in ```` ```json … ``` ```` fences **even when asked for JSON mode**. A naive `json.loads(raw)` then raises `JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, and four background maintenance jobs silently discard a perfectly valid response:

| Job | Site | Effect of a fenced response |
|-----|------|------------------------------|
| `auto_linker` | `_llm_classify_link` | classification dropped → no edge created |
| `conflict_detector` | `_check_conflict` | real conflict missed |
| `duplicate_merger` | `_check_duplicate` | real duplicate missed |
| `consolidator` | `_consolidate_cluster` | **raises** — no `try/except` around the parse |

Observed in the wild from qwen3.5:

```
raw='```json\n{\n  "relationship": "none",\n  "reason": "..."\n}\n```'
→ json.loads → JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The JSON *inside* is valid; only the fences break the parse.

## Fix

Add a shared `extract_json()` to `llm_client`:
1. direct-parse fast path (already-clean JSON passes through),
2. a ```` ``` ````-fence regex,
3. a first-`{`/last-`}` (and `[`/`]`) fallback for surrounding prose.

Route all four jobs through it before `json.loads`. The `raw is None` (LLM unavailable) vs parse-failure handling at each call site is **unchanged** — this only stops valid fenced JSON from being thrown away, so it does not alter when a node is considered resolved/unresolved.

`memory_engine`'s ingestion path already had an equivalent local `_extract_json`; this lifts the same logic into the shared client for the jobs that lacked it (the two can be consolidated in a later cleanup — left out here to keep the change focused and avoid touching the extraction path).

## Verification

- New `tests/test_llm_json.py`: `extract_json` over fenced/bare-fenced/clean/prose inputs, plus `auto_linker._llm_classify_link` recovering a fenced response and still returning `None` on genuinely unparseable output.
- Measured against a local qwen3.5 over real node pairs **with thinking disabled** (so fences are the only failure mode): **5/5** previously-failing classifications now parse, 0 remaining failures.

```
6 passed
```
(Pre-existing unrelated lint in `consolidator.py` lines 7/28 and an environment-dependent `test_config` case are untouched by this change.)

### DIFF
--- a/src/ormah/background/auto_linker.py
+++ b/src/ormah/background/auto_linker.py
@@ -83,7 +83,7 @@ def _llm_classify_link(settings, node_row, other_row) -> dict | None:
     Returns parsed dict with keys relationship, reason — or None if the LLM
     is unavailable or returns invalid output.
     """
-    from ormah.background.llm_client import llm_generate
+    from ormah.background.llm_client import extract_json, llm_generate
 
     def _get(row, key, default="unknown"):
         try:
@@ -107,7 +107,7 @@ def _llm_classify_link(settings, node_row, other_row) -> dict | None:
         return None
 
     try:
-        result = json.loads(raw)
+        result = json.loads(extract_json(raw))
         if "relationship" not in result:
             return None
         result["relationship"] = normalize_link_type(result["relationship"])
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -63,7 +63,7 @@ def _llm_check_conflict(settings, node_row, other_row) -> dict | None:
     Returns parsed dict or None if the LLM is unavailable or returns
     invalid output.
     """
-    from ormah.background.llm_client import llm_generate
+    from ormah.background.llm_client import extract_json, llm_generate
 
     def _get(row, key, default="unknown"):
         try:
@@ -87,7 +87,7 @@ def _llm_check_conflict(settings, node_row, other_row) -> dict | None:
         return None
 
     try:
-        result = json.loads(raw)
+        result = json.loads(extract_json(raw))
         if "conflict" not in result:
             return None
         if "type" in result:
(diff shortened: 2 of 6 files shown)


r-spade:
Thanks Andre — I reproduced the fenced-response failure and confirmed the core fix.

I pushed a small follow-up commit (`a940712`) using the enabled maintainer-edit permission. It hardens the shared extractor by:

- validating normal JSON before returning it,
- validating fenced candidates and accepting case-insensitive `json` labels,
- using `JSONDecoder.raw_decode()` for JSON surrounded by prose while preserving arrays,
- correcting the auto-linker test wording so it reflects the actual retryable behavior.

Verification on the exact PR branch:

- focused JSON/background tests: 38 passed
- full suite: 925 passed, 1 deselected
- Ruff on the changed files: passed

Could you take a quick look at the follow-up commit and confirm the approach? Once you have acknowledged it, this looks ready to merge.

AndreLFSMartins:
Reviewed and tested the follow-up commit `a940712` locally.

The three improvements are correct:
- Validating before returning clean JSON (fixes the false `startswith` fast-path)
- Iterating all fences with `finditer` and validating each candidate (my original `.search()` would have returned the first fence even if invalid)
- `raw_decode()` for prose-wrapped JSON (strictly better than the `rfind` fallback)

All 12 tests pass on the hardened branch. Good to merge.

r-spade (approved):
LGTM ! Thanks for the contribution, really do appreciate it 💯

PR #38, opened 2026-06-17
fix(embeddings): delta backfill (O(gap)) + continuous reconciliation, no more full re-embed (#32)

### LINKED ISSUE #32: perf(startup): a single missing embedding triggers a full ~25-min re-embed of all nodes (swallowed embed failures + O(n) recovery)
## Summary

On every server restart, `MemoryEngine.startup()` re-embeds the **entire** node set (~9k memories, ~25 min via local Ollama, synchronous, blocking the uvicorn port bind) whenever the vector store is missing **even a single** entry. The trigger is a coarse `vec_count < node_count` check; the recovery is an unconditional full re-embed. Combined with an embedding write-path that silently drops failures, this makes a 1-node gap cost a full re-index.

## Root cause — two compounding defects

**1. Leak (creates the gap).** `_index_embedding` swallows encode/upsert failures with no retry or reconciliation:

```python
# src/ormah/engine/memory_engine.py ~L1828
def _index_embedding(self, node):
    try:
        ...
        embedding = encoder.encode(text)
        vec_store.upsert(node.id, embedding)
    except Exception as e:
        logger.warning("Failed to index embedding for node %s: %s", node.id[:8], e)
```

The node row is already persisted, so when Ollama is slow/down/timing-out during a `remember`/ingest (typical for the overnight `session_watcher` batch, or a cold start), the vector is silently never written. Gaps accumulate. Reinforcing factors:
- `builder.full_rebuild` runs `DELETE FROM node_vectors` **without** re-embedding (the builder has no encoder), so any full rebuild zeroes the vectors.
- sqlite-vec `vec0` can silently drop rows in large transactions (already acknowledged in `_reindex_all_embeddings`, which chunks + checkpoints to mitigate).

**2. Amplifier (makes recovery O(n) instead of O(gap)).** `startup()`:

```python
# src/ormah/engine/memory_engine.py ~L122-133
vec_count = self.db.conn.execute("SELECT count(*) FROM node_vectors").fetchone()[0]
...
needs_reindex = (count > 0 and vec_count < count) or stored_version < _EMBEDDING_SCHEMA_VERSION
if needs_reindex:
    self._reindex_all_embeddings()   # re-embeds ALL nodes, regardless of gap size
```

A 1-node gap triggers a full re-embed of all N nodes, synchronously, before the HTTP port binds.

## Evidence (real store, ~9k nodes)

Two real-store re-embeds in one day, from the server log:

| time | vec | nodes | gap |
|------|-----|-------|-----|
| 00:02 | 7469 | 8355 | **886** (Ollama unstable overnight) |
| 18:46 | 9084 | 9085 | **1** (still paid a full ~25-min re-embed) |

The gap varies, so it is not a single fixed orphan node — it is "how many embeds failed/were lost since the last sync". `node_vectors` is `vec0(id TEXT PRIMARY KEY, embedding FLOAT[1024])`; the id↔rowid mapping lives in the shadow table `node_vectors_rowids`, which is what a delta query should join against.

## Proposed fix

**(a) Delta-only recovery (cheap, high impact).** When the reason is "missing entries", embed only the missing rows instead of all:

```sql
SELECT id, title, content FROM nodes
WHERE id NOT IN (SELECT id FROM node_vectors_rowids);
```

Keep the full re-embed path only for an `_EMBEDDING_SCHEMA_VERSION` bump. This bounds startup cost to O(gap): a 1-node gap ≈ 0.1s instead of ~25 min.

**(b) Robustness.** Add a bounded retry/backoff to `_index_embedding`, and/or run the delta backfill as a **post-bind background job** so the server binds the port immediately rather than blocking ~25 min on startup.

## Impact / relation to other issues

- Directly worsens cold-start UX on large stores (the web UI / MCP whisper is unavailable for ~25 min after every restart).
- Lowers the effective `n` health of the vector store, compounding #25 (brute-force vector scan) and #26 (auto_linker O(n²)), which both scale with the embedded set.
- Independent of #28 (bounded forgetting), but a healthy vector store makes pruning measurements cleaner.

Found while bringing a large real store (≈9.1k nodes, ~80% archival) online.

### LINKED ISSUE #109: bug(embeddings): backfill buffers all vectors in memory before persisting — any restart mid-encode loses all progress, freezing the auto_linker watermark
## Summary

`_embed_node_rows` encodes the **entire** batch into an in-memory list before it writes a single vector, so the embedding backfill is not durably incremental. Any hard stop during the encode phase (crash, `launchctl kickstart -k`, KeepAlive respawn, OOM) discards **100% of the run's progress** — the next run starts from zero. On a store where the server restarts at all often, the vector store therefore **never converges**, and because `auto_linker` fail-closes on the first vectorless node above its watermark, the maintenance cursor freezes permanently.

This is the durability gap left open behind #32: #32 fixed *what* to embed (delta anti-join instead of full re-embed), but the *how* still buffers all work in memory and commits only at the very end, so the delta it computes is repeatedly recomputed and never lands.

## Evidence (live store, ~12.1k nodes)

```
node_vectors = 26   (nodes = 12164; 12138 missing)     ← after days of restarts
auto_link_watermark = 333726   (backlog = 12135 nodes)
auto_linker run: {nodes_scanned: 500, pairs_attempted: 0, edges_created: 0,
                  cap_hit: false, backlog_nodes: 12130, duration_s: 0.015}
```

The log shows the backfill climbing — `Embedding backfill (delta): embedding 12138 missing nodes` → `Embedding memories: 10917/12138` — yet `node_vectors` stayed at 26, because the process was killed before it reached the phase-2 upsert. A subsequent restart re-logged `embedding 12138 missing nodes` (started over from scratch), confirming zero progress was persisted.

## Root cause

**Defect 1 — non-resumable encode/persist split (`memory_engine.py::_embed_node_rows`, ~L1295–1321).**

```python
all_items = []
for idx, n in enumerate(nodes):            # PHASE 1: encode ALL nodes into memory (~23 min for 12k)
    if stop_event is not None and stop_event.is_set():
        break
    ...
    all_items.append((n["id"], embedding))
    # logs "Embedding memories: N/total" here — this progress is IN-MEMORY ONLY

chunk_size = 100                            # PHASE 2: only now write to the DB
for i in range(0, len(all_items), chunk_size):
    if stop_event is not None and stop_event.is_set():
        break
    vec_store.upsert_batch(chunk)
    self.db.conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")
```

Nothing is persisted until phase 1 finishes in full. The cooperative `stop_event` only helps a *graceful* shutdown (and even then it persists at most what was already encoded). A **hard** kill — the ValidationError crash-loop, `kickstart -k`, `KeepAlive` respawn, OOM — never runs phase 2, so the whole in-memory `all_items` is lost. `backfill_embeddings` (L1479) delegates straight to this function for both `delta` and `schema` modes, so both are affected.

**Defect 2 — amplifier: `auto_linker` aborts the entire run on the first vectorless node (`auto_linker.py` ~L435–443).**

```python
node_vec = vec_store.get(node["id"]) if text else None
if text and node_vec is None:
    # ponytail: a node that never embeds blocks the watermark forever;
    # acceptable today, revisit if one ever stalls the cursor.
    stopped = True
    break
```

When the vector store is incomplete, the first node above the watermark has no vector, so the run `break`s immediately (0 pairs, ~15 ms) and the watermark never advances. The code comment predicted this exact failure; with Defect 1 keeping the store permanently incomplete, the condition is now permanent, not hypothetical. Net effect: `auto_linker` "executes successfully" every day while doing no work, and 12.1k nodes accumulate zero new edges (semantic recall is also degraded — only 26 vectors indexed).

## Proposed fix

**(a) Make the backfill durably incremental (root fix).** Interleave encode and upsert: encode a chunk of N (e.g. 100), upsert + `wal_checkpoint` that chunk, then move on — instead of buffering the entire batch first. Progress then survives any kill type and each run resumes from the delta anti-join (`node_vectors_rowids` LEFT JOIN) that #32 already added. Bounds worst-case loss to one chunk, not the whole run.

**(b) Stop one vectorless node from freezing the cursor (amplifier fix).** In `auto_linker`, **skip** a vectorless node (leave it unresolved so a later run reprocesses it, but keep advancing past resolved nodes behind it) or embed it on demand, instead of `break`ing the whole run. This decouples "one node lacks a vector" from "the entire linker is dead", so a transient embedding gap degrades gracefully rather than halting maintenance.

(a) is the durability root cause; (b) removes the single-point-of-failure amplification so an incomplete store can't paralyze the linker.

## Relation to other issues

- **#32** — fixed delta *selection* and moved backfill to a post-bind job; this issue is the remaining *persistence durability* gap (the delta is recomputed but never lands on restart).
- **#84** — non-atomic `full_rebuild` can zero `node_vectors`; that is one way the gap gets created, this is why it never closes.
- **#26** — introduced the `auto_linker` watermark; Defect 2's fail-closed abort rides on it, so an incomplete vector store now translates directly into a frozen cursor.

Found while bringing a ~12.1k-node real store back online after a crash-loop: the backfill had been restarting from zero for days and the auto_linker watermark had not moved.

AndreLFSMartins:
# fix(embeddings): delta backfill (O(gap)) + continuous reconciliation — closes #32

## TL;DR

A single embeddable node missing its vector used to trigger a **full ~25-min re-embed of every node** on startup, and embed failures were silently swallowed (so the gap never actually closed). This PR replaces that O(n) recovery with an **incremental delta backfill** (O(gap)), moves recovery **off the startup/bind path** into a background job, and makes a missing vector an **honest, observable degraded state** instead of a silent stall. The backfill worker's lifecycle was then hardened across **6 rounds of peer review** (cooperative cancellation, bounded/symmetric shutdown, atomic singleton, per-lifespan state).

---

## 1. The problem (#32)

From the issue: *"a single missing embedding triggers a full ~25-min re-embed of all nodes (swallowed embed failures + O(n) recovery)."*

Concretely, the old behaviour had three compounding faults:

1. **O(n) recovery** — any missing vector caused a re-embed of the **entire** store, not just the gap. On a large store that is ~25 minutes.
2. **Swallowed failures** — an embed failure during recovery was caught and ignored, so the node stayed missing and the next startup re-triggered the whole pass again (a recurring 25-min tax that never converged).
3. **Blocking startup** — recovery ran on the bind path, delaying readiness.

## 2. The fix

| Area | Change |
|------|--------|
| **Delta backfill** | Embed only the embeddable nodes that lack a vector, via an anti-join (`LEFT JOIN node_vectors_rowids … WHERE v.id IS NULL`). Cost is **O(gap)**, not O(n). |
| **Schema-bump path** | A separate full re-embed runs **only** when the embedding-scheme `embedding_schema_version` changes; the version advances **only after a complete pass**, so an interrupted bump re-runs as a bump next time (never marked done prematurely). |
| **Non-blocking startup** | Startup no longer blocks on a full re-embed; recovery moved to the `embedding_backfill` background job (post-bind first run ~10s after start). |
| **Continuous reconciliation** | The job runs on the scheduler; when the scheduler is unavailable, a **scheduler-independent fallback daemon** heals the gap off-thread with capped exponential backoff (never gives up — recovers when a transient encoder outage clears). This satisfies council finding I1 (recovery must not depend on the scheduler). |
| **Honest observability** | `run_embedding_backfill` raises on `missing > 0` instead of swallowing it → `/admin/health` reports `degraded` (both scheduler and no-scheduler paths); `POST /admin/tasks/run-all` returns **HTTP 503** + per-task results and records outcomes in the `JobTracker`; a permanently-failing ("poison") node stays visibly `missing` and is retried each tick rather than masked. |
| **Bounded write retry** | `_index_embedding` gets a small bounded retry to reduce transient gaps on the write path. |
| **Stats** | `engine.stats()` exposes `embedding_gap` and `embedding_schema_version` for diagnostics. |

### Key code files
- `src/ormah/engine/memory_engine.py` — `backfill_embeddings()` (delta + schema-bump), `_embed_node_rows()` (extracted, returns embedded/failed ids; cooperative cancellation), `_missing_embeddable_count()`.
- `src/ormah/background/embedding_backfill.py` — `run_embedding_backfill()` job runner (raises on incomplete).
- `src/ormah/background/scheduler.py` — registers the `embedding_backfill` job (post-bind first run, `max_instances=1`) and forwards a lifecycle stop event.
- `src/ormah/main.py` — lifespan startup/shutdown: scheduler-independent fallback daemon, bounded & symmetric shutdown, per-lifespan stop event.
- `src/ormah/api/routes_admin.py` — `/admin/health` degraded derivation, `run-all` 503 + JobTracker recording.
- `ui/src/api.ts`, `ui/src/components/AdminPanel.tsx` — AdminPanel surfaces a degraded sleep-cycle instead of false success.

---

## 3. Review history — 6 rounds of peer review

This branch was reviewed adversarially across multiple rounds. The severity dropped monotonically each round (2 criticals → 1 → caveat) until the holistic reviewer reached *Approved with caveats*. Findings and how each was resolved:

| Finding | Issue | Resolution | Commit |
|---------|-------|------------|--------|
| **C1/I1** | `run-all` reported success even when a task failed (operability masking) | `run-all` → `status=degraded` + HTTP 503 | `0544283` |
| **C2/CH1/CH2** | Fallback gave up after a fixed budget; no degraded signal without a scheduler | Lifecycle-aware fallback (infinite capped-backoff retry) + `_fallback_degraded` health flag | `f1fceb7` |
| **I2** | UI showed false success on a degraded sleep-cycle | AdminPanel tolerates 503 / names failed tasks | `512e1f2` |
| **CR2** | `/admin/health` stayed `ok` when the scheduler's `embedding_backfill` last run failed | Derive `degraded` from `JobTracker.snapshot()` | `b1f6592` |
| **CRA / IA** | Backfill was non-interruptible → `db.close()` raced a live worker (use-after-close); orphan thread blocked a new fallback | **Cooperative cancellation**: `stop_event` checked between encodes and **before every DB write** | `bb5830f` |
| **CRB** | Singleton `check-then-set` was not atomic | Module `threading.Lock` around check-and-set; reverted the band-aid CR1 (`2062f24`) | `b9b13ab` |
| **CRC** | A failed task in manual `run-all` never persisted to health | `run-all` records `record_success`/`record_failure` in the `JobTracker` | `96525b5` |
| **C-A / C-B** | A timed-out join still cleared the handle and let `engine.shutdown()` run with the worker alive (use-after-close + zombie singleton) | `_stop_backfill_fallback` joins, and the engine is closed only once the worker is confirmed dead | `132f828` |
| **C1/C2 (r4)** | The fix above could hang shutdown forever; the **scheduler** path had no cancellation at all | Forward a lifecycle `stop_event` to the scheduler job; cancel cooperatively on shutdown | `6d54746` |
| **A/B/D (r5)** | Scheduler shutdown still unbounded; schema-mode `DELETE` ignored the interrupt flag; no lifespan integration test | **Bounded, symmetric shutdown** (`_bounded_scheduler_shutdown` + `_should_close_engine`); guard `DELETE` with `not interrupted`; add `test_main_lifespan_shutdown.py` | `b02d98d` |
| **R1 (r6)** | A module-global stop event `clear()` on startup could *rearm* an orphan worker in a same-process reload | **Per-lifespan stop event** in `app.state` (no shared global to clear) | `59b57ba` |

---

## 4. Design decisions (the two that mattered)

**(a) Make the backfill cancellable — not remove the daemon.** When the lifecycle bugs clustered, one option was to delete the fallback daemon entirely (synchronous backfill at startup). It was rejected because it reopens council finding **I1** (continuous, scheduler-independent recovery) and only heals once at boot. Instead the backfill was made **cooperatively cancellable**, which collapses the use-after-close/zombie class while keeping continuous recovery.

**(b) Resolve "corruption vs hang" explicitly with a bounded, fail-safe shutdown.** A non-interruptible `encoder.encode()` means a join can't be instantaneous. Rather than choose between *use-after-close* (close the DB under a live worker) and *infinite hang* (join forever), shutdown now:
- signals the lifecycle stop event,
- **time-bounds** both the fallback join and the scheduler shutdown (`_SHUTDOWN_TIMEOUT`),
- and **skips `engine.shutdown()` entirely if a worker is still in flight** (`_should_close_engine`) — no DB close under a live worker, and no unbounded hang.

A background "finalizer" that closes the engine after the worker eventually exits was **deliberately not added**: it reintroduces an old-engine-vs-new-engine race that is worse than the benign leak it would fix.

---

## 5. Known limitation (documented, accepted)

A single `encoder.encode()` call is **not interruptible** (a fundamental limit; fixing it would require running the encoder in a killable subprocess). Consequence, only in a **same-process reload** (TestClient / `uvicorn --reload` in-process — **not** the production deploy path, which is SIGTERM → process exit → daemon threads die):
- a still-running orphan worker can briefly suppress a replacement fallback for the new engine, or
- the engine connection is left open until process exit (a benign leak).

**No data corruption** in any path, and production (SIGTERM) is unaffected. Tracked as a follow-up rather than blocking this PR.

---

## 6. Testing

- **Full suite green** except 4 pre-existing reds **unrelated** to this PR (present on `main`): `test_config.py::test_llm_provider_defaults_to_none` and `test_setup.py::TestRemoveFastembedCache` ×3.
- New coverage: delta vs schema-bump backfill, poison-node visibility, schema-version-not-advanced-on-interrupt, cooperative cancel before DB writes, atomic singleton under concurrent starts, bounded join (fallback) + bounded scheduler shutdown, `_should_close_engine` policy, per-lifespan stop event, run-all → health degraded, and an end-to-end recovery test.

## 7. Notes for the reviewer
- Rebased cleanly onto `main` (includes #35); no conflicts.
- The diff includes spec/plan docs under `docs/superpowers/` (design + the council review trail) — these are documentation, not runtime code. One unrelated `docs(spec)` commit for the graph-view sigma.js migration (`4174db0`) rode along on this branch; happy to drop it if you'd prefer the PR scoped to code + #32 docs only.

---
Also closes #109 — the restart-resumable backfill on this branch persists each 100-row chunk as it is encoded (`_embed_node_rows`: `chunk_size = 100`, per-chunk `_flush()` with `upsert_batch` + `wal_checkpoint(TRUNCATE)`), so a mid-encode restart loses at most the current chunk instead of all buffered vectors. Verified: `pytest tests/test_engine/test_embed_node_rows.py` → 6 passed, incl. `test_embed_node_rows_persists_incrementally` and `test_embed_node_rows_flushes_pending_on_cooperative_cancel`.

Closes #109

### DIFF
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -111,6 +111,14 @@ def health(request: Request):
     result: dict = {"status": "ok"}
     if tracker is not None:
         result["jobs"] = tracker.snapshot()
+    else:
+        # No scheduler: the backfill fallback is the only healer. Surface a
+        # persistent outage that would otherwise be invisible (council CH2).
+        from ormah import main as _main
+
+        if getattr(_main, "_fallback_degraded", False):
+            result["status"] = "degraded"
+            result["embedding_backfill"] = "degraded: fallback retrying"
     manager = getattr(request.app.state, "maintenance_manager", None)
     if manager is not None:
         result["maintenance"] = manager.get_status()
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -110,7 +110,18 @@ def health(request: Request):
     tracker = getattr(request.app.state, "job_tracker", None)
     result: dict = {"status": "ok"}
     if tracker is not None:
-        result["jobs"] = tracker.snapshot()
+        snap = tracker.snapshot()
+        result["jobs"] = snap
+        # CR2: promote to degraded when embedding_backfill's most recent run
+        # failed — otherwise a scheduler-backed outage stays invisible to callers
+        # that only read the top-level status.
+        job = snap.get("embedding_backfill")
+        if job and job.get("last_error"):
+            ls = job.get("last_success")
+            let = job.get("last_error_time")
+            if ls is None or (let is not None and let > ls):
+                result["status"] = "degraded"
+                result["embedding_backfill"] = "degraded: last run failed"
     else:
         # No scheduler: the backfill fallback is the only healer. Surface a
         # persistent outage that would otherwise be invisible (council CH2).
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from pathlib import Path
 
 from fastapi import APIRouter, HTTPException, Request
+from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
 
 router = APIRouter(prefix="/admin", tags=["admin"])
@@ -293,4 +294,12 @@ def run_all_tasks(request: Request):
         except Exception as exc:
             results[task_id] = f"error: {exc}"
 
+    has_errors = any(v.startswith("error:") for v in results.values())
+    if has_errors:
+        # I1: a partial failure must not look like success to HTTP-only callers
+        # (cron, scripts). Surface it as 503 while keeping the per-task body.
+        return JSONResponse(
+            status_code=503,
+            content={"status": "degraded", "results": results},
+        )
     return {"status": "completed", "results": results}
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -28,6 +28,7 @@ _TASK_RUNNERS = {
     "memory_backup": ("ormah.backup", "run_auto_backup"),
     "cloud_backup": ("ormah.cloud.jobs", "run_cloud_backup"),
     "restore_verification": ("ormah.cloud.jobs", "run_restore_verification"),
+    "embedding_backfill": ("ormah.background.embedding_backfill", "run_embedding_backfill"),
 }
 
 _TASK_DESCRIPTIONS = {
@@ -43,12 +44,14 @@ _TASK_DESCRIPTIONS = {
     "memory_backup": "Creates a local backup of memory source files when one is due.",
     "cloud_backup": "Encrypts and uploads a due cloud backup without changing the sync head.",
     "restore_verification": "Downloads, decrypts, rebuilds, and searches the latest cloud backup.",
+    "embedding_backfill": "Backfills missing vector embeddings (delta) or re-embeds all on an embedding-schema bump. Keeps vector search complete after restarts and overnight ingest (#32).",
 }
 
 # Order for sleep cycle (full maintenance pass)
 _SLEEP_CYCLE_ORDER = [
     "importance_scorer",
     "index_updater",
+    "embedding_backfill",
     "duplicate_merger",
     "conflict_detector",
     "auto_linker",
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -296,11 +296,14 @@ def run_task(task_id: str, request: Request):
 def run_all_tasks(request: Request):
     """Run all background tasks sequentially in sleep-cycle order."""
     import importlib
+    import time as _time
 
     engine = request.app.state.engine
+    tracker = getattr(request.app.state, "job_tracker", None)
     results: dict[str, str] = {}
 
     for task_id in _SLEEP_CYCLE_ORDER:
+        t0 = _time.monotonic()
         try:
             if task_id == "index_updater":
                 engine.builder.incremental_update()
@@ -310,8 +313,14 @@ def run_all_tasks(request: Request):
                 runner = getattr(module, func_name)
                 runner(engine)
             results[task_id] = "ok"
+            if tracker is not None:
+                tracker.record_success(task_id, (_time.monotonic() - t0) * 1000)
         except Exception as exc:
             results[task_id] = f"error: {exc}"
+            # CRC: persist the failure so /admin/health reflects this manual run
+            # instead of reverting to a stale ok.
+            if tracker is not None:
+                tracker.record_failure(task_id, str(exc), (_time.monotonic() - t0) * 1000)
 
     has_errors = any(v.startswith("error:") for v in results.values())
     if has_errors:
--- a/src/ormah/background/embedding_backfill.py
+++ b/src/ormah/background/embedding_backfill.py
@@ -1,0 +1,30 @@
+"""Vector-store reconciliation job: backfill missing embeddings (#32)."""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def run_embedding_backfill(engine) -> None:
+    """Reconcile the vector store. Raises if the store is left incomplete.
+
+    Unlike the other background jobs, this one does NOT swallow an incomplete
+    result: when the run ends with ``missing > 0`` (embeddable nodes still lack a
+    vector) it raises so ``tracked()`` records a job failure and ``/admin/health``
+    reflects the degradation -- the intended health signal. A permanently-failing
+    ("poison") node therefore stays visibly degraded instead of being masked.
+    """
+    result = engine.backfill_embeddings()
+    if result.get("embedded") or result.get("missing"):
+        logger.info(
+            "Embedding backfill (%s): embedded=%d failed=%d missing=%d vec=%d/%d",
+            result["mode"], result["embedded"], result["failed"], result["missing"],
+            result["vec_count"], result["node_count"],
+        )
+    if result.get("missing", 0) > 0:
+        raise RuntimeError(
+            f"Embedding backfill incomplete: {result['missing']} embeddable nodes "
+            f"still missing vectors (failed={result['failed']})"
+        )
--- a/src/ormah/background/embedding_backfill.py
+++ b/src/ormah/background/embedding_backfill.py
@@ -7,7 +7,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-def run_embedding_backfill(engine) -> None:
+def run_embedding_backfill(engine, stop_event=None) -> None:
     """Reconcile the vector store. Raises if the store is left incomplete.
 
     Unlike the other background jobs, this one does NOT swallow an incomplete
@@ -15,8 +15,12 @@ def run_embedding_backfill(engine) -> None:
     vector) it raises so ``tracked()`` records a job failure and ``/admin/health``
     reflects the degradation -- the intended health signal. A permanently-failing
     ("poison") node therefore stays visibly degraded instead of being masked.
+
+    If ``stop_event`` is provided and becomes set, the backfill exits early.
+    An interrupted run still raises if ``missing > 0`` — the incomplete state
+    is reported honestly so the caller knows to reschedule.
     """
-    result = engine.backfill_embeddings()
+    result = engine.backfill_embeddings(stop_event=stop_event)
     if result.get("embedded") or result.get("missing"):
         logger.info(
             "Embedding backfill (%s): embedded=%d failed=%d missing=%d vec=%d/%d",
--- a/src/ormah/background/scheduler.py
+++ b/src/ormah/background/scheduler.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 import logging
 
 from apscheduler.schedulers.background import BackgroundScheduler
@@ -155,6 +155,24 @@ def start_scheduler(engine: MemoryEngine) -> tuple[BackgroundScheduler, JobTrack
         id="restore_verification",
         name="Cloud restore verification",
         misfire_grace_time=_MISFIRE_GRACE,
+    )
+
+    from ormah.background.embedding_backfill import run_embedding_backfill
+
+    # Reconcile the vector store (delta + schema bump). A post-bind first run
+    # (~10s after start) guarantees recovery right after the port binds even when
+    # the operator sets the interval to a very large value (999999) to rely on the
+    # 02:00 sleep-cycle instead. max_instances=1: a slow run on a large gap must
+    # never overlap the next tick.
+    scheduler.add_job(
+        tracked(tracker, "embedding_backfill", run_embedding_backfill, engine),
+        "interval",
+        minutes=s.embedding_backfill_interval_minutes,
+        id="embedding_backfill",
+        name="Embedding backfill",
+        misfire_grace_time=_MISFIRE_GRACE,
+        max_instances=1,
+        next_run_time=datetime.now(timezone.utc) + timedelta(seconds=10),
     )
 
     scheduler.start()
--- a/src/ormah/background/scheduler.py
+++ b/src/ormah/background/scheduler.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 from datetime import datetime, timedelta, timezone
 import logging
+import threading
+from typing import Optional
 
 from apscheduler.schedulers.background import BackgroundScheduler
 
@@ -17,11 +19,17 @@ logger = logging.getLogger(__name__)
 _MISFIRE_GRACE = 120
 
 
-def start_scheduler(engine: MemoryEngine) -> tuple[BackgroundScheduler, JobTracker]:
+def start_scheduler(
+    engine: MemoryEngine,
+    stop_event: Optional[threading.Event] = None,
+) -> tuple[BackgroundScheduler, JobTracker]:
     """Register and start all background jobs.
 
     Returns ``(scheduler, tracker)`` so the caller can inspect job health
     via ``tracker.snapshot()``.
+
+    ``stop_event``, when provided, is forwarded to the embedding_backfill job so
+    it can cancel cooperatively between encodes on scheduler shutdown.
     """
     scheduler = BackgroundScheduler()
     tracker = JobTracker()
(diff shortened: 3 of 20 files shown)


AndreLFSMartins:
Force-pushed a cleanup commit: removed the 26 tracked `docs/superpowers/` files (local agent-tooling artifacts, not for upstream) from this branch and added `docs/superpowers/`, `.superpowers/`, `graphify-out/` to `.gitignore` so they can't creep back in — same fix you applied on #30 via `7e24151`. Diff vs `main` is now clean of those artifacts. No source changes.

AndreLFSMartins:
Rebased onto main @ 2e76b5b. Dropped the branch's local tooling-doc commits (docs/superpowers plans + rich-doc HTML companions + .council review artifacts — gitignored, not upstream-appropriate); kept the 23 code/test commits for #32. Resolved conflicts in test_config.py (both sides add independent tests), memory_engine.py (upstream nested stats() under 'store' — kept that, re-exposed embedding_gap/schema_version/vec_count at top level per the #32 observability contract, see the added reconciliation commit). Local suite green (same 11 environmental failures as clean main; 0 new).

AndreLFSMartins:
Pushed two commits (`e9a84c6`, `150d4a8`) that close a durability hole in the backfill this PR introduces. Found by running it for real on a ~12.1k-node store that had just been wiped (see #115) — the recovery itself was not restart-safe.

## The hole

`_embed_node_rows` is a two-phase loop: it encodes **every** node into an in-memory list, and only then calls `vec_store.upsert_batch(all_items)` once, at the end.

```python
all_items = []
for n in nodes:
    ...
    all_items.append((n["id"], embedding))   # nothing persisted yet
...
vec_store.upsert_batch(all_items)            # single write, after the last encode
```

So the whole run is all-or-nothing. On my store a full re-embed takes ~5 minutes; any restart, `launchctl kickstart`, crash or OOM inside that window discards **100%** of the work and the next run starts from zero. I lost a run at 10917/12138 (~90%) to exactly this.

It also interacts badly with the cooperative cancel added in `bcf4d10`: `stop_event` breaks out of the loop *before* the single write, so a clean, intentional shutdown is the one path guaranteed to persist nothing.

## What changed

**`e9a84c6`** — interleave encode and persist in chunks of 100. `_flush()` upserts the pending chunk, `PRAGMA wal_checkpoint(TRUNCATE)`, clears the buffer, and records the ids. A final `_flush()` after the loop covers the partial tail — so it runs on natural completion **and** on cooperative cancel. Work already done survives.

One placement detail that matters: `_flush()` sits in the `else` of the encode `try`, never inside it.

```python
try:
    embedding = encoder.encode(text)
except Exception as e:
    logger.warning("Failed to embed node %s: %s", n["id"][:8], e)
    failed_ids.append(n["id"])
else:
    pending.append((n["id"], embedding))
    if len(pending) >= chunk_size:
        _flush()          # outside the encode try — DB errors propagate, not misclassified
```

If `_flush()` were inside the `try`, a **DB** failure would be caught by the encoder's `except`, the node would be filed under `failed_ids`, `pending` would never clear, and the schema-mode cleanup would go on to DELETE a vector that was in fact persisted. It must fail loudly instead.

**`150d4a8`** — the regression test for that. The first version was toothless: it used 1 node, so the in-loop `_flush()` never fired and the test passed even with the bug reintroduced. It now uses 101 nodes (forcing exactly one in-loop flush) and asserts `calls["n"] == 1` — the upsert must be attempted once and the exception must propagate. I verified it by injecting the regression: `assert 4 == 1` fails.

## Verification

Same command on both sides, cloud suites excluded (missing optional deps in my env):

| | failed | passed |
|---|---|---|
| base `7db6330` | 50 | 1192 |
| + these 2 commits | 50 | 1195 |

The two failure sets are **identical** (`comm` on the sorted `FAILED` lines returns nothing on either side) — the 50 are pre-existing and environment-dependent (`test_setup*`, `test_config`, and a global `~/.config/ormah/.env` leaking into bare `Settings()`), and the delta is exactly the 3 new tests. `ruff check src/ tests/`: same 2 pre-existing errors before and after, none in the touched files.

Targeted suites: `tests/test_engine/test_embed_node_rows.py`, `tests/test_background/test_embedding_backfill.py` → 18 passed.

## Not addressed here

Schema-mode backfill still has no per-row progress tracking, so an interrupted schema pass re-encodes every row on the next run. Idempotent — wasted compute, no data loss. Filed under #109.

AndreLFSMartins:

I have to correct the verification table in my comment above. The conclusion survives, but the evidence I gave for it did not actually test what I claimed it tested.

That means both sides of my "base vs. patched" comparison executed the **same** source. The identical failure sets I presented as evidence of zero regressions were an artifact of the bug — they were guaranteed to be identical, and would have been identical even if these commits were badly broken. I caught it only because a test from a newer branch failed with an `AttributeError` on a model field that plainly existed in the file I was looking at.

| | failed | passed |
|---|---|---|
| base `7db6330` | 12 | 1230 |
| + `e9a84c6`, `150d4a8` | 12 | 1233 |

Failure sets identical in both directions (`comm` on the sorted `FAILED` lines returns nothing either way), delta is exactly the 3 new tests, and all 12 pre-existing failures are environment-dependent (`test_setup*`, `test_setup_json`, plus `test_config`/`test_session_watcher` reading a global `~/.config/ormah/.env`). **Zero regressions** — same conclusion as before, now actually measured.

One honest footnote: the *first* correct run of the patched side showed 13 failures, the extra one being `tests/test_background/test_hippocampus.py::test_new_file_triggers_ingestion`. It passes 3/3 in isolation on **both** base and patched, and did not reappear on a second full run of the patched side. It is an order/timing-dependent flake in a watchdog test, not something these commits caused — but it is worth knowing it exists.

The earlier figures (50 failed / 1192 vs 1195) should be disregarded.

r-spade (commented):
Thanks Andre. I reviewed this against current `main`, and I agree with the core direction: moving embedding recovery off startup and making the normal repair path O(gap) is the right fix for #32. The later chunked-persistence change also looks important; without it, a long backfill can make progress in logs while still losing everything on restart.

I pushed one tiny maintainer commit on top (`1faa0fa`) to clear the two Ruff failures:

- moved `import threading` into the normal import block
- removed an unused `result =` assignment in `test_schema_mode_does_not_delete_when_interrupted`

Verification on my side:

- `uv run ruff check src/ tests/` -> passed
- `ORMAH_LLM_PROVIDER=none uv run pytest tests/test_engine/test_backfill_embeddings.py tests/test_main_lifespan_shutdown.py -q` -> 18 passed
- earlier focused review on an auto-merged current-main worktree: 71 focused tests passed, and `npm run build` passed

The remaining points below are non-blocking from my side, but worth keeping explicit:

1. I think the scheduled `embedding_backfill` job is justified: gaps can be created after startup by transient encoder/write failures, rebuilds, or ingestion, and they should heal without waiting for a restart.

2. The scheduler-independent fallback thread is the part I had to think about more. It is not another OS-level Ormah daemon, but it is another in-process background worker path with meaningful lifecycle complexity. I can see the failure mode it protects against (scheduler unavailable while the server is otherwise up), and the added lifecycle tests make the current implementation defensible. If we keep it, we should consider that an intentional reliability tradeoff rather than incidental plumbing.

3. I would not close all of #109 with this PR unless we are deliberately scoping #109 down to the backfill durability piece. This PR fixes the root durability issue by persisting chunks as they are encoded. The auto-linker watermark amplifier described in #109 still looks like a separate follow-up.

4. Stale vectors remain a follow-up: if an existing node is updated/merged and `_index_embedding()` fails after retries, the old vector can remain while the node text has changed. That is invisible to the missing-vector anti-join. Now that delta backfill is cheap, a future fix could delete the vector on final failure, or store a vector text hash/revision so stale vectors are detectable.

Overall: after the lint fix, I am comfortable with the core #32 fix. I’d keep the remaining items as follow-ups / scope clarifications rather than blockers.

PR #40, opened 2026-06-23
Add feedback loop for whispered memories

### LINKED ISSUE #21: Adaptive feedback loop is effectively never closed: `submit_feedback` has no automated trigger, so affinity learning receives ~0 signal
## Summary

Ormah ships a complete **adaptive feedback loop** (whisper injection → `whisper_log` →
`submit_feedback` → `affinity` table → `compute_affinity_boost` re-ranks future recall).
Three of the four stages work and are well-instrumented. The **bridge stage —
`submit_feedback` — has no automated trigger whatsoever.** Its only caller is a *soft
instruction* asking the agent to voluntarily call the MCP tool when it judges a whispered
memory was used/ignored.

In practice agents almost never do this. On a real instance the loop is **~99% open**:
the affinity table stays nearly empty, `compute_affinity_boost` returns ≈0 for almost every
candidate, and the "learn which memories to surface over time" capability is dead **not
because the code is wrong, but because it is never fed**.

This also means **outcome/usefulness of whispering is completely uninstrumented**: the
system can prove it *stores* memories and keeps the graph clean (hygiene jobs run), but it
cannot show that injection *helps* — there is no closed signal to measure precision against.

- **Version:** 0.10.1
- **Surfaces affected:** whisper injection (Claude Code plugin + any MCP/HTTP client)
- **Severity:** high (a headline feature — adaptive, self-improving retrieval — is inert in
  normal operation, and its effectiveness is unmeasurable)

---

## Expected vs. actual

**Expected:** as memories are whispered and used, the system accumulates relevance signal,
and recall ranking adapts so that helpful memories surface more and noisy ones surface less
("the whisper learns").

**Actual:** the signal-collection step depends entirely on an LLM voluntarily calling
`submit_feedback`. With realistic agents this fires on a tiny fraction of injections, so the
affinity table never accumulates enough rows to move ranking. The loop is open by
construction.

---

### Stage 1 — Injection: works, fully instrumented
- Claude Code plugin hook `UserPromptSubmit` → `ormah whisper inject`
  (`integrations/claude-plugin/hooks/hooks.json:3-13`).
- Every surfaced candidate is logged to `whisper_log`
  (`src/ormah/engine/memory_engine.py:491-527`), including `session_id`, `prompt_text`,
  `prompt_vec` (BLOB), `node_id`, `score`, `was_injected`, `logged_at`.
- Table defined in `src/ormah/index/db.py:129`.
- **Key point: all the data needed to later attribute feedback already exists** — including
  the prompt embedding at injection time.

### Stage 2 — Feedback bridge: never triggered automatically
- `submit_feedback(node_id, signal, source)` (`src/ormah/engine/memory_engine.py:2052`)
  looks up the most recent `whisper_log` row for the node, then inserts a row into
  `affinity` (`memory_engine.py:2082-2091`). **This is the only code path that creates
  affinity rows.**
- The only things that call it are *instructions to the model*, not code:
  - `src/ormah/instructions.md:29` (item 12) — "If a whispered memory is genuinely useful and
    you actively draw on it… call `submit_feedback(..., source="implicit")`…"
  - `src/ormah/codex_instructions.md:29` — identical instruction for Codex.
  - Inline nudge in `src/ormah/engine/context_builder.py:30`.
  - Tool schema description `src/ormah/adapters/tool_schemas.py:176-208`.
- There is **no `Stop` / post-turn / end-of-turn hook** anywhere. `hooks.json` only wires:
  `UserPromptSubmit` → inject, `PreCompact` → store, `SessionEnd` → store.
- The `bin/` directory has only `ormah-mcp`, `ormah-whisper-inject`, `ormah-whisper-store`
  — there is no feedback executable.

### Stage 3 — Affinity storage: works
- `affinity` table defined in `src/ormah/index/db.py:156`.
- One row per signal: `prompt_vec`, `prompt_text`, `node_id`, `signal` (+1/-1), `source`
  (`explicit`/`implicit`), `confirmed_at`, `space`, `session_id`.

### Stage 4 — Ranking adaptation: works IF fed
- `src/ormah/engine/affinity.py:39` `compute_affinity_boost` weights each affinity row by
  `cosine(current_prompt_vec, stored_prompt_vec) × recency_decay × source_weight`, sums the
  signed signals, normalises, and scales by `affinity_max_boost`. Returns 0.0 when there are
  no rows above the similarity threshold (`affinity.py:59-60, 109-110`).
- Relevant defaults (`src/ormah/config.py:177-180`):
  - `affinity_similarity_threshold = 0.70`
  - `affinity_half_life_days = 30.0`
  - `affinity_max_boost = 0.15`
  - `affinity_implicit_weight = 0.8`

**Net effect:** stages 1, 3, 4 are sound. The pipeline produces a 0 boost in practice purely
because stage 2 receives almost no input.

---

## Root cause

The feedback-collection mechanism is **soft LLM compliance**, the least reliable trigger
possible:

> the loop only advances if an LLM remembers, on every turn, to call an MCP tool for each
> whispered node it did or didn't use.

There is no deterministic event (e.g. a `Stop` hook) that closes the loop, even though all
the attribution data (`whisper_log` keyed by `session_id` + prompt vector) is already
persisted for exactly this purpose (see the docstring at `memory_engine.py:491`: *"Log
surfaced memories so submit_feedback can learn from them later."*).

---

## Evidence (observed on one production instance, v0.10.1)

These counts are from a single real deployment, not synthetic. They illustrate the gap:

| Metric | Value |
| --- | --- |
| Memory nodes | 1,733 |
| Whisper events logged | 2,574 |
| Injected (passed the gate) | 2,118 (≈82% injection rate) |
| Distinct nodes ever shown | 615 (≈35% of the graph) |
| **`submit_feedback` rows total** | **15** |
| → positive / negative | 11 / 4 |
| → all `source` | implicit |
| Sessions that produced any feedback | 11 of 674 |

**Feedback coverage ≈ 15 / 2,118 = 0.7%.** With `affinity_similarity_threshold = 0.70` and
`affinity_implicit_weight = 0.8`, 15 down-weighted rows spread across 1,733 nodes produce a
boost indistinguishable from zero on essentially every query.

(Maintainers can reproduce on any instance with:
`SELECT (SELECT COUNT(*) FROM affinity) AS feedback, (SELECT COUNT(*) FROM whisper_log WHERE was_injected=1) AS injected;`)

---

## Why it matters

1. **Adaptive retrieval is inert.** The "memory that learns what to surface" property does
   not function in normal use.
2. **Effectiveness is unmeasurable.** With ~0 feedback there is no precision signal
   (`+1 / (+1 + -1)`), so there is no way to tell whether whispering helps, is neutral, or
   adds noise — despite injecting on ~82% of turns.
3. **Asymmetric risk.** High injection rate + no negative signal means the system can be
   degrading context on some turns and would never detect it.

---

## Secondary issues found while tracing this

1. **The hard part isn't wiring — it's the judgment.** A `Stop` hook can cheaply know *which*
   nodes were injected this turn (`whisper_log` by `session_id`), but deciding *used vs.
   ignored* requires judging the agent's response. A purely lexical heuristic can detect a
   likely `+1` but cannot reliably produce a `-1` (silence ≠ irrelevance). Honest negative
   signal probably needs a small LLM judge over `{injected memories, prompt, response}`.

2. **`+1`/`-1` asymmetry biases toward inflation.** A heuristic-only auto-feedback that only
   emits `+1` would entrench already-surfaced memories and never prune noise.

3. **One signal per `(node_id, session_id)`.** `submit_feedback` inserts with
   `ON CONFLICT (node_id, session_id) DO NOTHING` (`memory_engine.py:2088`). A per-turn
   auto-feedback hook firing repeatedly within a session would silently drop all but the
   first signal for a node. The conflict key likely needs to become `(node_id, prompt_hash)`
   or `(node_id, whisper_log_id)` for turn-level feedback to accumulate.

4. **"Feedback" ≠ "learning from errors" and ≠ "model learning".** Worth documenting clearly:
   - the signal is *relevance for retrieval*, not *correctness* (wrong memories go through
     `mark_outdated`, a separate and also-manual path — there is no error detector feeding
     either);
   - affinity adapts ormah's *retrieval ranking*, not the underlying model. Even with a fully
     closed loop, what improves is *which memory surfaces*, not the model's reasoning.

5. **Narrow transfer.** Because the boost is gated on `cosine ≥ 0.70` against the *prompt*
   vector and capped at `±0.15`, a single `+1` only helps near-duplicate future prompts, not
   the node globally. This is defensible, but combined with sparse feedback it means even the
   rare collected signal rarely transfers.

---

## Proposed solutions (loop-closing options, in increasing cost/fidelity)

**A. Deterministic heuristic `Stop` hook.** New `Stop` entry in `hooks.json` → new
`bin/ormah-whisper-feedback` → new CLI subcommand. At turn end, for each node injected this
session/turn (`whisper_log`), compute lexical/embedding overlap between the node content and
the agent's final message; emit `+1` on strong overlap. Cheap, no extra model call, but
effectively `+1`-only.

**B. LLM-judge `Stop` hook.** Same wiring, but the hook calls a small/cheap model with
`{injected memories, prompt, response}` and asks for a per-node verdict (used `+1` /
clearly-irrelevant `-1` / abstain). Closes both sides of the loop at the cost of one cheap
call per turn.

**C. Hybrid.** Heuristic for the obvious `+1`, LLM judge only on the ambiguous remainder.

All three also require fixing the `ON CONFLICT` key (secondary issue #3) so turn-level signal
can accumulate, and should `log()`/surface any nodes dropped so coverage isn't silently
capped.

Independently of which option is chosen, exposing a **feedback-coverage / whisper-precision
metric** (`affinity rows ÷ injected`, and `+1 / (+1+-1)`) would make the loop's health
observable and let injection thresholds be tuned against real outcomes.

---

## Files involved

| File | Role |
| --- | --- |
| `integrations/claude-plugin/hooks/hooks.json` | hook wiring — no `Stop`/feedback hook today |
| `integrations/claude-plugin/bin/` | `ormah-mcp`, `ormah-whisper-inject`, `ormah-whisper-store` — no feedback bin |
| `src/ormah/engine/memory_engine.py:491` | logs surfaced memories to `whisper_log` |
| `src/ormah/engine/memory_engine.py:2052` | `submit_feedback` — only writer of `affinity` |
| `src/ormah/engine/affinity.py:39` | `compute_affinity_boost` — consumer (returns 0 when unfed) |
| `src/ormah/index/db.py:129,156,174` | `whisper_log`, `affinity`, `review_log` schemas |
| `src/ormah/config.py:177-180` | affinity tuning defaults |
| `src/ormah/instructions.md:29` / `codex_instructions.md:29` | the soft instruction that is the only trigger |
| `src/ormah/adapters/tool_schemas.py:176` | `submit_feedback` tool schema |

## Open questions for maintainers

- Was the implicit-feedback step always intended to be model-driven, or was an automated
  trigger planned and not yet wired?
- Is a small LLM judge in a `Stop` hook acceptable (latency/cost) for the reference Claude
  Code plugin, or should the default ship as heuristic-only with the judge opt-in?
- Should `affinity`'s conflict key change to allow per-turn accumulation, or is one
  signal per `(node, session)` deliberate?

r-spade:
## Context

Closes #21.

The issue was that Ormah had an affinity system, but the feedback loop was effectively open-ended: `submit_feedback(...)` could write useful signal, but there was no automatic trigger that converted normal agent sessions into feedback. In practice, most whispered memories were surfaced, used or ignored by the assistant, and then forgotten by the learning loop unless someone explicitly submitted feedback.

This PR closes that loop for watched agent transcript files after they stop changing for the
configured debounce window, while keeping the default behavior conservative. It is not tied
to a first-class "session closed" event from the agent.

### 1. Transcript watcher now records whisper usage signals

When the session watcher processes a debounced transcript file, it looks up injected `whisper_log` rows for that session and compares each logged user prompt with the assistant response that followed it.

It now records signal rows such as:

- `whisper_referenced`
- `whisper_unreferenced`
- `whisper_judged_used`
- `whisper_judged_irrelevant`
- `whisper_judged_uncertain`

Why this is needed:

The signals ledger gives us a durable audit trail before changing affinity. That matters because not every observation should immediately affect ranking. For example, "assistant did not visibly mention the memory" is useful telemetry, but it is not safe negative feedback by itself.

### 2. Local heuristics provide the free baseline

The heuristic path looks only for clear positive evidence that the assistant used a memory. It checks for:

- short node id references
- title phrase matches
- copied or paraphrased memory sentences
- distinctive token overlap after subtracting tokens already present in the user prompt

Why this is justified:

This gives Ormah a zero-cost local path for obvious positive references. It also avoids the dangerous case: inferring negative feedback from silence. If the assistant did not mention a memory, the memory might still have helped frame the answer, so the heuristic only writes neutral observations for non-references.

### 3. Affinity writes are tied back to `whisper_log`

Promoted feedback writes affinity rows with the original prompt context from `whisper_log`, including:

- `prompt_vec`
- `prompt_text`
- `space`
- `session_id`
- `whisper_log_id`

Why this is needed:

Affinity is only meaningful relative to the prompt that caused the memory to be surfaced. Storing feedback against `whisper_log_id` keeps the learning turn-level rather than treating a memory as globally good or bad.

### 4. Optional LLM judge handles ambiguous cases

The LLM judge is disabled by default. When enabled, it only receives rows that were not clear heuristic positives.

Config:

```bash
ORMAH_FEEDBACK_LLM_JUDGE_ENABLED=true
ORMAH_FEEDBACK_LLM_JUDGE_MIN_CONFIDENCE=0.75
ORMAH_LLM_PROVIDER=litellm
ORMAH_LLM_MODEL=<model>
```

The judge returns:

- `used`
- `irrelevant`
- `uncertain`

High-confidence `used` becomes positive affinity. High-confidence `irrelevant` becomes negative affinity. `uncertain` and low-confidence outputs remain neutral signal rows only.

Why this is justified:

Negative feedback is higher risk than positive feedback. The LLM judge has access to the user prompt, assistant response, and injected memory content, so it can make a better judgment than the local heuristic about whether a memory was truly off-topic. Even then, the result is gated by `feedback_llm_judge_min_confidence`.

### 5. Judge calls use schema-first structured output

The feedback judge now requests compact JSON Schema structured output:

```json
{
  "verdicts": [
    {
      "whisper_log_id": 123,
      "verdict": "used",
      "confidence": 0.9
    }
  ]
}
```

Runtime call parameters:

- `response_format`: JSON Schema when supported
- `temperature=0`
- `max_tokens=512`
- fallback to JSON-object mode if schema output is rejected

Why this is justified:

The judge is part of a background feedback pipeline, so malformed output should be rare and recoverable. The schema removes free-form fields from the required output and keeps the result small. The fallback preserves compatibility with providers that support JSON mode but not JSON Schema.

### 6. Codex watcher setup now uses source-specific metadata

Claude Code transcript support was more polished because its transcript filename is already the hook session id, and its parent directory encodes the project path. Codex uses a different layout: session files can live under date folders such as `~/.codex/sessions/2026/06/24`, and rollout filenames can contain the hook session id inside a longer filename.

This PR now handles that explicitly:

- when `session_watcher_dir` is left at the default Claude path, Ormah also watches `~/.codex/sessions` if it exists
- Codex rollout filenames are resolved back to the hook `session_id` by matching existing `whisper_log` rows
- Codex transcript space comes from the matching `whisper_log.space`, not from date folders
- Claude behavior remains unchanged: Claude filenames still work directly, and Claude project folders still provide the fallback space

Why this is justified:

The feedback loop depends on joining transcript turns back to the exact `whisper_log.session_id` written during whisper injection. Without this, Codex transcripts can parse successfully but fail to produce feedback for the injected memories, or they can be ingested under a bogus space such as a date segment.

## Working Example

Suppose the user asks:

> Where does someone set `feedback_llm_judge`, which LLM is used, and what happens to heuristics?

Ormah injects three memories:

| `whisper_log_id` | Memory | Expected outcome |
|---:|---|---|
| `101` | LLM judge config: enabled by `ORMAH_FEEDBACK_LLM_JUDGE_ENABLED`, uses `llm_provider` / `llm_model`, heuristics remain the baseline | Useful |
| `102` | Graph view color cleanup notes | Irrelevant |
| `103` | NVIDIA service key storage details | Ambiguous |

The assistant answers:

> Set `ORMAH_FEEDBACK_LLM_JUDGE_ENABLED=true`. The judge uses the configured `ORMAH_LLM_PROVIDER` and `ORMAH_LLM_MODEL`. Heuristics still run first as the free local baseline.

After the transcript file stops changing for the debounce window and is processed:

1. Memory `101` has clear evidence in the response.
   - Signal: `whisper_referenced`
   - Affinity: positive `auto_heuristic`
   - LLM judge: skipped for this row

2. Memory `102` has no heuristic positive evidence.
   - Signal: `whisper_unreferenced`
   - If the LLM judge is enabled, the judge sees the prompt, answer, and graph-memory content.
   - Likely judge output:

```json
{
  "verdicts": [
    {
      "whisper_log_id": 102,
      "verdict": "irrelevant",
      "confidence": 0.96
    }
  ]
}
```

   - Signal: `whisper_judged_irrelevant`
   - Affinity: negative `auto_llm_judge`

3. Memory `103` is related to setup but not clearly used by the answer.
   - Signal: `whisper_unreferenced`
   - Likely LLM verdict: `uncertain`
   - Affinity: no change

So the system learns from the observed turn without treating every omitted memory as bad.

## When This Actually Works

The feedback loop becomes active when all of these are true:

1. The user has a version of Ormah with this PR installed and the Ormah server is running.
2. Whisper injection is active for the agent, so `ormah whisper inject` is called on user prompts.
3. The whisper call has a `session_id`, so surfaced memories are written to `whisper_log` for that session.
4. `session_watcher_enabled=true`.
5. At least one effective watched transcript directory exists and receives JSONL transcript files.
6. The transcript file stops changing for `session_watcher_debounce_seconds`, default `60`.
7. The transcript has at least `session_watcher_min_turns`, default `5`.

Then the benefit appears automatically on later similar prompts.

Concrete flow:

1. User asks something in Claude Code, Codex, or another hooked client.
2. `ormah whisper inject` retrieves memories and injects some into context.
3. Ormah logs those injected memories in `whisper_log` with the current prompt, prompt vector, session id, space, and node id.
4. The assistant answers.
5. The transcript file is updated.
6. Session watcher sees the file change, waits for the debounce window, parses it, and finds the assistant response after that prompt.
7. It compares the injected memories against the response.
8. If a memory was clearly used, it records a positive signal and positive affinity.
9. If LLM judge is enabled, ambiguous non-positive rows can be judged as `used`, `irrelevant`, or `uncertain`.
10. Next time a similar prompt comes in, affinity can boost useful memories and suppress confidently irrelevant ones.

For a normal user, the minimal config for the free/local benefit is:

```bash
ORMAH_SESSION_WATCHER_ENABLED=true
ORMAH_SESSION_WATCHER_DIR=~/.claude/projects
```

With the default watcher directory, Ormah watches the Claude Code transcript directory and also watches `~/.codex/sessions` if that Codex directory exists. If a user wants a custom transcript location instead, they can point `ORMAH_SESSION_WATCHER_DIR` directly at it.

For the LLM negative/ambiguous path:

```bash
ORMAH_FEEDBACK_LLM_JUDGE_ENABLED=true
ORMAH_FEEDBACK_LLM_JUDGE_MIN_CONFIDENCE=0.75
ORMAH_LLM_PROVIDER=litellm
ORMAH_LLM_MODEL=nvidia_nim/meta/llama-4-maverick-17b-128e-instruct
```

## Before / After

Before this PR:

- `whisper_log` recorded surfaced memories
- affinity only learned when explicit feedback was submitted
- most normal sessions produced no learning signal
- Codex transcripts could parse but still miss feedback mining when rollout filenames did not equal the hook session id

After this PR:

- debounced transcript files produce observable feedback signals
- clear positive references can improve future ranking for similar prompts
- confident LLM judgments can add positive or negative affinity
- ambiguous/low-confidence cases stay neutral
- automatic negative feedback is only created by explicit feedback or the LLM judge, never by local silence heuristics
- Codex rollout transcripts can be joined back to `whisper_log` and ingested under the correct space

## Model Notes

I ran a local live-model matrix for the LLM judge task. The eval harness is intentionally not included in this PR, but the useful ordering from the broader 15-judgment check was:

1. `nvidia_nim/meta/llama-4-maverick-17b-128e-instruct`
2. `nvidia_nim/qwen/qwen3-next-80b-a3b-instruct`
3. `nvidia_nim/z-ai/glm-5.1`
4. `nvidia_nim/deepseek-ai/deepseek-v4-pro`
5. `nvidia_nim/meta/llama-3.3-70b-instruct`
6. `nvidia_nim/meta/llama-3.1-70b-instruct`

The main differentiator was ambiguous-case handling. The top models handled clear positive and clear negative examples well.

## Verification

- Focused Ruff on touched Python files: passed
- Focused pytest for session watcher: `27 passed`
- Transcript parser tests: `19 passed`
- Codex rollout hook path test: passed
- Full pytest with clean LLM baseline, excluding the local untracked eval guard: `935 passed, 1 deselected, 1 warning`
- Live NVIDIA NIM smoke through the runtime judge path using `llama-4-maverick`: passed

## Review Focus

@AndreLFSMartins, it would be very useful to get your eyes on this and appreciate your help in making Ormah better.

I would especially appreciate review on:

- whether the boundary between neutral signal and affinity promotion is conservative enough
- whether the LLM judge should remain disabled by default
- whether the schema/fallback behavior is the right compatibility tradeoff
- whether `feedback_llm_judge_min_confidence=0.75` is a reasonable default
- whether the Codex watcher metadata handling is the right way to keep client-specific transcript layout at the watcher boundary

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -53,6 +53,15 @@ ORMAH_LLM_BASE_URL=http://localhost:11434
 # Maximum seconds to wait for an LLM response before giving up.
 # ORMAH_LLM_TIMEOUT_SECONDS=60
 
+# ── Feedback judge (optional; requires ORMAH_LLM_PROVIDER != none) ────────────
+# Session watcher always records free/local heuristic feedback signals. Enable
+# this to have the configured LLM judge ambiguous transcript turns and promote
+# high-confidence "used" / "irrelevant" verdicts into positive / negative affinity.
+# Cheap JSON-capable models such as llama3.2, gpt-4o-mini, or claude-haiku are
+# typically enough for this classification task.
+# ORMAH_FEEDBACK_LLM_JUDGE_ENABLED=false
+# ORMAH_FEEDBACK_LLM_JUDGE_MIN_CONFIDENCE=0.75
+
 # ── API Keys (only when ORMAH_LLM_PROVIDER=litellm) ──────────────────────────
 #
 # API keys are inherited from your shell environment at daemon startup.
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ Ormah supports both deliberate recall and involuntary recall.
 
 When an agent knows it needs something, it can explicitly search memory. But memory should not always wait to be asked. Ormah is built to whisper the right memory at the right time, before the next prompt, so the agent starts with context instead of having to go looking for it.
 
+Whisper feedback can learn from transcript files after they stop changing: a local heuristic records clear usage for free, and an optional LLM judge can classify ambiguous turns into positive, negative, or uncertain retrieval signals.
+
 Read more: [Whisper - Involuntary Recall](https://www.ormah.me/docs/how-ormah-works/whisper), [Search and Ranking](https://www.ormah.me/docs/how-ormah-works/search-and-ranking), [Affinity and Feedback](https://www.ormah.me/docs/operations/affinity-and-feedback)
 
 ### Memory Capture
--- a/docs/09 - Affinity and Feedback.md
+++ b/docs/09 - Affinity and Feedback.md
@@ -20,7 +20,10 @@ applied.
 
 Ormah does not infer negative feedback from silence alone. Session-watcher heuristics may
 record neutral `whisper_unreferenced` signals for observability, but only clear positive
-references are promoted into affinity rows.
+references are promoted into affinity rows. If `feedback_llm_judge_enabled` is on and an
+LLM provider is configured, the session watcher can ask the LLM to judge ambiguous turns;
+only confident `used` / `irrelevant` verdicts are promoted into positive / negative
+affinity.
 
 ## Where Feedback Comes From
 
@@ -93,6 +96,18 @@ heuristics record:
 Only `whisper_referenced` heuristic signals are promoted to `affinity`; unreferenced signals
 remain observational.
 
+If the optional LLM judge is enabled, those ambiguous unreferenced rows are grouped by
+prompt/response and sent to the configured LLM. The judge returns `used`, `irrelevant`, or
+`uncertain` with confidence. Ormah records the judge output in `signals` and promotes only
+high-confidence verdicts:
+
+- `whisper_judged_used` with polarity `+1` becomes positive affinity
+- `whisper_judged_irrelevant` with polarity `-1` becomes negative affinity
+- `whisper_judged_uncertain` with polarity `0` remains observational
+
+Low-confidence judge verdicts are normalized to `whisper_judged_uncertain` so they do not
+change ranking.
+
 That is why the system needs `whisper_log` first: affinity rows are learned from previously logged whisper candidates.
 
 ## Stored Fields
@@ -147,6 +162,8 @@ For each candidate node:
 | `affinity_half_life_days` | `30.0` |
 | `affinity_max_boost` | `0.15` |
 | `affinity_implicit_weight` | `0.8` |
+| `feedback_llm_judge_enabled` | `false` |
+| `feedback_llm_judge_min_confidence` | `0.75` |
 
 ## Math
 
@@ -203,6 +220,9 @@ The review candidate is selected from recent `whisper_log` rows where:
 3. Ormah records a signal tied to the exact `whisper_log` row
 4. trusted positive judgments create an affinity row tied to that same prompt vector
 5. on a future prompt with similar wording, that node can receive a small positive score boost
+
+With the optional LLM judge enabled, a clearly off-topic memory can instead receive a
+trusted negative affinity row when the judge returns a confident `irrelevant` verdict.
 
 ## Code Anchors
 
--- a/docs/09 - Affinity and Feedback.md
+++ b/docs/09 - Affinity and Feedback.md
@@ -1,21 +1,26 @@
 # Affinity and Feedback
 
-Verified against the current repository state on 2026-04-13.
+Verified against the current repository state on 2026-06-13.
 
 Affinity is Ormah's feedback-based score adjustment layer for whisper. It learns whether a memory tends to be useful in prompts similar to the current one.
 
 ## How Feedback Enters the System
 
-Affinity learning is driven by feedback submitted through `submit_feedback(...)`. That feedback can be explicit or implicit.
+Affinity learning is driven by trusted feedback about previously surfaced memories. Feedback
+can arrive through `submit_feedback(...)` or from automatic session-watcher signal mining.
 
 In this system:
 
 - `explicit` feedback means a direct judgment is submitted because the user or agent intentionally marks a memory as useful or not useful
 - `implicit` feedback means the client or agent infers usefulness from the interaction and submits that judgment without the user explicitly rating it
 
-In both cases, Ormah learns through `submit_feedback(...)`; the difference is where the judgment came from, not how it is stored.
+In both explicit and implicit `submit_feedback(...)` cases, Ormah learns through the same
+turn-level affinity table; the difference is where the judgment came from, not how it is
+applied.
 
-Ormah does not currently infer negative feedback from silence alone. Affinity rows are created when `submit_feedback(...)` is called.
+Ormah does not infer negative feedback from silence alone. Session-watcher heuristics may
+record neutral `whisper_unreferenced` signals for observability, but only clear positive
+references are promoted into affinity rows.
 
 ## Where Feedback Comes From
 
@@ -28,7 +33,8 @@ When feedback is submitted:
 1. Ormah resolves the node id against `whisper_log`
 2. it looks up the latest logged prompt vector for that node
 3. it inserts an `affinity` row using that stored prompt context
-4. explicit feedback also marks relevant `review_log` entries as answered
+4. it records a corresponding `signals` ledger row
+5. explicit feedback also marks relevant `review_log` entries as answered
 
 Whispered short ids work here too: the resolver accepts full ids first, then falls back to a unique prefix match against `whisper_log`.
 
@@ -64,7 +70,7 @@ So `whisper_log` is the staging table that says:
 
 > "For this prompt/session, Ormah considered these memories, and here is whether each one was actually injected."
 
-### Step 3: later feedback converts logged candidates into affinity rows
+### Step 3: later feedback or usage detection converts logged candidates into signals
 
 When `submit_feedback(node_id, ...)` is called, Ormah does **not** recompute prompt context. It looks up the most recent `whisper_log` entry for that node and copies:
 
@@ -72,8 +78,20 @@ When `submit_feedback(node_id, ...)` is called, Ormah does **not** recompute pro
 - `prompt_text`
 - `space`
 - `session_id`
+- `whisper_log_id`
 
-into the `affinity` table along with the submitted `signal`.
+into the `affinity` table along with the submitted `signal`. It also appends a
+`feedback_submitted` row to `signals`.
+
+When the transcript watcher processes a normalized transcript, it compares injected `whisper_log` rows
+against the assistant response that followed the matching user prompt. Conservative local
+heuristics record:
+
+- `whisper_referenced` with polarity `+1` when the response clearly cites or reuses the memory
+- `whisper_unreferenced` with polarity `0` when no clear reference is found
+
+Only `whisper_referenced` heuristic signals are promoted to `affinity`; unreferenced signals
+remain observational.
 
 That is why the system needs `whisper_log` first: affinity rows are learned from previously logged whisper candidates.
 
@@ -89,6 +107,22 @@ Current stored affinity rows include:
 - `confirmed_at`
 - `space`
 - `session_id`
+- `whisper_log_id`
+
+Current stored signal rows include:
+
+- `whisper_log_id`
+- `node_id`
+- `signal_type`
+- `polarity`
+- `strength`
+- `source`
+- `session_id`
+- `surface`
+- `space`
+- `prompt_hash`
+- `evidence`
+- `created`
 
 ## How Boost Is Computed
 
@@ -125,7 +159,7 @@ for row in affinity_rows:
         continue
 
     recency = exp(-days_ago * ln(2) / half_life)
-    source_weight = implicit_weight if row.source == "implicit" else 1.0
+    source_weight = 1.0 if row.source == "explicit" else implicit_weight
     weight = sim * recency * source_weight
 
     weighted_sum += row.signal * weight
@@ -151,7 +185,8 @@ It can rescue a borderline candidate or slightly suppress a noisy one, but it is
 
 On the first message of a session, whisper may surface one held-back candidate as a review suggestion. That review block asks the client/agent to call `submit_feedback(...)` later if the relevance can be judged.
 
-This is the current bridge between whisper behavior and future affinity learning.
+This remains one bridge between whisper behavior and future affinity learning. The transcript
+watcher now provides a second bridge by mining completed transcripts for clear memory usage.
 
 The review candidate is selected from recent `whisper_log` rows where:
 
@@ -164,9 +199,10 @@ The review candidate is selected from recent `whisper_log` rows where:
 ## Walkthrough Example
 
 1. whisper surfaces a node during a prompt about database decisions
-2. later, the agent calls `submit_feedback(node_id=..., signal=1, source="implicit")`
-3. Ormah records an affinity row tied to the prompt vector from `whisper_log`
-4. on a future prompt with similar wording, that node can receive a small positive score boost
+2. later, either the agent calls `submit_feedback(...)` or the transcript watcher sees the assistant clearly used that node
+3. Ormah records a signal tied to the exact `whisper_log` row
+4. trusted positive judgments create an affinity row tied to that same prompt vector
+5. on a future prompt with similar wording, that node can receive a small positive score boost
 
 ## Code Anchors
 
--- a/docs/09 - Affinity and Feedback.md
+++ b/docs/09 - Affinity and Feedback.md
@@ -98,7 +98,9 @@ remain observational.
 
 If the optional LLM judge is enabled, those ambiguous unreferenced rows are grouped by
 prompt/response and sent to the configured LLM. The judge returns `used`, `irrelevant`, or
-`uncertain` with confidence. Ormah records the judge output in `signals` and promotes only
+`uncertain` with confidence. For LiteLLM-compatible providers, Ormah first requests a
+compact JSON Schema response and falls back to JSON-object mode if the provider rejects
+schema output. Ormah records the judge output in `signals` and promotes only
 high-confidence verdicts:
 
 - `whisper_judged_used` with polarity `+1` becomes positive affinity
--- a/docs/10 - Hippocampus and Session Watcher.md
+++ b/docs/10 - Hippocampus and Session Watcher.md
@@ -52,21 +52,25 @@ The session watcher ingests normalized agent transcript JSONL files.
 It only starts when:
 
 - `session_watcher_enabled == True`
-- the configured watch dir exists
+- at least one effective watch dir exists
 
 Current defaults:
 
 - `session_watcher_enabled = False`
 - `session_watcher_dir = ~/.claude/projects`
+- existing Codex sessions under `~/.codex/sessions` are also watched when the primary
+  directory is left at the default
 - `session_watcher_debounce_seconds = 60`
 - `session_watcher_min_turns = 5`
 - `session_watcher_lookback_hours = 72`
 - `feedback_llm_judge_enabled = false`
 - `feedback_llm_judge_min_confidence = 0.75`
 
-The current default watch directory remains the historical Claude Code path for compatibility.
-The parser and downstream ingestion path are agent-normalized; adding another client should
-mean adding a transcript source/adapter, not changing memory ingestion or signal mining.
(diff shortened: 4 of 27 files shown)


AndreLFSMartins:
Hey @r-spade — first off, apologies for the slow reply on this; I wanted to give your implementation the read it deserved rather than a drive-by 👍. Just went through the merged diff.

Really happy with how you closed this. You took the three secondary concerns from #21 head-on:

- **The judgment problem (#1)** — the hybrid heuristic + LLM judge with `used` / `irrelevant` / `uncertain` verdicts is exactly the shape I was hoping for. Cheap path for the obvious cases, judge only on the ambiguous remainder.
- **The +1/-1 inflation risk (#2)** — restricting negative affinity to explicit feedback or a *confident* judge verdict (never local silence) is the right call. That was my biggest worry and you defused it.
- **The `ON CONFLICT (node_id, session_id)` key (#3)** — moving the uniqueness to `whisper_log_id` (`idx_affinity_node_whisper_log_unique`) with the legacy session index for transition is precisely what was needed for per-turn signal to accumulate instead of being dropped after the first row.

One gap I'd flag as a follow-up, and it's the one that originally motivated the issue for me:

**There's still no coverage/precision metric exposed.** The whole framing of #21 was that whisper effectiveness is *unmeasurable* — and that's still true after this merge. We've closed the collection side, but there's no surfaced `affinity rows ÷ injected` or `+1 / (+1 + -1)` number, so we can't actually verify the loop went from ~0.7% coverage to something healthy on a real instance. Without it, the loop is closed but blind. Would you be open to a small follow-up (CLI/stats surface or a whisper-health line) so the improvement is observable? Happy to take that one.

One design note for the record: going through the `session_watcher` (mining completed transcripts) instead of a deterministic `Stop` hook is a reasonable trade, but it means the loop closes with some latency and rides on the same watcher path I'm touching in #34 (byte-cursor ingest). Worth keeping the two in mind together so the feedback miner and the turn-cursor don't fight over transcript state.

Thanks for picking this up and for the repeated nudge to review — genuinely appreciate the collaboration. 🚀

r-spade:
> Hey @r-spade — first off, apologies for the slow reply on this; I wanted to give your implementation the read it deserved rather than a drive-by 👍. Just went through the merged diff.
> 
> Really happy with how you closed this. You took the three secondary concerns from #21 head-on:
> 
> * **The judgment problem ([feat(mcp): whisper inject — pre-fetch context on initialize, smart get_context cache #1](https://github.com/r-spade/ormah/pull/1))** — the hybrid heuristic + LLM judge with `used` / `irrelevant` / `uncertain` verdicts is exactly the shape I was hoping for. Cheap path for the obvious cases, judge only on the ambiguous remainder.
> * **The +1/-1 inflation risk ([feat(mcp): expose whisper to desktop clients #2](https://github.com/r-spade/ormah/pull/2))** — restricting negative affinity to explicit feedback or a _confident_ judge verdict (never local silence) is the right call. That was my biggest worry and you defused it.
> * **The `ON CONFLICT (node_id, session_id)` key ([feat(improvements): conflict surfacing + gate tuning + decay alerts + versioning + session summary #3](https://github.com/r-spade/ormah/pull/3))** — moving the uniqueness to `whisper_log_id` (`idx_affinity_node_whisper_log_unique`) with the legacy session index for transition is precisely what was needed for per-turn signal to accumulate instead of being dropped after the first row.
> 
> One gap I'd flag as a follow-up, and it's the one that originally motivated the issue for me:
> 
> **There's still no coverage/precision metric exposed.** The whole framing of #21 was that whisper effectiveness is _unmeasurable_ — and that's still true after this merge. We've closed the collection side, but there's no surfaced `affinity rows ÷ injected` or `+1 / (+1 + -1)` number, so we can't actually verify the loop went from ~0.7% coverage to something healthy on a real instance. Without it, the loop is closed but blind. Would you be open to a small follow-up (CLI/stats surface or a whisper-health line) so the improvement is observable? Happy to take that one.
> 
> One design note for the record: going through the `session_watcher` (mining completed transcripts) instead of a deterministic `Stop` hook is a reasonable trade, but it means the loop closes with some latency and rides on the same watcher path I'm touching in #34 (byte-cursor ingest). Worth keeping the two in mind together so the feedback miner and the turn-cursor don't fight over transcript state.
> 
> Thanks for picking this up and for the repeated nudge to review — genuinely appreciate the collaboration. 🚀

Hey @AndreLFSMartins 
  No worries at all. Appreciate you taking the time. 
  Apologies that I merged it before you had a chance to look at it. For the next PR if you are reviewing it do you mind leaving a comment that review is in progress so that I dont cut you short on time ? Would that work ? Happy to know what works for you.
  
  Great point on the whisper effectiveness measurement. I missed that. If you can issue a PR for that it would be absolutely great. 
  
  Have an Ormah App coming out soon which does show a bit of stats, the whisper effectiveness will be a killer addition. 
  
  Been genuinely enjoying the collaboration. Appreciate it :)

AndreLFSMartins:
Sounds like a plan on both counts 👍

**Process:** that works perfectly for me. From the next PR on I'll drop a quick "review in progress" comment as soon as I start reading, so you know there's an in-flight review and don't have to hold the merge guessing. If I go quiet for too long after that, feel free to merge anyway — I'd rather not block you.

**Whisper effectiveness metric:** I'll take it. I'll open a PR that surfaces the coverage/precision numbers (`affinity rows ÷ injected` and `+1 / (+1 + -1)`) as a whisper-health line so the loop stops being blind. I'll keep #34 (byte-cursor ingest) in mind while I'm in there so the feedback miner and the turn-cursor don't fight over transcript state.

And great to hear about the Ormah App — happy to shape the metric so it's easy to plug straight into the stats surface there. Talk soon 🚀

PR #53, opened 2026-06-25
feat(stats): whisper-health metric — surface coverage/precision (follow-up to #40)

AndreLFSMartins:
👀 Heads-up before you read the diff: review in progress is welcome — no rush to merge, but flagging early per our process chat on #40.

## What & why

Follow-up to the review on #40: that PR closed the *collection* side of the whisper feedback loop, but whisper effectiveness was still **unmeasurable** — there was no surfaced number to verify the loop moved from the original ~0.7% coverage to something healthy. This PR adds the read/observability side, nothing on collection.

A new read-only `compute_whisper_health(conn, now)` aggregates `whisper_log` + `affinity` and is exposed under a `whisper_health` key in `engine.stats()` — so `GET /admin/stats` and `ormah stats` both surface it for free, and the upcoming Ormah App can plug straight in.

## What it reports

Over two recuts, `all_time` and `last_7d`:

| field | meaning |
|---|---|
| `injected` | `COUNT(*)` whisper_log rows with `was_injected = 1` |
| `feedback_rows` | distinct injected whispers that drew feedback |
| `coverage` | `feedback_rows / injected` (`None` when injected = 0) |
| `positive` / `negative` | signal tallies |
| `precision` | `positive / (positive + negative)` (`None` when no feedback) |
| `unlinked_feedback_rows` | (all_time only) legacy `whisper_log_id IS NULL` feedback, surfaced not hidden |

Raw counts sit beside the ratios so a tiny denominator is never masked; ratios are `None` (never a fake `0.0`) when their denominator is zero.

## Key design decisions

- **Coverage can't exceed 100%.** Every feedback aggregate anchors in `whisper_log` via `JOIN ... AND wl.was_injected = 1`. This keeps numerator and denominator in one universe and excludes feedback on held-back candidates (`was_injected = 0`) and legacy `NULL` rows. `COUNT(DISTINCT whisper_log_id)` collapses multiple per-turn signals.
- **`last_7d` is a single cohort.** Both sides filter `wl.logged_at >= since`, so recent feedback for an old injection can't push coverage past 1.0. As a bonus, `confirmed_at` (written as `datetime('now')`, a different string format than `logged_at`'s `.isoformat()`) is never compared lexicographically.
- **Consistent snapshot.** The aggregate reads run inside one `BEGIN DEFERRED` (guarded by `in_transaction`), so a concurrent writer can't interleave between the `injected` and `feedback_rows` counts and momentarily report coverage > 100%.
- **Legacy feedback is visible, not silent.** Pre-feedback-loop `affinity` rows (`whisper_log_id IS NULL`) are excluded from the linked-only ratios but exposed via `unlinked_feedback_rows`.

## Tests

12 unit/integration tests covering: empty store, no-feedback, mixed signals, the DISTINCT guard, held-back exclusion, legacy-NULL surfacing, the 7-day cohort boundary, the `confirmed_at` format trap, the snapshot/no-transaction-leak, plus a seeded `engine`-fixture integration test on the real schema and an `/admin/stats` presence assert.

## Known limitation / follow-up (out of scope here)

`submit_feedback` resolves the *latest* whisper_log per node without requiring `was_injected = 1`, so feedback on a re-logged held-back row is undercounted by this metric. The fix belongs in attribution (collection side); I'll open a separate issue. A composite `(was_injected, logged_at)` index on `whisper_log` is also a sensible later optimization if `stats()` is hit hard at scale.

Reviewed locally through two adversarial review rounds before opening. Happy to adjust naming or shape to fit the Ormah App's stats surface.

### DIFF
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -19,6 +19,7 @@ from ormah.engine.maintenance_signal import (
     is_maintenance_due_signal,
 )
 from ormah.engine.tier_manager import TierManager
+from ormah.engine.whisper_health import compute_whisper_health
 from ormah.engine.traversal import (
     format_node_with_neighbors,
     format_search_results,
@@ -1070,6 +1071,9 @@ class MemoryEngine:
             "total_nodes": total,
             "by_tier": tier_counts,
             "total_edges": edge_count,
+            "whisper_health": compute_whisper_health(
+                self.db.conn, datetime.now(timezone.utc)
+            ),
         }
 
     # --- Merge operations ---
--- a/src/ormah/engine/whisper_health.py
+++ b/src/ormah/engine/whisper_health.py
@@ -1,0 +1,84 @@
+"""Whisper effectiveness metrics derived from whisper_log + affinity.
+
+Read-only aggregation for the feedback loop closed in #21 — surfaces coverage
+(share of injected memories that drew any feedback) and precision (positive
+share of feedback on injected memories) so whisper effectiveness stops being
+unmeasurable.
+
+Semantics & known limitations (council-reviewed):
+- coverage/precision are LINKED-ONLY: they count affinity rows joined to an
+  injected whisper (whisper_log_id NOT NULL AND was_injected = 1). Legacy
+  pre-#21 rows (whisper_log_id IS NULL) are excluded but surfaced separately as
+  `unlinked_feedback_rows` on `all_time`, so the loss is visible, not silent.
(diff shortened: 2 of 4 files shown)


AndreLFSMartins:
Opened the promised follow-up for the held-back attribution undercount: #54 (collection-side fix in `submit_feedback`; this PR's read metric stays as-is).

r-spade (approved):
Reviewed locally. The whisper_health aggregation is sound, tests pass, and the known attribution limitation is already tracked separately. We'll handle stats endpoint consolidation in a follow-up PR.

PR #57, opened 2026-06-25
fix(session-watcher): never ingest subagent transcripts (follow-up to #34)

AndreLFSMartins:
## Problem

After #34 merged, the session watcher started ballooning the store with
low-value memories pulled from **subagent (Task-tool) transcripts**.

Claude Code writes subagent transcripts under
`~/.claude/projects/<project>/<uuid>/subagents/agent-*.jsonl`. The
watcher's recursive `rglob("*.jsonl")` + recursive observer have always
captured these, but two things made it surface now:

- The `min_turns` gate is bypassed for **idle** files (a finished short
  session is flushed so it isn't stranded — `_ingest_session`). Subagent
  transcripts idle within seconds with only 1–2 turns.
- #34's safe-payload / completion-signal turn closing made those short
  subagent transcripts produce a valid closed payload.

Net effect on one heavy multi-agent afternoon: **580+ nodes** created in
a junk `space="subagents"` (derived from `path.parent.name`). This is not
duplication — embeddings show ~0 semantic overlap — it is over-ingestion
of internal agent scratch that was never meant to be a user session.

## Fix

- Add `_is_subagent_transcript(path)` — true when a `subagents` segment
  appears anywhere in the path (`"subagents" in path.parts`), covering
  nested layouts.
- Skip such transcripts at the single ingest chokepoint
  (`_ingest_session`, covering both the catch-up scan and the realtime
  event path) plus the live `on_created`/`on_modified` handlers to avoid
  pointless debounce timers.
- Apply the same predicate in `setup.py::_discover_transcripts`, so the
  `ormah setup` backfill never offers/ingests subagent transcripts
  either (sibling caller).

## Tests

- `test_subagent_transcript_is_not_ingested` — direct ingest skipped
  even with turns above `min_turns`.
- `test_scan_skips_subagents_keeps_primary` — scan ingests the primary
  session transcript but not its sibling subagent transcripts.
- `test_skips_subagent_transcripts` (setup) — backfill discovery
  excludes subagent transcripts.

43 passed locally; ruff clean.

### DIFF
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -562,6 +562,15 @@ def _record_whisper_usage_signals(
     return recorded
 
 
+def _is_subagent_transcript(path: Path) -> bool:
+    """True for subagent transcripts (Claude Code writes them under ``<uuid>/subagents/``).
+
+    These are internal agent scratch, not user-facing sessions — ingesting them balloons
+    the store with low-value granular memories under a junk ``subagents`` space.
+    """
+    return path.parent.name == "subagents"
+
+
 def _space_from_encoded_dir(dirname: str) -> str | None:
     """Extract project space from an encoded transcript directory name.
 
@@ -706,6 +715,8 @@ def _ingest_session(
     on_defer_active=None,
 ) -> bool:
     """Ingest a single JSONL session transcript if changed. Returns True if ingested."""
+    if _is_subagent_transcript(path):
+        return False
     rel = str(path.relative_to(watch_dir))
 
     try:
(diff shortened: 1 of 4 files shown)


AndreLFSMartins:
Rebased onto main @ 2e76b5b. Local suite green (same 11 environmental failures as clean main; 0 new).

r-spade (approved):
Thanks @AndreLFSMartins . This is a useful fix and I agree the need is real: Claude Code subagent transcripts are internal Task-tool scratch, and recursively ingesting them creates noisy low-value memories under a junk subagents space.

I reviewed the watcher ingestion path, realtime event handling, setup backfill discovery, and the regression tests. I also verified the branch locally on Python 3.12 and checked a temporary merge with current main: the merge was clean, affected watcher/setup tests passed, and Ruff was clean. This looks good to merge.

PR #60, opened 2026-06-26
fix(session-watcher): disk-truth reconcile safety net for dropped transcripts (#59)

### LINKED ISSUE #59: bug(session-watcher): live path intermittently drops a fraction of transcripts (whole sessions + held-back tails); recovered only by a restart's catch-up, >72h escapes lost
> **2026-06-25 — issue substantially rewritten after an empirical diagnosis.** The original framing ("the live path goes silent during normal uptime; a restart drains hundreds of sessions the live watcher never ingested") was **largely refuted**. The real defect is narrower and is documented below. A **Correction** section at the end records exactly what the first version got wrong and why, so the history stays honest.

> Related: #52 fixes a *consequence* (the catch-up drain blocking the port bind on startup). This issue is about the *cause* — why some backlog exists to be drained at all.

## Summary

The session-watcher ingests transcripts in real time via a `watchdog` `Observer`, and **for the most part this works**: on active days the live path ingests tens of sessions while a restart's catch-up drains **zero**. The defect is **intermittent, partial loss** — the live path silently fails to ingest a *fraction* of transcripts (both whole sessions and the held-back *tails* of ongoing sessions) because the filesystem event never reaches the handler. Those escapes sit uningested until the next restart's `_scan_sessions` catch-up recovers them. Without a restart they accumulate, and any escape older than the `72h` lookback is **permanently lost**.

This is **not** "the live path goes silent and everything piles up." Steady-state live ingestion works; a measurable minority leaks.

## Evidence (re-measured against the authoritative daemon log `/tmp/ormah-dev.err`, 06-17→06-25)

- **Live works on active days.** 06-18→06-24 every day had **catch-up = 0** (e.g. 06-22: 47 sessions ingested live, 06-24: 48; quieter days fewer). Isolated restarts on 06-22 and 06-23 also caught up **0** — the live cursor was current.
- **Idle flush works.** A session that goes idle is ingested ~60s after its last write (debounce). Proven in a controlled test: stream → stop → live ingests 65s later, no restart.
- **Real intermittent loss is confirmed.** ~56 whole sessions across the week were *ingestible* (`safe_user_turn_count == user_turn_count`, no `leading_orphan`), idle for **days**, with the server up and ingesting *other* sessions — yet had **zero** live ingest attempt (handler never fired; the only log line for them is the catch-up). Verified by running the real parser on them. The filesystem event never reached the handler.
- **The "265-session catch-up" was misread in the first version.** It was a *single* scan at 13:22 on 06-25 and decomposes as:
  - **144 subagent transcripts** — a *separate* bug (subagent transcripts should never be ingested), fixed the same day in `366a65b`, already purged from the DB. Pure noise inflating the count (~half of it).
  - **65 partial tail-deltas** — sessions already ingested earlier, of which only the *trailing* turns were still pending (e.g. 7/91, 19/531, 32/801 new/total turns; full re-parse = 0, partial = 65, measured). The safe-boundary held the tail at live time; catch-up committed it. New turns, **not** re-processing.
  - **56 whole never-ingested sessions** — the genuine live escapes — spread across 06-22:6 · 06-23:16 · 06-24:18 · 06-25:15 · 06-16:1.
  So the real "the live path missed this" signal is **~121 legitimate items** (56 whole sessions + 65 tails), not 265 — and the 3 largest (267/182/140 turns, 0–4 memories) were never-ingested because their **06-17 extraction failed** (qwen-mlx, that day's `llm_fail` was high), not because the live path dropped them.
- **The 06-25 node spike was self-inflicted, not restart-shaped growth.** That day had a dev session *on the watcher itself*: 33 server restarts (crash-loop) + uvicorn reloads from editing `session_watcher.py`, plus the 144 subagent-noise ingests. Ordinary growth (~400 nodes/day) is **real heavy usage × high per-session extraction** (~7–19 memories/session, no importance filter), ingested live, normally — not a restart artifact.

## Why it's silent (still valid)

A live path that drops an event logs nothing. There is no functional heartbeat asking "has the live path ingested anything recently while transcripts on disk advanced?", so the intermittent leak is invisible until a restart's catch-up surfaces it. Uptime ("PID alive") is not a health signal for the live path.

## Root cause — open (not proven)

What is **not** yet proven is *why* the filesystem event fails to reach the handler for some files: FSEvents coalescing/dropping events, a race in the handler, the Observer thread degrading across macOS sleep/wake, or something else. The live path is unsupervised — `start_session_watcher()` does `observer.start()` once with no `is_alive` poll or restart (`src/ormah/background/session_watcher.py:1003-1005`) — and debounce/idle use in-memory `threading.Timer`s lost on restart (`:909-933`), so a session deferred at restart time loses its timer and waits for a scan. But which of these is the actual leak mechanism is **unproven by design**: the failure is silent, so the logs don't capture the dropped event. Confirming it requires instrumenting the live path (below).

(Note: the earlier "cursor state tracks 482 of 3299 `.jsonl` on disk" figure is **not** evidence of breakage — most of those 3299 are old/out-of-lookback/subagent transcripts the live cursor is never expected to carry.)

## Impact

- **A fraction of memory is silently not captured in real time** and depends on a restart's catch-up to recover. Most memory *is* captured live; the leak is a minority, but real.
- **Hard data loss beyond the lookback window.** The catch-up is bounded by `session_watcher_lookback_hours = 72` (`src/ormah/config.py:78`). An escape whose tail ages past **72h** without a restart is **never ingested** — permanently lost.
- **Couples with the bind-blocking issue (#52):** the larger the silent backlog grows, the longer the next restart's catch-up blocks the port.

1. **Supervise / replace the Observer.** Poll `observer.is_alive()` on a timer and recreate it if dead, or replace the single long-lived Observer with a periodic reconciling scan that does not depend on FSEvents liveness. Re-arm cleanly across macOS sleep/wake.
2. **Add a functional heartbeat / liveness metric.** Track "last live ingestion" and warn when the live path has been idle while transcripts on disk advanced — so a leak is detected without a restart.
3. **Don't rely on a 72h lookback as the safety net** for the reconciling scan.

## Investigation needed (to pin the mechanism)

- [ ] Instrument the live path: log every `on_created`/`on_modified` and every `_do_ingest`, plus a periodic `observer.is_alive()` line. Run across a sleep/wake cycle and a long uptime.
- [ ] Determine whether the Observer thread dies (and when) vs. stays alive but receives no events for some files.
- [ ] During a known-active no-restart window, compare transcripts advanced on disk vs. ingested live, and identify what distinguishes the dropped files.

## Acceptance criteria

- [ ] Every ingestible transcript (whole session or closed tail) is ingested live without waiting for a restart.
- [ ] A stalled/leaking live path is detected and recovered (or made impossible) **without** a restart.
- [ ] No pending transcript tail is lost due to the 72h lookback bound.
- [ ] A liveness signal exists so "the live path dropped events" is observable.

## Correction — what the first version of this issue got wrong (2026-06-25)

The original report overstated the defect by reading a contaminated day (06-25) as the steady state:

1. **"The live path goes silent during normal uptime."** False — live ingests tens of sessions/day on active days with catch-up = 0.
2. **"265 sessions the live watcher never ingested."** Misread: 144 were subagent noise (separate bug, fixed `366a65b`), 65 were legitimate tail-deltas of already-ingested sessions, only **56** were whole live escapes.
3. **"Growth is restart-shaped (898/901 on restart days)."** The 06-25 spike was a self-inflicted dev session on the watcher (33 restarts + subagent noise), not organic restart-driven growth.
4. **"Cursor tracks 482 of 3299 → most sessions never carried live."** Not evidence of breakage; most of the 3299 are out-of-lookback/subagent/old transcripts never expected on the live cursor.

What **survives**: the live path *does* intermittently drop a fraction of transcripts (≈56 whole sessions + tails this week, parser-verified), the loss is silent (no heartbeat), and escapes older than the 72h lookback are permanently lost. The root-cause mechanism remains **open**.

Full diagnosis: `SESSION_WATCHER_DIAGNOSIS_2026-06-25.md` (local repo).

AndreLFSMartins:
## Summary

Fixes #59 — the session-watcher's live FSEvents path intermittently drops a fraction of transcripts (whole sessions + held-back tails), recoverable only by a restart's catch-up, and anything past the 72h lookback is lost permanently.

Adds a mechanism-agnostic safety net on top of the existing live path:

- **`SessionHandler.reconcile()`** — a stat-only disk-truth scan that re-ingests transcripts the live path missed (candidate = never-seen within lookback, or a state cursor not at EOF). Routes through the existing `_do_ingest` (single state owner → no clobber / no double-ingest).
- **Periodic scheduler job + Observer supervision** — runs reconcile every 5 min (configurable), recreating a dead Observer (`is_alive()`) first; heartbeat log on recovery.
- **Tri-state `IngestResult`** (OK / NO_PROGRESS / TRANSIENT) — reconcile parks only files whose *content* can't progress (corrupt / frozen safe boundary), never transient external (LLM/DB) failures.
- **Park/retry keyed on a `(size, mtime_ns)` change-token** — a same-size content rewrite un-parks; persistently-TRANSIENT files are *deprioritized* (oldest-first FIFO, never hard-parked) so they can't monopolize the per-tick cap and starve valid candidates.
- **Per-tick wall-clock budget + mtime-priority** — bounds scheduler-thread occupancy; freshly-dropped transcripts recovered soonest.
- **Narrow `_state_lock`** — serializes only the state write, leaving parse/LLM/DB lock-free so reconcile never starves the live path.
- **Standalone `scripts/diag/fsevents_monitor.py`** — proves *why* FSEvents misses events (logs event types incl. `on_moved`, raw native flags, disk-truth vs events) without touching production.

## Review

Peer-reviewed via a 3-round adversarial council (cursor + codex). Findings F1/F2/F3 and H1/H2 fixed and re-confirmed. H1' (a same-size rewrite that *preserves* `mtime_ns`) is accepted as a documented known limitation — closing it would reintroduce O(n)-per-tick hashing for a pattern that doesn't occur in append-only transcript workloads (see the code comment at the fully-consumed skip in `session_watcher.py`).

## Stacked on #57

This branch is stacked on #57 (`fix/exclude-subagent-transcripts`) — `reconcile()` reuses `_is_subagent_transcript`. Until #57 merges, this PR's diff includes #57's commits; they drop automatically once #57 lands.

## Test plan

- [ ] `pytest tests/test_background/test_session_watcher.py tests/test_config.py` — 60 reconcile/watcher tests (tri-state, park/un-park, deprioritize FIFO, per-tick budget, lookback<0, Observer recreate) + config validators
- [ ] Run `scripts/diag/fsevents_monitor.py` across a sleep/wake cycle to confirm the leak mechanism

## Deferred follow-ups (non-blocking)

- Same-size `mtime_ns`-preserved rewrite (above)
- `rglob` incremental index (scan scale)
- `on_moved` live handler (gated on fsevents_monitor evidence)
- Non-atomic DB+cursor dup-on-crash (pre-existing; dedup jobs are the backstop)

### DIFF
--- a/scripts/diag/fsevents_monitor.py
+++ b/scripts/diag/fsevents_monitor.py
@@ -79,6 +79,10 @@ def poll_disk_truth(root: Path, seen: set[str], lock: threading.Lock,
             advanced = key not in last_mtimes or m > last_mtimes[key]
             last_mtimes[key] = m
             if advanced and key not in events_this_window:
+                # A write landing between the swap-and-clear above and this rglob can show a
+                # one-off spurious LEAK (its event arrives next window). A *real* leak (e.g. an
+                # on_moved/atomic-replace the production handler ignores) repeats EVERY window —
+                # so trust a path that leaks repeatedly, not a lone hit.
                 leaks.append(key)
         if leaks:
             log.warning("LEAK %d file(s) advanced on disk with NO fsevent this window:", len(leaks))
--- a/scripts/diag/fsevents_monitor.py
+++ b/scripts/diag/fsevents_monitor.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+"""Standalone FSEvents leak monitor (diagnostic for r-spade/ormah#59).
+
+Proves WHY the session-watcher's live path misses some transcripts, WITHOUT touching
+production code. Run it alongside the server across a sleep/wake cycle and a heavy day,
+then read the log. It:
+  1. Logs every watchdog event by TYPE (created/modified/MOVED/deleted). The production
+     handler implements only on_created/on_modified — a transcript written via
+     rename/atomic-replace arrives as on_moved and is missed deterministically.
+  2. Enables watchdog's FSEvents DEBUG line: the raw native flags per event
+     (is_coalesced / must_scan_subdirs / *_dropped) — the kernel saying it dropped events.
+  3. Every --poll seconds, diffs disk truth vs. events: files whose mtime advanced but for
+     which NO event fired in the window = the leaks.
+
+Usage:
+  uv run --with watchdog python scripts/diag/fsevents_monitor.py \
+      --dir ~/.claude/projects --poll 30 --log /tmp/fsevents_monitor.log
+"""
+from __future__ import annotations
+
+import argparse
+import logging
+import threading
+import time
+from pathlib import Path
+
+from watchdog.events import FileSystemEventHandler
+from watchdog.observers import Observer
+
+log = logging.getLogger("fsevents_monitor")
+
+
+class LoggingHandler(FileSystemEventHandler):
+    def __init__(self, seen: set[str], lock: threading.Lock) -> None:
+        self._seen = seen
+        self._lock = lock
+
+    def _record(self, kind: str, path: str) -> None:
+        if not path.endswith(".jsonl"):
+            return
+        with self._lock:
+            self._seen.add(path)
+        log.info("EVENT %-9s %s", kind, path)
+
+    def on_created(self, event):
+        if not event.is_directory:
+            self._record("created", event.src_path)
+
+    def on_modified(self, event):
+        if not event.is_directory:
+            self._record("modified", event.src_path)
+
+    def on_moved(self, event):
+        # The production handler does NOT implement this — a leak if it fires for .jsonl.
+        if not event.is_directory:
+            self._record("MOVED", getattr(event, "dest_path", event.src_path))
+
+    def on_deleted(self, event):
+        if not event.is_directory:
+            self._record("deleted", event.src_path)
+
+
+def poll_disk_truth(root: Path, seen: set[str], lock: threading.Lock,
+                    interval: float, stop: threading.Event) -> None:
+    last_mtimes: dict[str, float] = {}
+    while not stop.wait(interval):
+        with lock:
+            events_this_window = set(seen)
+            seen.clear()
+        leaks = []
+        for f in root.rglob("*.jsonl"):
+            if "subagents" in f.parts:
+                continue
+            try:
+                m = f.stat().st_mtime
+            except OSError:
+                continue
+            key = str(f)
+            advanced = key not in last_mtimes or m > last_mtimes[key]
+            last_mtimes[key] = m
+            if advanced and key not in events_this_window:
+                leaks.append(key)
+        if leaks:
+            log.warning("LEAK %d file(s) advanced on disk with NO fsevent this window:", len(leaks))
+            for k in leaks:
+                log.warning("  LEAK %s", k)
+        else:
+            log.info("poll ok: every advanced file had an event")
+
+
+def main() -> None:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--dir", default="~/.claude/projects")
+    ap.add_argument("--poll", type=float, default=30.0)
+    ap.add_argument("--log", default="/tmp/fsevents_monitor.log")
+    ap.add_argument("--duration", type=float, default=0.0, help="seconds; 0 = until Ctrl-C")
+    args = ap.parse_args()
+
+    logging.basicConfig(
+        filename=args.log, level=logging.INFO,
+        format="%(asctime)s %(levelname)s %(name)s %(message)s",
+    )
+    logging.getLogger().addHandler(logging.StreamHandler())
+    # Raw native FSEvents flags (is_coalesced / must_scan_subdirs / *_dropped):
+    logging.getLogger("watchdog.observers.fsevents").setLevel(logging.DEBUG)
+
+    root = Path(args.dir).expanduser().resolve()
+    seen: set[str] = set()
+    lock = threading.Lock()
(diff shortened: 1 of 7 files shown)


AndreLFSMartins:
Rebased onto main @ 2e76b5b (resolved conflicts in test_setup.py imports and test_config.py — both sides added independent tests, kept both). Local suite green (same 11 environmental failures as clean main; 0 new).

r-spade (approved):
Thanks @AndreLFSMartins . This is a useful fix and I agree the need is real: the live watcher can miss an ingestible transcript or closed tail silently, and a periodic disk-truth cursor scan is the right safety net rather than betting everything on FSEvents.

I reviewed the session watcher reconcile path, scheduler/startup wiring, config defaults, and regression tests. I rebased this onto current main now that the subagent-transcript exclusion has landed separately, and ran `UV_CACHE_DIR=/tmp/uv-cache ORMAH_LLM_PROVIDER=none uv run pytest tests/test_background/test_session_watcher.py tests/test_config.py -q` plus Ruff on the touched files. CI lint is green and the test check is still running; assuming that finishes green, this looks good to merge.

PR #68, opened 2026-06-27
perf(startup): drain session-watcher catch-up off the bind path (#52)

### LINKED ISSUE #52: perf(startup): session-watcher catch-up scan runs on the bind path — HTTP unavailable for minutes on restart
## Summary

`start_session_watcher()` runs its catch-up scan **synchronously inside the FastAPI lifespan
startup**, before uvicorn binds the port. The scan drains every transcript with appended content
since the last ingest, and **each pending file triggers a full LLM extraction + embeddings**,
serially. With many recently-active sessions, the HTTP server is unavailable for the entire drain
— observed **13+ minutes** on a single restart in a heavy-usage environment.

This is the same class of problem #32 fixed for the embedding backfill (moved "off the bind path —
never blocks startup"). The session-watcher catch-up is the sibling that never got that treatment.

## Evidence

`src/ormah/main.py` `lifespan()` (around L234) calls the watcher start synchronously:

```python
# Start session watcher for agent transcripts
session_observers = start_session_watcher(engine)   # <-- blocks until catch-up finishes
app.state.session_watcher_observers = session_observers
yield   # uvicorn binds the socket only AFTER startup yields
```

`src/ormah/background/session_watcher.py` `start_session_watcher()` (around L1012) does the scan
before starting the observer:

```python
for watch_dir in watch_dirs:
    # Catch-up scan  <-- synchronous; one _ingest_session per pending file, each an LLM call
    ingested = _scan_sessions(engine, watch_dir, min_turns, lookback_hours)
    ...
    observer.start()   # real-time watcher only starts after the scan
```

uvicorn's `LifespanOn.startup()` is awaited **before** the sockets are bound, so a long lifespan
startup delays the bind. Confirmed in the logs: `Started server process [PID]` →
`Waiting for application startup.` → (minutes of `Session watcher ingested ...`) with **no**
`Application startup complete` / `Uvicorn running` and nothing listening on the port.

## The byte cursor is working — this is breadth, not re-reading

Measured on one restart (57 ingests, 57 distinct files, **zero** repeats, **zero** full re-ingests):

| new turns per ingest | count |
|----------------------|-------|
| 1 | 31 |
| 2 | 22 |
| 4 / 7 | 1 each |
| 19 / 32 | 1 each (long-running sessions) |

53 of 57 are tiny tails (1–2 turns). The incremental byte cursor (#33/#34) is doing its job — each
file reads only its appended tail. The cost is **not** depth (bytes re-read); it is **width**: N
distinct sessions each fire one full LLM extraction regardless of tail size. The cursor caps bytes
read, not the number of LLM calls.

The batch piles up when live processing has a gap (watcher stalled, or a long active session burst):
appends accumulate across many files, and the next restart drains them all at once — on the bind path.

## Impact

- HTTP (whisper inject, recall, UI, `/agent/whisper`, `/ingest/*`) is **down for the whole drain**
  on every restart that has a backlog. Under `KeepAlive`, a crash-restart compounds it.
- Scales with the number of recently-active sessions × per-session LLM latency, not with store size.

Mirror the #32 fix. Start the real-time observer immediately and run the catch-up `_scan_sessions`
off the bind path (a daemon thread, like `_start_backfill_fallback`), so uvicorn binds at once and
the backlog drains in the background. Live events arriving during the drain are already debounced
and guarded by the in-flight set, so ordering is safe.

Optional follow-ups (separate, lower priority):
- Cooperative cancellation on shutdown so a long drain doesn't delay a clean stop.
- Bound the per-restart LLM fan-out (the scan is serial today; a backlog of N sessions = N serial
  extractions).

## Acceptance criteria

- [ ] uvicorn binds the port without waiting for the catch-up scan (HTTP serves within startup, not
      after the backlog drains).
- [ ] The catch-up still ingests every pending tail exactly once (no regression of the byte cursor).
- [ ] The background drain stops cooperatively on shutdown.

AndreLFSMartins:
Fixes #52.

## Problem

`start_session_watcher` runs the initial catch-up scan **synchronously on the bind path**: on a restart with a backlog of transcripts, the FastAPI lifespan blocks inside `_scan_sessions` (rglob + parse + LLM extraction per file) before uvicorn finishes binding, so **HTTP is unavailable for minutes**.

## Approach

Observers come up **immediately** (live from t0, nothing missed), and the catch-up backlog drains in a **non-daemon background thread off the bind path** (`ormah-session-catchup`):

- A shared `Semaphore(session_watcher_catchup_concurrency)` (default 1) bounds live + catch-up LLM extractions so the backlog drain can't stampede the model.
- Catch-up and live ingest share one in-flight guard (`_ingesting`, claimed under `_lock` with the stop-check *before* the claim), so a file is never double-ingested and the two paths coordinate exactly-once.
- `catchup_ingest` never touches live debounce timers.

## Shutdown safety (use-after-close)

`stop_session_watcher` now fully drains both ingest sources before returning (the lifespan calls `engine.shutdown()` / `db.close()` right after):

- Under-`_lock` stop checks reject **new** ingests; the drain then polls `in_flight_count()` to 0 so nothing touches the DB after `db.close()`.
- The drain is **uncapped by design** — a deadline cap would abandon a running ingest and re-open the use-after-close window. A watchdog logs every ~5s so a stuck drain is visible instead of a silent hang.
- Startup is **transactional**: a failure after the first observer starts tears down everything already running (stop_event + cancel timers + stop/join observers + drain in-flight) and re-raises, so a partially-started watcher can't leak an observer that writes after close.

`main.py` is unchanged (`SessionWatcherHandle` is a drop-in for the old observer list).

## Tests

56 unit tests covering the off-bind catch-up, the shared semaphore bound, exactly-once dedup between live and catch-up, post-stop rejection, drain-before-`engine.shutdown()` ordering, and transactional-startup cleanup.

## Relationship to #60

This overlaps with #60 (disk-truth reconcile safety net): both recover sessions the live FSEvents path missed. They fix **different layers** — this PR is startup/shutdown *lifecycle* (catch-up off the bind path + drain-before-close), #60 is ongoing *recovery quality* (periodic stat-scan + dead-observer recreation).

If #60 lands first, this PR should be **slimmed to the two lifecycle fixes** (off-bind startup + drain-before-close) and the catch-up thread dropped in favor of running #60's `reconcile()` once off-bind at startup — the two then compose instead of duplicating recovery. Happy to sequence whichever way you prefer.

### DIFF
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -8,10 +8,10 @@ import logging
 import re
 import time
 from dataclasses import dataclass
-from datetime import datetime, timezone
+from datetime import UTC, datetime
 from enum import Enum
 from pathlib import Path
-from threading import Lock, Timer
+from threading import Event, Lock, Thread, Timer
 
 from watchdog.events import FileSystemEventHandler
 from watchdog.observers import Observer
@@ -447,7 +447,7 @@ def _record_whisper_usage_signals(
     if not rows:
         return 0
 
-    now_iso = datetime.now(timezone.utc).isoformat()
+    now_iso = datetime.now(UTC).isoformat()
     recorded = 0
 
     heuristic_records: list[dict] = []
@@ -778,7 +778,7 @@ def _ingest_session(
             logger.info("Session watcher recovering legacy mid-response cursor for %s", rel)
             prev_offset = 0
             result = parse_transcript(path, start_offset=0)
-    except Exception as e:
+    except Exception as e:  # noqa: BLE001 - transcript parsers can raise provider-specific errors
         logger.warning("Session transcript parse error for %s: %s", path, e)
         return IngestResult.NO_PROGRESS
 
@@ -836,7 +836,7 @@ def _ingest_session(
             logger.warning("Session watcher ingestion failed for %s: %s", path, ingested)
             return IngestResult.TRANSIENT
         count = len(ingested) if isinstance(ingested, list) else 0
-    except Exception as e:
+    except Exception as e:  # noqa: BLE001 - engine providers surface heterogeneous failures
         logger.warning("Session watcher ingestion error for %s: %s", path, e)
         return IngestResult.TRANSIENT
 
@@ -850,7 +850,7 @@ def _ingest_session(
     entry = {
         "hash": h,
         "end_offset": payload_offset,
-        "last_ingested": datetime.now(timezone.utc).isoformat(),
+        "last_ingested": datetime.now(UTC).isoformat(),
         "session_id": result.session_id,
         "source": result.source,
         "space": space,
@@ -942,6 +942,8 @@ class SessionHandler(FileSystemEventHandler):
         self._pending: set[str] = set()
         self._lock = Lock()
         self._state_lock = Lock()
+        self._reconcile_lock = Lock()
+        self._stop_event = Event()
         self._reconcile_attempts: dict[str, tuple[int, int, int]] = {}  # rel -> (size, mtime_ns, no_progress_count)
         self._reconcile_transient: dict[str, tuple[int, int, int]] = {}  # rel -> (size, mtime_ns, count)
 
@@ -949,6 +951,8 @@ class SessionHandler(FileSystemEventHandler):
         """Schedule a debounced ingestion for the given file."""
         key = str(path)
         with self._lock:
+            if self._stop_event.is_set():
+                return
             if key in self._timers:
                 self._timers[key].cancel()
             timer = Timer(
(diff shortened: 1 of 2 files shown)


AndreLFSMartins:
Rebased onto main @ 2e76b5b (resolved .gitignore conflict — kept upstream's granular dev-council ignore block). Local suite green (same 11 environmental failures as clean main; 0 new).

r-spade (approved):
Thanks Andre. I rebased this onto current main and kept the fix aligned with the existing reconcile architecture: observers now start immediately, the first disk-truth reconcile runs off the bind path, and shutdown rejects/drains in-flight ingest before engine shutdown.\n\nLocal verification: session watcher tests pass, touched-file Ruff passes, config/main adjacent tests pass, and the full pytest suite passes locally. This is code-ready from my side; the remaining GitHub lint failure is the known repo-wide Ruff/tooling issue, not this session-watcher change.

PR #69, opened 2026-06-27
fix(engine): double-checked locking in _get_hybrid_search (#27)

### LINKED ISSUE #27: MemoryEngine._get_hybrid_search: lock-free lazy init can double-construct HybridSearch under concurrent recall
## Summary

`MemoryEngine._get_hybrid_search` is a lock-free check-then-set lazy initializer:

```python
def _get_hybrid_search(self):
    if self._hybrid_search is not None:
        return self._hybrid_search
    try:
        from ormah.embeddings.hybrid_search import HybridSearch
        self._hybrid_search = HybridSearch(self.db, self.settings)
        return self._hybrid_search
    except ImportError:
        return None
```

Now that #19 moves `recall_search` off the event loop into the Starlette threadpool, this runs concurrently. The first burst of concurrent recalls (before `_hybrid_search` is populated) can have two or more threads all see `None` and each construct a `HybridSearch`, with last-write-wins on the instance attribute.

## Discovered

Raised by @AndreLFSMartins during review of #19 (the server-freeze fix).

## Impact (currently benign)

This is not a correctness bug today:

- After #19, every `HybridSearch` resolves its connection per-thread, so two instances are independently thread-safe (no shared `sqlite3.Connection`).
- `get_encoder` is memoized in `_adapter_cache` and warmed at startup via `MemoryEngine._warmup_embedder()`, so a second `HybridSearch` construction reuses the cached encoder. There is no double model load.
- The orphaned instance is simply garbage collected.

Net cost is a few throwaway objects on the very first concurrent hit. No data races, no double model load.

## Possible fixes

1. Double-checked locking around the lazy init (a `threading.Lock` on the engine, check-lock-recheck). Avoids per-call lock cost after warmup.
2. Construct `HybridSearch` eagerly during `startup()` so the lazy path never races.
3. Leave as-is and document the intentional benign race.

Note: the same lock-free check-then-set pattern exists in `get_encoder` itself (`_adapter_cache`), masked in practice by the startup warmup. Worth considering together if we decide to harden this.

## Decision

Left as-is in #19 per simplicity-first (benign, zero correctness impact). Tracking here so it is not lost.

AndreLFSMartins:
## Summary

Closes #27.

`MemoryEngine._get_hybrid_search` was a lock-free check-then-set lazy initializer. Since #19 moved `recall_search` off the event loop into the Starlette threadpool, the first burst of concurrent recalls (before `_hybrid_search` is populated) could have multiple threads each see `None` and each construct a `HybridSearch`, with last-write-wins on the instance attribute.

This guards the lazy init with **double-checked locking**:

- Fast path stays lock-free once warmed (the existing `is not None` early-return).
- A `threading.Lock` serializes only the cold construction; a recheck inside the lock ensures exactly one instance is built and shared.

Chosen over eager construction in `startup()` because it closes the race independently of whether `startup()` runs before any recall — same cost, more robust.

## Test

Adds `test_get_hybrid_search_constructs_once_under_concurrency`: 8 threads released simultaneously via a `Barrier`, with a stub `HybridSearch` whose `__init__` widens the race window. Asserts a single construction and a shared instance. Fails (`construct_count == 8`) without the lock; passes with it.


r-spade:
Thanks Andre, this is a clean fix. I pushed a maintainer merge commit to resolve the drift with current main while keeping your double-checked locking change and regression test intact. I verified the focused concurrency/test slice locally before merging.

r-spade (approved):
Reviewed locally. The fix is scoped and the regression test captures the cold-start construction race; focused concurrency tests and Ruff passed.

PR #71, opened 2026-06-30
fix(setup): stop clobbering user config (co-tenant hooks + .env comments) (#70)

### LINKED ISSUE #70: setup: install paths clobber pre-existing user content (hooks lists + .env)
## Summary

`ormah setup` clobbers pre-existing, user-owned content in several external config files instead of surgically merging only Ormah's own entries. The shared primitive `_merge_json_file` does a **one-level** dict merge, so for hook configs (where the per-event value is a *list*) it replaces the entire list — wiping any co-tenant hook registered under the same event. The `.env` writer has the same family of bug: it rewrites the whole file and drops everything it doesn't model.

Notably, the **uninstall** paths already do the right thing (per-matcher filtering via `_is_ormah_hook`), so the install paths are simply asymmetric with the removal logic that already exists in the same file.

### 1. Claude Code hooks — LOSSY
`configure_claude_hooks` → `_merge_json_file("~/.claude/settings.json", {"hooks": hooks})` (`src/ormah/setup.py:196`).
Events claimed: `UserPromptSubmit`, `PreCompact`, `SessionEnd`. Any third-party hook already registered under those event names is replaced by Ormah's single hook.

### 2. Codex hooks — LOSSY (same root cause)
`configure_codex_hooks` → `_merge_json_file("~/.codex/hooks.json", {"hooks": hooks})` (`src/ormah/setup.py:305`).
Events claimed: `UserPromptSubmit`, `Stop`. `Stop` is a commonly co-used event, so the blast radius is larger here.

### 3. `.env` writer — LOSSY (different shape, same philosophy)
`_read_env_file` (`src/ormah/setup.py:755`) skips every comment and blank line (`if line.startswith("#"): continue`); `_write_env_file` (`src/ormah/setup.py:769`) rewrites the entire file as bare `KEY=VALUE` lines. Result: on any setting change, all comments, section headers, blank-line grouping, and formatting in `~/.config/ormah/.env` are silently destroyed.

## Root cause

`_merge_json_file` (`src/ormah/setup.py:133-152`):

```python
for key, value in updates.items():
    if isinstance(value, dict) and isinstance(existing.get(key), dict):
        existing[key].update(value)   # shallow: replaces list values wholesale
    else:
        existing[key] = value
```

It merges one level deep. That's fine for `mcpServers` (keyed by server name → only the `ormah` key is touched, co-tenants preserved), but wrong for `hooks`, where each event maps to a *list* of hook groups: `dict.update` swaps the whole list.

## NOT affected (already surgical — for reference)

- MCP registrations (Claude Code / Desktop / Codex): keyed by `"ormah"` server name; only own key touched.
- Codex TOML (`_enable_codex_feature`, `_upsert_codex_mcp_config`): surgical key/table upsert.
- CLAUDE.md (`_install_markdown_block`): sentinel-wrapped, own block only.
- All **uninstall** paths (`src/ormah/setup.py:700-726`, `:1255-1285`): per-matcher filter via `_is_ormah_hook`, preserve co-tenants.

## Proposed fix

1. **Hooks (Claude + Codex):** make install symmetric with uninstall — append Ormah's hook to the existing event list after stripping prior Ormah entries (reuse `_is_ormah_hook`), instead of replacing the list. Idempotent, preserves co-tenant hooks. Factor a shared `_merge_hooks(existing_hooks, ormah_hooks)` helper used by both `configure_claude_hooks` and `configure_codex_hooks`.
2. **`.env`:** preserve comments, blank lines, and ordering — update only the `KEY=` lines Ormah owns and append new keys, rather than rebuilding the file from a `dict`.

## Repro (hooks)

1. Add any hook to `~/.claude/settings.json` under `UserPromptSubmit` (e.g. another tool's `command`).
2. Run `ormah setup`.
3. The non-Ormah `UserPromptSubmit` hook is gone.

AndreLFSMartins:
## Summary

Fixes #70 — `ormah setup` overwrote pre-existing, user-owned content in external config files instead of surgically merging only Ormah's own entries.

Two root causes:

- **Hooks (`~/.claude/settings.json`, `~/.codex/hooks.json`):** `_merge_json_file` did a one-level merge that *replaced the whole list* under each hook event, wiping any third-party (co-tenant) hook registered under the same event (`UserPromptSubmit` / `PreCompact` / `SessionEnd` / `Stop`).
- **`.env` (`~/.config/ormah/.env`):** `_write_env_file` rebuilt the file from a dict, destroying all comments, blank lines, and ordering on every write.

## What changed

- **`_is_ormah_hook`** is now a single module-level, **argv-aware** predicate (was duplicated and substring-based). It recognizes both the CLI form (`…/ormah whisper inject|store`) and the plugin wrapper form (`…/ormah-whisper-inject|-store`), so a third-party command that merely *contains* the substring is no longer misclassified. It is **total** — never raises on malformed input. Reused by both install and uninstall.
- **`_merge_hooks`** (new, pure): strips only Ormah's own hook entries per claimed event, preserves every co-tenant hook, appends Ormah's, and leaves unclaimed events untouched. Idempotent. Only drops a matcher it actually emptied by removing Ormah hooks; never drops a matcher it didn't modify.
- **`_install_hooks`** (new): the shared, **fail-closed** install path used by both `configure_claude_hooks` and `configure_codex_hooks`. It refuses to write (warns, returns `False`) when the config is unparseable, not a JSON object, has a non-object `hooks` section, or has a non-list value under a claimed event — and a `try/except` backstop guarantees it never crashes setup nor partial-writes on any other malformed structure. The success message (and, for Codex, the feature-flag enable) is gated on an actual write.
- **`_atomic_write`** (new): temp file in the same dir + `os.replace`, with `chmod` applied before replace — no partial-write corruption, and the `.env` `0600` permission is preserved.
- **`_write_env_file`** rewritten to preserve comments, blank lines, and ordering: existing `KEY=` lines are updated in place, keys absent from the dict are dropped, new keys appended, and the file written atomically with `0600`.

`_merge_json_file` is retained — it is still correct for the `mcpServers` callers (keyed by server name, not a list to merge).

## Known behavior (intentional)

- `_atomic_write` uses `os.replace`, which swaps the inode. A config file that is a symlink/hardlink to another location is detached after the first write — standard atomic-write trade-off, fine for these config paths.
- A `.env` hand-edited with duplicate `KEY=` lines is collapsed to one line on the next write.
- An inline trailing comment on a key whose *value* Ormah rewrites is dropped; full-line comments and untouched keys keep theirs.

## Test plan

- `tests/test_setup.py` / `tests/test_setup_json.py`: added coverage for co-tenant preservation (all claimed events), idempotent re-runs, fail-closed on corrupt / non-object / non-list-event / non-iterable-hooks config (file byte-unchanged, no false "installed" message, Codex feature flag not enabled), malformed/non-dict/non-string-command hook entries preserved without crashing, `.env` comment/order preservation, duplicate-key collapse, and `0600` preserved on replace.
- Full suite green except 9 pre-existing, environment-specific failures unrelated to this change (real `claude`/`codex` MCP subprocesses + fastembed cache dir).
- `ruff check` clean.

### DIFF
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -868,13 +868,34 @@ def _read_env_file() -> dict[str, str]:
 
 
 def _write_env_file(env: dict[str, str]) -> None:
-    """Write env dict to the global config file with secure permissions."""
-    ENV_DIR.mkdir(parents=True, exist_ok=True)
-    lines = []
+    """Write env dict to the global config file, preserving comments and ordering.
+
+    Existing KEY= lines are updated in place; keys absent from `env` are dropped;
+    full-line comments, blank lines, and ordering are kept verbatim; new keys
+    append at end.
+
+    CONTRACT: callers MUST pass the FULL env (from `_read_env_file()`) unless they
+    intentionally want absent keys removed — a partial dict deletes the missing
+    user keys. Non-goal: an inline trailing comment on a key whose VALUE this call
+    rewrites is dropped (full-line comments and untouched keys keep theirs).
+    """
+    original = ENV_PATH.read_text().splitlines() if ENV_PATH.exists() else []
+    seen: set[str] = set()
+    out: list[str] = []
+    for line in original:
+        stripped = line.strip()
+        if stripped and not stripped.startswith("#") and "=" in stripped:
+            key = stripped.partition("=")[0].strip()
+            if key in env:
+                out.append(f"{key}={env[key]}")
+                seen.add(key)
+            # key removed by caller -> drop the line
+        else:
+            out.append(line)  # comment / blank / other -> verbatim
     for key, value in env.items():
-        lines.append(f"{key}={value}")
-    ENV_PATH.write_text("\n".join(lines) + "\n")
-    os.chmod(ENV_PATH, 0o600)
+        if key not in seen:
+            out.append(f"{key}={value}")
+    _atomic_write(str(ENV_PATH), "\n".join(out) + "\n", mode=0o600)
 
 
 def generate_server_wrapper(ormah_bin: str) -> Path:
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -6,6 +6,7 @@ import contextlib
 import glob
 import json
 import os
+import shlex
 import shutil
 import socket
 import subprocess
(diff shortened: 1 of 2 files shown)


r-spade:
Hey Andre, thanks for this,  I went through the issue and PR in detail, and the core fix looks solid.

While reviewing it I found two smaller preservation edge cases and pushed a follow-up commit, 7e4c8a4, directly onto your branch:

- Atomic writes now update a symlink target without replacing the symlink itself, so dotfile-managed configs stay linked.
- Uninstall now drops a matcher group only when it actually removed an Ormah hook. Empty, missing, or malformed hook groups are otherwise left untouched, using the same shared logic for Claude and Codex.

I also moved the uninstall writes through the atomic writer and added regression tests for both cases.

Could you take a look at 7e4c8a4 and chime in on whether the changes fit your intended preservation behavior? Once you are happy with it, I am planning to merge and release the fix as a priority.

AndreLFSMartins:
Reviewed 7e4c8a4 — LGTM, both changes match the intended preservation behavior. 👍

**Symlink-preserving atomic write:** resolving `realpath` only when the path is a symlink, then `os.replace` onto the resolved target, is exactly right — dotfile-managed configs keep their link identity instead of getting silently converted to plain files. The relative-symlink regression test covers it well (link survives, target rewritten, mode applied).

**`_strip_ormah_hooks` uninstall refactor:** consolidating both clients onto one helper is a clear win, and the "only rewrite a matcher that actually lost an Ormah hook" invariant is the correct guard — empty/missing/malformed `hooks` values now pass through verbatim, and a matcher is only dropped when its own `event_changed` fired. Routing the uninstall writes through `_atomic_write` closes the same mid-write corruption gap as the install path. This is the symmetric completion of the #70 fix.

**Verified locally** on the PR head:
- 172 setup tests pass; the 5 new tests (`TestStripOrmahHooks`, symlink preservation, empty/missing matcher preservation) all green.
- Ruff clean.
- Full suite: 1049 passed. The remaining failures are all pre-existing/environmental (I reproduced every one on the pre-7e4c8a4 base f77c7a7): the stale-base auto-linker test you flagged (fixed on main), plus MCP/fastembed-cache/`detect_clients` tests that assert against a machine with no real Claude/Codex configs. None trace to this commit.

Happy with it — good to merge and release. Thanks for catching both edge cases and for the shared-logic cleanup.

r-spade (approved):
LGTM. Thanks a ton for catching and fixing this !!!

PR #75, opened 2026-07-01
feat(whisper): Wave 2 — silence instrumentation + absolute-signal injection gate

r-spade:
Second of the audit PRs (stacked on #74). Fixes the audit's root-cause finding (C2: the injection gate applies an absolute threshold to a rank-relative score → 96% injection rate, gate effectively decorative) and makes silence measurable (I10).

## I10 — silence & outcome instrumentation

- New `whisper_decisions` table: **exactly one row per whisper call, including silent ones** — outcome (`injected` / `silent_short` / `silent_conversational` / `silent_topic_shift` / `silent_no_candidates` / `silent_gate` / `silent_blackout` / `silent_error`), candidate/injected counts, max absolute gate score. Stores `prompt_hash` only.
- `GET /agent/whisper-stats`: silence_rate + injection_rate (they partition all prompts), outcome breakdown, rolling window. Named to avoid colliding with the F09 `/agent/stats` counter, whose `whisper_log` semantics are untouched.

## I2 — the score contract

**Gates cut absolute signals; blended scores only order.**

- `reranker.py` attaches `ce_absolute` (CE score linearly rescaled to [0,1]); `hybrid_search.py` attaches `raw_cosine` (pre-penalty similarity).
- The injection gate now cuts `ce_absolute` (fallback `raw_cosine` when the reranker didn't run) **plus the affinity delta**, so learned feedback can still lift/suppress across the gate. Blended score remains the ordering key.
- The pre-rerank floor passes a candidate if **either** blended score or raw cosine clears it. Found live: question-mode blending (85% cosine weight) + the length penalty inverted ranking — a long runbook had the *highest* raw cosine (0.74) and the *lowest* blended score (0.22), so it never reached the reranker and whisper stayed silent while a perfect memory existed.
- `config.py` now documents which score-kind every threshold cuts.

## Calibration

Sweep of gate ∈ {0.40…0.67} × min_score ∈ {0.30, 0.45}: **flat** on the golden corpus — CE scores there are bimodal (targets > +5 raw, noise < −5), so 0.50 (≙ raw CE −3) is kept as a principled midpoint. Production calibration is exactly what `whisper_decisions.max_gate_score` now records. The residual FP rate at even gate=0.95 traces to the exploration slot bypassing the gate — Wave 3's fix (I6).

## Eval (53 prompts, incl. 2 new gate cases)

| metric | before | after |
|---|---|---|
| F1 | 0.884 | 0.889 |
| precision | 0.895 | 0.900 |
| FP rate | 0.120 | 0.113 |
| suppression | 6/7 | 7/8 |

New golden cases: `whisper-gate-001` (plausible-but-irrelevant memories must stay silent — the case the old gate mathematically couldn't reject) and `whisper-longdoc-001` (long-doc rescue — **silent before this PR**, verified end-to-end now injecting exactly the runbook with zero noise).

Unit tests: `TestGateScoreContract` (high-blend/low-CE rejected, low-blend/high-CE injected, cosine fallback, legacy fallback), `TestWhisperDecisions` (one row per call, per-outcome), whisper-stats endpoint tests.


r-spade:
Addressed codex review (rebased stack, force-pushed). **#75 fixes:** reranker now judges the effective context-enhanced/intent-overridden query (not the bare prompt), so continuation prompts like "and the second one?" are gate-scored with their context — new regression test proves it. Also reverted the identity-gate exemption from the prior round: golden eval showed it fail-opened (id-007 injected a UI pref on "what's my phone number"), and the identity silences it targeted are caused by wave3's fail-closed topical filter, not the gate. Kept the raw_cosine-sentinel fix, identity-independent, plus space/confidence gate demotion. Full suite green; whisper eval continuation precision 0.88→1.00.

PR #76, opened 2026-07-02
feat(whisper): Wave 3 — exploration discipline, fail-closed topical filter, served-topic memory

r-spade:
Third audit PR (stacked on #75). Silence-principle integrity: the three mechanisms that could break or wrongly enforce silence.

## Changes

- **I6 — exploration respects silence**: the exploration slot now requires at least one gate-passed result (piggyback, never manufacture an injection from silence) and renders as `**[exploring]**` instead of masquerading as a confident whisper.
- **I7 — topical filter fails closed**: no-token-overlap candidates need a voucher — identity link, `ce_absolute ≥ whisper_no_overlap_ce_floor` (0.45), or `raw_cosine ≥ 0.70` when the reranker is off. Temporal/follow-up prompts and recency-vouched results keep pre-contract behavior (their raw prompt is underspecified by design; a CE verdict against it isn't a fair voucher — first attempt without this exemption cratered temporal recall 1.0→0.25 in eval, caught and fixed before commit).
- **I9 — topic-shift suppression requires a served topic**: the "same topic → skip" only fires when the session already had an injection on a similar topic (read from `whisper_log` — no new state channel, works on every entry path, survives restarts). A gate-rejected first prompt no longer condemns the whole conversation thread.

## The honest gate recalibration

Wave 2's gate sweep looked flat (0.40–0.67 identical) — **because the exploration slot was re-injecting gate-rejected candidates unmarked**, masking the gate from the eval entirely. With exploration disciplined, the sweep became decisive:

| gate | floor | F1 | FP rate | suppression |
|---|---|---|---|---|
| 0.40 | 0.45 | 0.889 | 0.113 | 7/8 |
| **0.45** | **0.45** | **0.889** | **0.094** | **8/8** |
| 0.50 | 0.61 | 0.867 | 0.094 | 8/8 |

`whisper_injection_gate` = 0.45 (raw CE −3.9): real paraphrase matches land ~raw −3 (the "why did we choose this embedding model" case scores −3.2 against a memory that says "Chose bge-base…" — zero shared tokens), true noise sits below −5.

## Eval vs Wave 2

F1/precision/recall **held** (0.889 / 0.900 / 0.911); FP rate **0.113 → 0.094**; suppression **7/8 → 8/8**; identity FP **0.5 → 0.375**. The previous decision-category "pass" via unlabeled exploration of a gate-rejected candidate is now an honest pass via a correctly calibrated gate.

New tests: exploration-silence, fail-closed filter (weak-CE dropped / strong-CE paraphrase survives / cosine voucher), served-vs-starved topic-shift (3 cases against a real whisper_log).


r-spade:
Addressed codex review. **#76 fix:** builder default `whisper_no_overlap_ce_floor` 0.61 → 0.45 to match production config (direct callers/tests now exercise the shipped value); doc + comment aligned. Verified the companion `whisper_no_overlap_cosine_floor` (0.70) had no drift. 97 whisper tests green.

PR #77, opened 2026-07-02
feat(recall): Wave 4 — recall relevance floor + reranker graceful degradation

r-spade:
Fourth audit PR (stacked on #76). The deliberate-recall fixes and the first-run whisper blackout.

## I8 — recall stops padding with cross-space noise

Two compounding bugs fixed:
- **Space penalty after truncation**: `search()` truncated to `limit` before `_apply_space_scores` ran, so the ×0.6 cross-space penalty could only reorder survivors. Now: fetch a **3× pool → space scores → floor → cut**. A current-space match at raw rank 11 survives; a penalized cross-space hit doesn't.
- **No floor**: recall padded to `limit` with ~0.30 noise. New `recall_min_relevance_score` (0.35) drops sub-floor results — recall legitimately returns fewer. Temporal supplements exempt (recency-vouched by design).

Details that mattered:
- **Whisper bypasses the floor** (`min_relevance=0.0`): the floor initially re-killed the Wave-2 long-doc rescue (blended 0.22, raw cosine 0.74) before the reranker could see it — caught by the `whisper-longdoc-001` golden case, fixed before commit. Whisper has its own floors + absolute gate.
- FTS-only fallback gets the pool widening but no floor — BM25 scores aren't on a [0,1] scale.
- The recall eval runner now passes `default_space` like every production caller, so space scoring is actually measured.

## I4 — no more first-run whisper blackout

Reranker unavailable (model still downloading, or failed) previously returned **empty** — whisper dark for the entire first-run window. Now it degrades to embedding-only whisper with a raised cosine-scale gate (`whisper_injection_gate_no_reranker` = 0.60; the gate cuts `raw_cosine` on this path per the Wave-2 score contract, and bge's noise floor is ~0.5).

## Eval

| | full pipeline | degraded (no reranker) |
|---|---|---|
| F1 | 0.889 (= Wave 3) | 0.852 (was: **silent**) |
| FP rate | 0.094 | 0.075 — *quieter*, as designed |
| suppression | 8/8 | 8/8 |

Recall eval: precision@8 **0.61 → 0.64**, recall@8 1.00 held, MRR 0.95. New golden case `recall-pad-001` reproduces the audit's live padding case ("ormah monetization pricing" query + bitcoin-space distractors) — returns exactly the 3 relevant results, zero padding.


r-spade:
Addressed codex review. **#77 fix:** confirmed temporal supplements bypassed space prioritization (get_recent_nodes orders by recency across all spaces; nothing re-applied space scoring). `_supplement_temporal` now carries the same current/global/other space factor and orders by (space_factor, created); whisper's temporal re-sort is space-first; cross-space regression test added. Also documented that graph-activated results intentionally bypass the relevance floor. 343 engine tests green.

PR #78, opened 2026-07-02
perf(whisper): Wave 5 — encode once per call, evict dead session buffers

r-spade:
Final PR of the audit series (stacked on #77). Pure performance hygiene, behavior-neutral.

- **I15 — encode once**: `prompt_vec` is computed once up front and reused by topic-shift, affinity boost, and whisper_log — previously the same prompt went through the embedding model twice (topic-shift + prompt_vec). Per-whisper `encode()` calls drop 3→2. The third encode (`encode_query` inside search) is *deliberately* not deduplicated: providers apply query prefixes, so `encode()` and `encode_query()` vectors aren't interchangeable — reusing one for the other would silently change retrieval. Verified behavior-neutral: eval identical to Wave 4 (F1 0.889, FP 0.094, suppression 8/8).
- **I12 — session buffer eviction**: `_session_buffers` entries whose newest prompt is past the session gap are evicted on access (runs on the event loop, no lock needed). Previously dead sessions leaked forever on long-running servers.

Tests: encode-count assertion (exactly one `encode()` per whisper), route-level eviction test (dead evicted, live and current kept).


r-spade:
Addressed codex review. **#78 fix:** confirmed `classify()` encodes the identical string with the identical encoder instance, so the old "encode once" claim was really 3→2. Took the preferred path: `PromptIntent` now carries the classifier's vector and the whisper path reuses it — genuinely one encode per call. Renamed the misleading test to count real encoder calls with the classifier active. 344 tests green.

PR #79, opened 2026-07-02
feat(ingestion): ingest-only LLM provider — headless Claude CLI extraction (subscription, no API)

### LINKED ISSUE #73: feat(ingestion): ingest-only LLM provider — headless Claude CLI extraction (subscription, no API), maintenance untouched
## Problem

Server-side transcript extraction (`_extract_memories_llm`) is driven by whatever `llm_provider` is
configured — in practice a local Ollama model (e.g. `gemma3`). That has two costs:

1. **Quality/latency:** a small local model does the memory extraction even when the user's own
   session is driven by a far more capable model. Extraction quality is the floor of the whole
   memory graph.
2. **Resource churn:** keeping a local extraction model resident (or reloading it on each live
   `/ingest/conversation`) burns RAM/GPU on the user's machine continuously through the day, not just
   during scheduled maintenance.

There is currently no way to route **extraction** to a different, higher-quality LLM without also
migrating the background maintenance jobs (auto_linker / consolidator / conflict_detector /
duplicate_merger), which share the single `llm_provider`.

## Proposal

Add an **ingest-only** LLM provider seam so extraction can use a headless Claude CLI (`claude -p`,
subscription auth — **no paid API key required**) as Haiku, independent of the session model and
independent of the maintenance LLM.

Key points:

- **New `claude_cli` `LLMAdapter`** that shells out to `claude -p --model haiku --output-format json`,
  feeding the transcript on **stdin** (never argv — avoids leaking conversation text to the process
  list and `ARG_MAX` failures on large transcripts).
- **Separate `ingest_llm_provider` setting** used only by `_extract_memories_llm`. `llm_provider`
  (maintenance) is untouched — this is a hard non-goal, so maintenance never gets silently migrated.
- **Subscription-only:** the adapter strips `ANTHROPIC_API_KEY` from the child env so it can never
  bill the paid API. Fails to defer (returns `None`) rather than falling back to a paid path.
- **Recursion guard:** the `claude -p` child writes its own session transcript under
  `~/.claude/projects/<encoded-cwd>/`, which the session-watcher would otherwise ingest → runaway
  self-extraction. The watcher must exclude the extractor's own transcript at its single ingest
  chokepoint AND on the live FSEvents path. (On macOS the cwd is symlink-resolved: `/tmp` →
  `/private/tmp`, so the encoded dir name must use the real path.)
- **Untrusted-input boundary:** the transcript is a prompt-injection vector; the worker runs with all
  agent tools denied (`--allowed-tools ""`) — it can emit text, never act. Hooks are disabled on the
  child (`--settings '{"hooks":{}}'`) so it does not re-trigger ingestion.
- **Fail-to-defer, no silent loss:** any extraction failure returns `None`, the ingest cursor does
  not advance, and the next watcher scan retries. The whisper-out hook defers to a **running**
  watcher (checked live via `/admin/health`), and refuses to advance its own cursor on a
  200-with-`status:error` ingest response.

## Non-goals

- No gemma fallback for extraction (Claude-or-defer).
- No migration of background maintenance jobs off their current LLM path.
- No change to whisper-in / recall.

## Notes / open questions for maintainers

- This is initially macOS-focused (transcript path encoding, launchd deployment). The provider seam
  itself is portable; the recursion-guard path encoding needs a per-platform check.
- Would you prefer the ingest-provider seam generalized (any provider override for extraction) rather
  than `claude_cli`-specific? The adapter is one concrete provider behind a generic
  `get_adapter(settings, provider=...)` override, so a `litellm`/`ollama` ingest override falls out
  for free.

I have a working design + spike (subscription auth confirmed headless without an API key, envelope
shape captured, tool-deny + hooks-off verified, transcript recursion path reproduced) and am
implementing this on a fork. Happy to open a PR if the direction is welcome, or adjust the seam to
match your preferences first.

AndreLFSMartins:
Closes #73.

## What

Adds headless `claude -p` (subscription OAuth, **no paid API key**) as an **opt-in** server-side transcript-extraction LLM provider, routed through a generic ingest-only seam so background maintenance jobs keep using their existing LLM.

Aligned with the design feedback on #73:
- **Local LLM is not removed.** `claude_cli` is one more provider alongside `ollama`/`litellm`. No existing path changes.
- **Recursion is resolved at the source, per-adapter** — `claude -p --no-session-persistence` writes zero transcript under `~/.claude`, so there is **no watcher-guard / workdir / purge machinery**. The watcher stays provider-agnostic.
- **Generic seam:** `ingest_llm_provider` + `ingest_llm_model`. Both empty → fall back to `llm_provider`/`llm_model` (current behavior preserved). When the ingest provider is overridden, the model is **required** (validated).

## How it routes

- Extraction (`_extract_memories_llm`) → new `ingest_llm_generate` → `ingest_llm_provider`/`ingest_llm_model`.
- Maintenance (`auto_linker`, `consolidator`, `conflict_detector`, `duplicate_merger`) → unchanged `llm_generate` → `llm_provider`/`llm_model`.
- `ClaudeCliAdapter` fully behind `LLMAdapter`: subprocess/auth/env/parsing live only inside the adapter. Prompt on stdin, `ANTHROPIC_API_KEY` stripped from the child env (forces subscription), bounded concurrency, `--settings '{"hooks":{}}'` + `--allowed-tools ""` (untrusted-transcript trust boundary), returns `None` on any failure mode.

## Config (new)

- `ingest_llm_provider` (`""` | `ollama` | `litellm` | `claude_cli` | `none`), `ingest_llm_model`
- `claude_cli_timeout_seconds` (120), `claude_cli_bin`, `claude_cli_max_concurrency` (1)

## Hardening (whisper-out hook)

- Cursor is **not** advanced when `/ingest/conversation` returns HTTP 200 `{"status":"error"}` (a `claude_cli` timeout/`is_error` → `None` → error), nor when the body is a non-object — the slice is retried instead of being silently lost. A legitimate empty extraction (`status:"processed"`, `extracted:0`) still advances.
- The whisper-store client timeout covers the `claude_cli` extraction budget when that provider is active (else it fires before the server finishes and the cursor never advances).

## Verification

- Full suite: 1075 passed (10 pre-existing environmental failures in `test_setup*`/`test_hippocampus`, unrelated to this diff).
- End-to-end with real subscription auth: extracts memories from a conversation via Haiku and writes **zero** new transcript (`--no-session-persistence` confirmed), maintenance path untouched.
- `--no-session-persistence` verified on claude 2.1.156.


AndreLFSMartins:
Peer review (cursor + codex) surfaced findings now addressed:

**HIGH (security) — tool boundary was fail-open.** `--allowed-tools ""` is inert: `claude -p` loads the operator's `~/.claude/settings.json`, and an inherited `defaultMode: bypassPermissions` auto-approves every tool in the child. Verified against claude 2.1.156 that the extractor could `Read` arbitrary files from untrusted transcript text. Fixed by a `--settings` override whose `permissions` block replaces the operator's: `defaultMode "default"` (escapes the inherited bypass — the load-bearing key; the `--permission-mode` CLI flag and a bare deny list do **not** override a settings-file bypass), `allow []`, and a bare-name `deny` list. A `"*"`/`"mcp__*"` glob is rejected as invalid on this CLI and silently discards the whole `--settings` block (fail-open), so bare names are used. Subscription auth is unaffected (token in the macOS Keychain). Added a skip-guarded integration test against the real binary. (`d07b189`)

**MEDIUM — cursor advanced on unknown 200 bodies.** The whisper-store cursor advanced on any 200 dict that was not `status:error` (including `{}`, missing status, unknown status). Now advances **only** on an explicit `status:"processed"`; everything else — including a client `ReadTimeout` — is retried, never silently lost. (`ccb24b3`, `1811b2f`)

**Minor — config guard + tests.** Reject `claude_cli_timeout_seconds < 1`; added behavioral tests for unknown/missing status and client-timeout→cursor-intact. (`1811b2f`)

Suite: green in a clean environment. Kept as draft pending your review.

AndreLFSMartins:
Rebased onto main @ 2e76b5b (0.13.6, incl. whisper-waves + #89 consolidation-settings). Resolved conflicts across config.py, consolidator.py (upstream moved consolidation params to Settings via #89 — dropped the now-dead module constants, kept the claude_cli strict-schema + cluster-signature), conftest.py (kept both autouse fixtures), memory_engine.py (upstream's nested stats() + the claude_cli ingest-chunking helpers). Local suite: 9 failures, all pre-existing environmental (subset of clean main's 11) — in fact this branch's `_isolate_settings_from_global_env` fixture FIXES 2 of the env-leak failures. 0 regressions.

PR #80, opened 2026-07-03
feat(eval): local eval gate, whisper miner, fail-below bars, minimal CI

r-spade:
Stacked on #78. F01 "measurement before tuning": makes the eval system big enough to trust, without putting any corpus data in the repo.

## What's here (code only — corpora are local-only and gitignored)

- **`make eval` local gate**: runs whisper + recall golden evals with `--fail-below` bars set just under the honest baseline.
- **`ormah eval whisper run --fail-below f1=…,suppression=…`** (mirrors the recall CLI), plus `--corpus` for alternate corpus files.
- **`ormah eval whisper mine`**: mines provisional eval cases from the live `whisper_log` (strictly read-only), stratified by intent with silent-prompt oversampling; drafts labels from live evidence, marked `provisional` until human review; `ormah eval whisper import-labels` is the sign-off.
- **Recall seeder**: supports `connections` (spread-activation cases), backdated `created`, and explicit `stability` — `updated`/`last_accessed` stay "now" so FSRS decay doesn't silently drop backdated seeds.
- **Minimal CI** (`ci.yml`): pytest + non-blocking ruff. Eval gating deliberately deferred until the corpus stabilises.
- Corpus design + honest-labeling rules documented in `eval/whisper/corpus/README.md`.

## Local corpus expansion (not in the diff, by design)

- Whisper golden: 53 → **100 prompts** targeting weak slices (preference, identity, noise, differentiators). Honest re-baseline: overall F1 0.89 → **0.69** — the corpus got harder, not the pipeline worse. Preference F1 0.42 and identity F1 0.58 are now well-sampled failure surfaces.
- Recall golden: 11 → **25 cases** (pad ratios, floor-boundary, temporal, spread-activation, space-rescue, paraphrase). Baseline: Recall@8 0.97, Precision@8 0.61, F1 0.69.
- 80 mined provisional whisper cases awaiting review.

Labels were fixed before every eval run; no case was relabeled to pass (2 mechanical fixes documented in case notes).

https://claude.ai/code/session_01YTsWTHHDxNqLUtUQgLdXeF


r-spade:
Addressed all codex review findings (force-pushed). **#80 fixes:**
1. Provisional (mined, unreviewed) cases now skipped by `whisper run`; zero confirmed cases exits non-zero; `--include-provisional` for smoke runs.
2. Mining groups by (session_id, prompt_hash), not prompt_hash alone.
3. **Recall contamination excluded**: miner mines only prompts backed by a `whisper_decisions` row (deliberate-recall exposures also land in whisper_log); fails cleanly on DBs predating that table.
4. Deterministic candidate truncation (injected first, then score, then id).
5. Mine DB path from `Settings().db_path`, not hardcoded.
6. Shared `eval/settings.py` pins retrieval knobs to production defaults so recall baselines are env-independent (verified: 0.99/0.46/0.57/0.01/0.64 unchanged — my env wasn't skewing them).
7. Removed the dead capture-session workflow.
8. Regression comparison is k-aware and now also flags rising fp/fn rates.
9. Added miner + provisional-CLI regression tests.
10. Dropped `|| true` from CI lint — cleared all 36 pre-existing ruff findings first so the gate is real.

Full suite 1148 passed (only the known env-local `test_llm_provider_defaults_to_none`); `make eval` passes both gates; ruff clean. Stack rebased in order; no corpus data in any branch.

PR #92, opened 2026-07-10
feat(background): per-run maintenance stats on /admin + staggered LLM job schedules (#90)

### LINKED ISSUE #90: observability(background): maintenance runs don't report backlog/throughput; four daily LLM jobs fire in one synchronized burst
## Problem

Per-run maintenance logs report `llm_calls`, `cap`, `cap_hit` — but nothing about the *remaining* work, so there is no way to tell whether the configured caps converge or the backlog silently grows.

Production example: every observed run logs `cap_hit=True` for the three pairwise jobs, which looks like healthy full utilization — while a manual DB query shows **9,453 of 10,287 nodes have never appeared in any `auto_link_checked` pair**. That divergence is invisible from logs and `/admin/health`.

Separately: the four LLM jobs (auto_linker, conflict_detector, duplicate_merger, consolidator) are all registered with the same default 1440-min interval and no explicit `next_run_time`, so their first run lands together ~24h after boot and every 24h after that — a synchronized daily burst instead of a staggered schedule.

## Proposal

1. **Per-run summary** (log line + `/admin` surface): pairs evaluated, pairs/s, `cap_hit`, and a backlog estimate — nodes behind the watermark for auto_linker; unchecked-candidate estimate for dedup/conflict.
2. **Last-run stats per job** on the health/status payload, so cap calibration is data-driven instead of guessed.
3. **Stagger**: explicit `next_run_time` offsets for the four LLM jobs (e.g. +5/+15/+30/+45 min).

Related: #81 (selection half) — this issue provides the measurement layer needed to calibrate both #81's re-scan fix and the pair-batching/caps work.

AndreLFSMartins:
Closes #90.

## What

The four LLM maintenance jobs were unobservable and fired in one synchronized daily burst. This PR makes each run report per-run stats on `/admin`, and spreads the four schedules.

**Per-run stats.** Each `run_*` returns a stats dict; `tracked()` captures it into `JobTracker.last_stats`, which `routes_admin.py` already surfaces via `snapshot()` — no route changes needed.

| Job | Keys |
| --- | --- |
| `auto_linker` | `nodes_scanned`, `pairs_attempted`, `pairs_evaluated`, `edges_created`, `backlog_nodes`, `duration_s`, `pairs_per_s` |
| `duplicate_merger` | `nodes_scanned`, `pairs_attempted`, `pairs_evaluated`, `proposals_created`, `duration_s`, `pairs_per_s` |
| `conflict_detector` | `candidates_found`, `pairs_attempted`, `pairs_evaluated`, `edges_created`, `duration_s`, `pairs_per_s` |
| `consolidator` | `clusters_found`, `clusters_consolidated`, `duration_s` |

`auto_linker` reports `backlog_nodes` (rows behind the watermark). Dedup and conflict have no cheap persistent backlog measure, so `pairs_attempted` vs `pairs_evaluated` is their per-run signal until persistence lands.

`pairs_attempted` counts LLM calls made; `pairs_evaluated` counts valid decisions returned. The difference is "provider unavailable, or output that violated the schema" — without the split, a total LLM outage reported a full `pairs_evaluated` with a near-zero `duration_s`, i.e. an inflated `pairs_per_s`.

**Staggered schedules.** The four jobs previously passed no `next_run_time`, so APScheduler first-fired them all at `now + interval` — identical periods, identical start, permanent lockstep. They now get nominal offsets of 5/15/30/45 minutes, scaled by one factor shared across all four (`min(1, shortest_configured_interval / 60)`). One shared factor, not per-job: scaling each job by its own interval let jobs with different intervals collide (`auto_link=3min` → `5*(3/60)` and `conflict=1min` → `15*(1/60)` both land at 15s). Offsets stay proportional, hence always distinct; the largest is `0.75 * shortest_interval`, so no first run is pushed to or past one interval. At the shipped 1440-minute defaults the offsets are exactly 5/15/30/45.

Note this prevents the *boot burst*, not recurring overlap: for periods `p` and `q`, `min|Δ| ≤ gcd(p,q)/2` for any offsets, so co-prime intervals will eventually overlap regardless. Preventing that needs mutual exclusion across the LLM jobs — out of scope here, worth its own issue.

## Failures are no longer invisible

The observability goal forced three fixes to failure paths:

- `tracked()` treated any dict return as success. The three pairwise `run_*` catch their exceptions and return `{"error": ...}`, so a failed run advanced `last_success` with `error_count` at 0. It now routes error-bearing results to `record_failure` while retaining the stats.
- The four candidate finders each ended in `except Exception: return []`, making a database/encoder/vector-store failure indistinguishable from "no candidates". They now propagate.
- `get_maintenance_batches` (Claude-in-the-loop phase 1) consequently gets per-finder error capture rather than all-or-nothing propagation: finders fail independently (`consolidator` never touches the encoder), so healthy batches must still reach phase 2. It returns an additive `batch_errors` key, and persists a `maintenance_phase1_errors` marker that phase 2 reads to skip stamping `last_maintenance_run` — otherwise a broken finder silenced the maintenance signal for a whole interval while `/admin` stayed green. `MaintenanceManager._run_phase1` records a tracker failure when `batch_errors` is non-empty, while still letting the healthy batches through.

## Known gap, not addressed here

`MaintenanceManager._run_phase2` has the same shape: `apply_maintenance_results` catches every per-item failure and never re-raises, so a payload where every edge/merge/consolidation fails to apply still returns `{"edges": 0, "merges": 0, ...}` and records success. Fixing it means changing phase 2's return contract. Filed as a follow-up rather than expanded into this PR.

## Verification

`tests/test_background/` 179 passed · `tests/test_engine/` 346 passed · `tests/test_api/` 27 passed. Merged against current `main` (`98ca398`), the combined suites give `554 passed`.

The single remaining failure, `test_llm_judge_disabled_by_default`, reproduces on unmodified `main` — it reads the host's real `~/.config/ormah/.env`. Unrelated to this branch.

Reviewed across four adversarial peer rounds; every fix is covered by a test proven to fail when the fix is reverted.


AndreLFSMartins:
Re-based onto main @ 2e76b5b (was 2 commits behind after 0.13.6 release; clean rebase, no conflicts) so the branch shares the exact upstream tip used by the rest of the sync series. Local suite green (same 11 environmental failures as clean main; 0 new).

PR #93, opened 2026-07-10
perf(background): reuse stored vectors for maintenance probe nodes (#88)

### LINKED ISSUE #88: perf(background): pairwise maintenance jobs re-encode probe nodes instead of reusing stored vectors
## Problem

`auto_linker`, `duplicate_merger` and `conflict_detector` each call `encoder.encode(text)` for every probe node on every run to build the KNN query vector — but that node's embedding already sits in `node_vectors`. The `consolidator` does it right: `vec_store.get(node_id)`.

Because `duplicate_merger`/`conflict_detector` re-scan broadly each run (#81), this is O(scanned nodes) embedding compute per run, growing with the store (10k+ nodes in a production deployment).

## Proposal

Use `vec_store.get(node_id)` for the query vector in the three jobs, falling back to `encoder.encode` only when the stored vector is missing (shouldn't happen after `embedding_backfill`; log a warning if it does).

Small, independent, no LLM/provider implications.

AndreLFSMartins:
Closes #88.

## What

The pairwise maintenance jobs (`auto_linker`, `duplicate_merger`, `conflict_detector`) re-encoded each probe node on every run, even though its embedding already sits in `node_vectors`. They now read the stored vector.

- `stored_or_encoded(...)` returns the stored embedding, re-encoding only on a miss (with a warning — a miss should not happen after backfill).
- `run_auto_linker` reuses the vector it already fetched for its fail-closed watermark check, so the encoder is no longer constructed there at all.
- New `ormah/embeddings/text.py` holds `embedding_text` as the **single source of truth** for probe text. It could not live in `encoder.py` or `vector_store.py`: both import numpy, and `memory_engine` deliberately keeps numpy off its startup import path.

## Behaviour change (intended)

Probe and corpus now speak the same language. Corpus vectors were always built from `embedding_text`, which truncates content to `embedding_max_content_chars` (512); the old probes encoded `title + full content`. For nodes with content longer than 512 chars, **similarity scores change** — the two sides of the KNN were previously misaligned, and this PR aligns them. That realignment is the point of #88, not a side effect.

## Review

Reviewed by the Dev Council (Cursor + Codex, 1 round, 0 criticals).

The adversarial focus was exactly the question above: *do the stored vectors embed the same probe text the removed `encoder.encode` used?* They do not — by design, per the paragraph above. Codex read that divergence as a regression and recommended reverting to full-content semantics; that was rejected, because the old probe was the misaligned side.

The one defect the review found that this PR had genuinely introduced (raised by Cursor) is fixed in b972a37: `stored_or_encoded`'s **fallback** re-encoded the caller's raw, untruncated probe text, comparing a full-content probe against a truncated corpus — reintroducing the exact asymmetry the helper exists to remove, in the one path it exists to serve. The fallback now rebuilds the probe through `embedding_text`.

Note that `duplicate_merger` still builds its own untruncated `text` for `_token_overlap` — truncating it there would have silently changed the composite duplicate score.

## Verification

- `pytest tests/` → `1163 passed, 11 failed`. The 11 failures reproduce identically on a clean worktree of the base commit (`98ca398`) with none of these changes applied: they are environment-dependent (`~/.config/ormah/.env`, real MCP client config, fastembed cache), not a regression.
- `ruff check src/ tests/` → clean.

## Follow-ups (not in this PR)

- `mark_outdated` mutates `content` and calls `index_single`, which preserves the vector (`keep_vectors=True`), without re-embedding. It is the only content mutator that does not re-embed (`update` and `execute_merge` both do). Low severity: the `[Outdated: reason]` suffix only reaches the embedding for nodes under 512 chars.
- `_index_embedding` swallows encode failures and leaves the previous vector in place — a *stale* vector, invisible to the `vec_count < node_count` check. Documented as a comment on #32; see also #38, which implements that issue's delta backfill and retry, and which also touches `memory_engine.py` (currently CONFLICTING with main).
- The new regression tests cover the `_find_*_candidates` paths; `run_auto_linker` and `run_duplicate_detection` are not yet mirrored.

### DIFF
--- a/src/ormah/background/auto_linker.py
+++ b/src/ormah/background/auto_linker.py
@@ -172,7 +172,7 @@ def _find_link_candidates(engine, limit: int = 8) -> list[dict]:
     """
     try:
         from ormah.embeddings.encoder import get_encoder
-        from ormah.embeddings.vector_store import VectorStore
+        from ormah.embeddings.vector_store import VectorStore, stored_or_encoded
 
         settings = engine.settings
         encoder = get_encoder(settings)
@@ -195,7 +195,7 @@ def _find_link_candidates(engine, limit: int = 8) -> list[dict]:
             if not text:
                 continue
 
-            query_vec = encoder.encode(text)
+            query_vec = stored_or_encoded(vec_store, encoder, node["id"], text)
             similar = vec_store.search(query_vec, limit=6)
 
             for match in similar:
@@ -308,7 +308,6 @@ def run_auto_linker(engine) -> None:
     """Incrementally link nodes with seq above the watermark; advance only past
     fully-resolved nodes."""
     try:
-        from ormah.embeddings.encoder import get_encoder
         from ormah.embeddings.vector_store import VectorStore
 
         settings = engine.settings
@@ -316,7 +315,6 @@ def run_auto_linker(engine) -> None:
             logger.debug("Auto-linker skipped: LLM not enabled")
             return
 
-        encoder = get_encoder(settings)
         vec_store = VectorStore(engine.db)
         conn = engine.db.conn
         threshold = settings.auto_link_similarity_threshold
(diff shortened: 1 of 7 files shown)


r-spade (approved):
Thanks Andre. This is a solid and useful performance fix. The need is real: these maintenance jobs were paying per-node embedding cost even though the probe vectors already exist in node_vectors, and the change also lines up the probe representation with the canonical stored embedding text for long memories.

I reviewed the background job paths, the stored-vector fallback helper, and the new regression coverage. I also ran focused tests, the full test suite, and Ruff locally; everything passed. This looks good to merge.

PR #94, opened 2026-07-10
config(background): expose consolidator per-run limits as settings (#89)

### LINKED ISSUE #89: config(background): consolidator limits are hardcoded — expose _MAX_CLUSTERS_PER_RUN etc. as settings
## Problem

The other LLM maintenance jobs have env-configurable per-run caps (`auto_link_max_llm_calls_per_run`, `duplicate_check_max_llm_calls_per_run`, `conflict_check_max_llm_calls_per_run`, all validated `>= -1`). The consolidator's limits are hardcoded in `consolidator.py`: `_MAX_CLUSTERS_PER_RUN = 10`, `_MIN_CLUSTER_SIZE = 2`, `_CLUSTER_THRESHOLD = 0.6`, max 5 nodes per cluster.

In a production deployment (10,287 nodes, ~3,561 in the working tier) the consolidator completes its 10 clusters every single run and cannot be tuned up without a code change, while the working tier keeps growing.

## Proposal

Promote at least `consolidation_max_clusters_per_run` (default 10 — no behavior change) to `Settings`, following the validator pattern of the sibling caps. Optionally also expose min/max cluster size and the cluster threshold.

AndreLFSMartins:
Closes #89.

## What

Promotes the four hardcoded consolidator limits to `Settings`
(`ORMAH_CONSOLIDATION_*`), so operators can tune the maintenance
consolidation pass without code changes. **Defaults preserve the previous
behavior exactly.**

| Setting | Default | Was |
|---|---|---|
| `consolidation_max_clusters_per_run` | 10 | `_MAX_CLUSTERS_PER_RUN` |
| `consolidation_min_cluster_size` | 2 | `_MIN_CLUSTER_SIZE` |
| `consolidation_cluster_threshold` | 0.6 | `_CLUSTER_THRESHOLD` |
| `consolidation_max_cluster_nodes` | 5 | hardcoded cap `5` |

## Hardening

Two safety layers on top of the raw settings:

- **Runtime guard** (`consolidator.py`): if `max_cluster_nodes < min_cluster_size`,
  no cluster can ever be emitted — return early with a warning instead of doing
  useless DB/vector work.
- **Config validation** (`config.py`): reject impossible values at construction —
  `max_clusters_per_run >= 0`, `min_cluster_size >= 2`,
  `max_cluster_nodes >= min_cluster_size`, and `cluster_threshold` in `[0, 1]`
  (also rejects NaN/inf). This closes a misconfiguration where
  `min_cluster_size=0` + `max_cluster_nodes=0` slipped past the runtime guard
  (`0 < 0` is false), emitting single-node clusters that `_apply_consolidation`
  would duplicate and archive — silent memory degradation. The runtime guard
  stays as a complementary layer because pydantic does not validate
  post-construction attribute assignment.

## Tests

- `tests/test_background/test_consolidator.py`: defaults, env override, cap
  propagation into `run_consolidation`, inverted-bounds guard.
- `tests/test_config.py`: validation rejects negative / below-min / out-of-range
  / non-finite / inverted-bounds values.

Local run: `ORMAH_LLM_PROVIDER=none pytest tests/test_config.py tests/test_background/test_consolidator.py` — all consolidation tests green; `ruff check` clean.

### DIFF
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -26,6 +26,14 @@ def _find_consolidation_clusters(engine, limit: int = 4) -> list[list[dict]]:
     max_nodes = s.consolidation_max_cluster_nodes
     threshold = s.consolidation_cluster_threshold
 
+    if max_nodes < min_size:
+        logger.warning(
+            "consolidation_max_cluster_nodes (%d) < consolidation_min_cluster_size (%d); "
+            "no cluster can ever be emitted",
+            max_nodes, min_size,
+        )
+        return []
+
     rows = conn.execute(
         "SELECT id, title, content, space FROM nodes WHERE tier = 'working'"
     ).fetchall()
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -7,20 +7,12 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-# Maximum number of clusters to consolidate per run.
-_MAX_CLUSTERS_PER_RUN = 10
-
-# Minimum cluster size to justify consolidation.
-_MIN_CLUSTER_SIZE = 2
-
-# Cosine similarity threshold for clustering.
-_CLUSTER_THRESHOLD = 0.6
-
 
 def _find_consolidation_clusters(engine, limit: int = 4) -> list[list[dict]]:
     """Find clusters of similar working-tier nodes for consolidation.
 
-    Returns up to *limit* clusters, each a list of node dicts (max 5 nodes).
+    Returns up to *limit* clusters, each a list of node dicts (max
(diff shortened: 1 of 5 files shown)


r-spade (approved):
Thanks Andre. This is a useful fix and I agree the need is real: the consolidator cap was hardcoded while larger memory stores can hit it every run, so exposing these limits through Settings is the right direction. The defaults preserve the existing behavior, and the validation/hardening around the new knobs makes this safer to operate.

I reviewed the diff and ran focused plus full verification locally; this looks good to merge. I will get through a few other PRs today and aim to do a release later today.

r-spade:
@AndreLFSMartins released in the latest version : https://github.com/r-spade/ormah/releases/tag/v0.13.6

Thanks :smile: :raised_hands:

PR #95, opened 2026-07-10
feat(background): batch K pairs per LLM call in sleep-cycle maintenance jobs (#87)

### LINKED ISSUE #87: perf(background): pairwise maintenance jobs send 1 pair per LLM call — batch K pairs per call and denominate caps in pairs
## Problem

`auto_linker`, `duplicate_merger` and `conflict_detector` each call `llm_generate` once **per candidate pair**. Every call re-sends the job's fixed instruction block (~2.5–4.5k chars, ~50–70% of the prompt). The `consolidator` already batches (up to 5 memories in one call) — the multi-item prompt + JSON-schema pattern is proven in-tree, but the three pairwise jobs don't use it.

On providers with per-call overhead this dominates the sleep cycle. Measured on a production deployment (10,287 nodes / 21,162 edges, maintenance running through a headless Claude CLI provider — the subscription-CLI lineage of #73 / PR #79, which spawns one `claude -p` process per call, serialized):

- every observed run hits the cap: `llm_calls=100 cap=100 cap_hit=True` for all three jobs
- wall-clock per 100-call job: 10m46s–38m24s (~6–23 s/pair); a full cycle ≈310 calls ≈ 1h–1h15
- coverage: only 1,138 pairs in `auto_link_checked`; **9,453 of 10,287 nodes have never appeared in any checked pair**. At ~100 pairs/job/run the cycle never converges on a growing store.

## Proposal

1. **Shared batch prompt builder** for the three pairwise jobs: K numbered pairs per call, the fixed instruction block sent once, response schema = array of verdicts keyed by `pair_id`, plus an explicit "judge each pair independently; do not let other pairs influence a verdict" instruction. Each job keeps its own instruction block and verdict item shape.
2. New setting `maintenance_pairs_per_call`, **default 1** → zero behavior change out of the box; K=1 degenerates to today's exact prompt and flow. Operators on per-call-expensive providers set 10–16.
3. **Denominate per-run caps in pairs evaluated**, not LLM calls (`auto_link_max_llm_calls_per_run` & siblings cap calls today; with batching, calls is the wrong unit). Raise defaults accordingly (e.g. 500–1,000 pairs where today's default is 100 calls) and raise `auto_link_max_nodes_per_run` in step, otherwise the node cursor becomes the binding constraint.
4. **Failure ladder**: whole-batch parse failure → split the batch in half and retry (bisect) → single-pair failure → record `error` with the existing 6h backoff. Apply valid verdicts even if some `pair_id`s are missing from the array; missing ones are simply retried next run (not recorded).
5. **Timeout scaling**: `base + per_pair × K`.

## Quality gate

Batch-quality impact is assumed, not measured. Before raising the default K anywhere, run an A/B eval (existing `eval/` harness) comparing single vs batched verdicts on ~100 golden pairs.

## Notes

- Related: #81 — that issue is the *selection* half of convergence (what gets scanned per run); this one is the *throughput* half (pairs judged per run). Both are needed.
- The same production logs show 53 "LLM returned invalid JSON" warnings in auto_linker (13 in consolidator) — paid calls, discarded. The bisect ladder in (4) also contains that failure mode.

AndreLFSMartins:
Closes #87. Sibling of #92 (both from the sleep-cycle performance work).

## Problem
The three pairwise LLM jobs (auto_linker, duplicate_merger, conflict_detector) send **one pair per LLM call**, re-sending the fixed instruction block (~50–70% of the prompt) every time, through a **new `claude -p` process per call, serialized** (`claude_cli_max_concurrency=1`). Observed: ~6–23 s/pair, 10–38 min per 100-call job, ~1h–1h15 per full cycle. Every production run hits `cap=100 cap_hit=True` while the backlog is ~9.4k of ~10.3k nodes never checked — at this throughput the cycle **never converges**.

## Change
**Batch K pairs per LLM call** — the consolidator already proves the multi-item prompt + JSON-array pattern internally. One code path with a **per-provider batch-size knob** (not two implementations). Same wall-clock, ~3–5× throughput; fixed-prompt tokens drop ~90%.

Estimated (claude_cli, batch=10): 100 pairs go from ~22 min to ~4–7 min; inverted, the same 22 min buys ~300–500 pairs instead of 100.

## Notes
- Rebased onto main @ 2e76b5b; local suite green (same 11 environmental failures as clean main; 0 new).
- Conflict-resolution notes vs #88 vec-reuse / #90 exceptions are in the #92 discussion.
- Follow-up (tracked): redefine the cap in **pairs** rather than calls and raise it, so throughput ≥ arrival and the backlog drains.

### DIFF
--- a/eval/maintenance/cli.py
+++ b/eval/maintenance/cli.py
@@ -1,0 +1,43 @@
+"""CLI: python -m eval.maintenance.cli {mine|run|report} ...
+
+Local A/B eval gate for #87. Mines auto-link candidate pairs from a production
+store (read-only), runs them through the single and batched paths, and reports
+whether the batched verdicts agree with the single ones (gate for raising K).
+Only the aggregate report is safe to share — never the mined pairs.jsonl.
+"""
+import argparse
+import json
+import sys
+
+from eval.maintenance import miner, report, runner
+
+
+def main() -> int:
+    p = argparse.ArgumentParser(prog="eval.maintenance")
+    sub = p.add_subparsers(dest="cmd", required=True)
+    m = sub.add_parser("mine")
+    m.add_argument("--db", required=True)
+    m.add_argument("--out", default="eval/maintenance/corpus/pairs.jsonl")
+    m.add_argument("-n", type=int, default=100)
+    r = sub.add_parser("run")
+    r.add_argument("--pairs", required=True)
+    r.add_argument("--mode", choices=["single", "batched"], required=True)
+    r.add_argument("--k", type=int, default=10)
+    r.add_argument("--out", required=True)
+    g = sub.add_parser("report")
+    g.add_argument("--single", required=True)
+    g.add_argument("--batched", required=True)
+    a = p.parse_args()
+    if a.cmd == "mine":
+        print(f"mined {miner.mine_pairs(a.db, a.out, a.n)} pairs -> {a.out}")
+    elif a.cmd == "run":
+        runner.run(a.pairs, a.mode, a.k, a.out)
+    else:
+        res = report.agreement(json.load(open(a.single)), json.load(open(a.batched)))
+        print(json.dumps(res, indent=2))
+        return 0 if res["gate_pass"] else 1
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/eval/maintenance/corpus/.gitignore
+++ b/eval/maintenance/corpus/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/eval/maintenance/miner.py
+++ b/eval/maintenance/miner.py
@@ -1,0 +1,68 @@
+"""Mine auto-link candidate pairs from a production store, read-only (#87 eval).
+
+Pairs are sampled across similarity bands so the A/B corpus exercises the whole
+decision surface, not just obvious matches. The mined file contains PERSONAL
+memory content and must NEVER be committed (see corpus/.gitignore) — only the
+aggregate agreement metrics leave the machine.
+"""
+from __future__ import annotations
+
+import json
+import random
+import sqlite3
+from pathlib import Path  # noqa: F401 -- convenient for callers building paths
+
+import sqlite_vec
+
+BANDS = [(0.65, 0.75), (0.75, 0.85), (0.85, 1.01)]
+
+
+def _connect_ro(db_path: str) -> sqlite3.Connection:
+    conn = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
+    conn.row_factory = sqlite3.Row
+    conn.enable_load_extension(True)
+    sqlite_vec.load(conn)
+    conn.enable_load_extension(False)
+    return conn
+
+
+def mine_pairs(db_path: str, out_path: str, n: int = 100, seed: int = 42) -> int:
+    conn = _connect_ro(db_path)
+    rng = random.Random(seed)
+    nodes = conn.execute("SELECT id, title, content, type, space FROM nodes").fetchall()
+    rng.shuffle(nodes)
+    per_band = n // len(BANDS)
+    got = {b: 0 for b in BANDS}
+    seen: set[tuple[str, str]] = set()
+    out = open(out_path, "w", encoding="utf-8")
+    written = 0
+    for node in nodes:
+        if written >= n:
+            break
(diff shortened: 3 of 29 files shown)


AndreLFSMartins:
Pushed `bcad7e7` — it fixes a regression this PR introduces. I hit it on a live ~12.1k-node store and it silently froze maintenance for days.

## The regression

The rewrite of the node loop changed how a vectorless node is handled. On `main` the node is simply left unresolved and the loop **continues**:

```python
# main
if text and node_vec is None:
    node_resolved = False        # park this node, keep going
```

On this branch it aborts the entire run:

```python
# feat/87-pair-batching
if text and node_vec is None:
    stopped = True
    break                        # <-- the whole run dies on the first vectorless node
```

The comment above it still says *"leave it unresolved and let a later run reprocess it"* — but `break` doesn't leave it unresolved, it stops everything. The fail-closed contract was preserved for the node and lost for the run.

## Why it matters

`node_vectors` can legitimately be empty or partial for a while: mid `full_rebuild`, mid delta-backfill (#32/#38), or after a dimension-mismatch wipe (#115). During that window the **first** node above the watermark has no vector — so every run breaks immediately, the watermark never advances, and nothing after it is ever linked.

What I saw in production: watermark frozen at seq 333726, a 12,135-node backlog, and each run reporting `pairs_attempted=0` after 15 ms. Maintenance "succeeded" daily while doing nothing — no error, no degraded health, just a cursor that stopped moving. Under `main`'s behaviour the same store would have kept linking every node that *did* have a vector.

## The fix

Park the node, continue with the rest:

```python
node_vec = vec_store.get(node["id"]) if text else None
if text and node_vec is None:
    logger.warning("auto_linker: node %s (seq %s) has no vector; watermark parked, "
                   "continuing with later nodes", node["id"][:8], node["seq"])
    active.append({"node": node, "resolved": False, "pending": 0, "collected": True})
    continue
```

The fail-closed guarantee is unchanged and still enforced by the same mechanism: `_advance_watermark()` pops only while `collected and pending == 0 and resolved`, so `resolved: False` blocks the cursor at exactly that node. The difference is that every *later* node in the window still gets its candidates and its edges, and the WARNING makes the parked node visible instead of silent.

New test `test_vectorless_node_skipped_then_heals` in `tests/test_background/test_auto_linker.py`: a vectorless node followed by an embeddable one — asserts the later node still gets linked, and that once the missing vector appears the watermark moves past both.

## Verification

Same command on both sides, cloud suites excluded (missing optional deps in my env):

| | failed | passed |
|---|---|---|
| base `b4a45fe` | 50 | 1183 |
| + `bcad7e7` | 50 | 1184 |

Failure sets **identical** (`comm` on the sorted `FAILED` lines returns nothing either way) — the 50 are pre-existing and environment-dependent (`test_setup*`, `test_config`, a global `~/.config/ormah/.env` leaking into bare `Settings()`). Delta is exactly the new test. `ruff check src/ tests/`: clean.

## Known limitation (not fixed here)

A **permanently** un-embeddable node parks the watermark forever. Because each run selects only `auto_link_max_nodes_per_run` nodes after the watermark, everything beyond that window is starved. Reproduced: two consecutive runs rescanning the same 500 nodes with a 501-node backlog; node 501 is never reached. This is strictly better than the `break` (which starved *everything*, including the current window), but the proper fix is a retry/dead-letter set decoupled from the monotonic cursor. Tracked in #109 — deliberately out of scope here, since it needs a new table and a migration.

AndreLFSMartins:

I have to correct the verification table in my comment above. The conclusion survives, but the evidence I gave for it did not actually test what I claimed it tested.

That means both sides of my "base vs. patched" comparison executed the **same** source. The identical failure sets I presented as evidence of zero regressions were an artifact of the bug — they were guaranteed to be identical, and would have been identical even if `bcad7e7` were badly broken. I caught it only because a test from a newer branch failed with an `AttributeError` on a model field that plainly existed in the file I was looking at.

| | failed | passed |
|---|---|---|
| base `b4a45fe` | 12 | 1221 |
| + `bcad7e7` | 12 | 1222 |

Failure sets identical in both directions (`comm` on the sorted `FAILED` lines returns nothing either way), delta is exactly the new test, and all 12 pre-existing failures are environment-dependent (`test_setup*`, `test_setup_json`, plus `test_config`/`test_session_watcher` reading a global `~/.config/ormah/.env`). **Zero regressions** — same conclusion as before, now actually measured.

The earlier figures (50 failed / 1183 vs 1184) should be disregarded.

PR #105, opened 2026-07-12
feat(sync): mutation stamping and deleted_at tombstones (Sync v1 Step 0)

r-spade:
## What

**Sync v1 Step 0** — the merge-blocking prerequisite defined in E06/E08 §4.1 of the paid-tier plans. Sync's last-write-wins merge needs three invariants this PR establishes:

1. Every content mutation advances `updated` to `now(UTC)` before `file_store.save()` — via a new `MemoryNode.touch_updated()` helper (replaces the 5 hand-rolled datetime assignments too).
2. `soft_delete` stamps `deleted_at` into frontmatter **before** moving the file to `deleted/`, so tombstones carry their deletion time (node-vs-tombstone ordering uses `deleted_at`, never `updated`). Unparseable files fall back to the bare rename with a warning — deletion never crashes.
3. Read-side access metadata (`last_accessed`, `access_count`, FSRS `stability`/`last_review`) explicitly does **not** advance `updated` — reads must not cause sync conflicts.

### Now stamped (was missing)
| Site | Mutation |
|---|---|
| `memory_engine.py` `connect()` | edge append (named explicitly in E08 §4.1) |
| `memory_engine.py` merge neighbor remap | connection retarget removed→kept |
| `memory_engine.py` merge kept-node cleanup | connection prune (only when it actually changes) |
| `memory_engine.py` `undo_merge()` | restore = new mutation event; also clears `deleted_at` so the live node outranks the stale tombstone left in `deleted/` |
| `memory_engine.py` `_link_to_self()` | defines-edge append on self node |
| `memory_engine.py` `_auto_link_node()` | auto-link connection appends on remember |
| `background/importance_scorer.py` | importance change (named in E08 §4.1) |
| `background/auto_linker.py` | edge append |
| `background/auto_cluster.py` | space reassignment |
| `background/consolidator.py` | about_self tag append |
| `background/conflict_detector.py` | contradicts/evolved_from edge append |

### Already stamped (now via `touch_updated()`)
`update_node`, `mark_outdated`, `execute_merge` kept node, `TierManager.promote/demote` (covers decay demotions, core-cap enforcement, consolidator demotions), and — per review — both `_migrate_identity_tiers` phases (the one-time completion flag means they never rerun, so an unstamped tier change could lose permanently to a stale remote under LWW).

### Deliberately NOT stamped
| Site | Why |
|---|---|
| `FileStore.touch_access`, `MemoryEngine._touch_access` | read-side metadata + FSRS review state — must not cause sync conflicts (guarded by tests) |
| `_migrate_fsrs` (one-time FSRS migration) | sets stability/last_review only — FSRS review-state metadata, same class as access tracking |
| new-node creation (`remember`, `_ensure_self_node`) | `updated` set by construction default |
| `undo_merge` snapshot restore | stamped (see above) — listed here for completeness since the snapshot's own timestamps are otherwise preserved |

Legacy files without `deleted_at`/`updated` need no backfill: the E06 merge treats missing stamps as epoch (always lose) and warns.

## Tests

- New `tests/test_engine/test_mutation_stamping.py` — 16 tests, one per stamped path plus negatives (access tracking never advances `updated`; tombstone keeps its old `updated`).
- `test_markdown.py`: `deleted_at` round-trip, omitted-when-None, legacy parse.
- Full suite: **1201 passed** (whisper pipeline suite untouched: 99 passed). `ruff` clean.
- Live smoke on a scratch store: `connect()` advanced `updated` on disk; deleted node's tombstone carries `deleted_at`. 

Known pre-existing failure (unrelated, also fails on clean main): `test_config.py::test_llm_provider_defaults_to_none` reads the developer's real `~/.config/ormah/.env` — test-isolation bug, filed separately.

Part of the Sync v1 effort (E06); ships alone before any merge/sync code per the plan.

### DIFF
--- a/src/ormah/background/auto_cluster.py
+++ b/src/ormah/background/auto_cluster.py
@@ -46,6 +46,7 @@ def run_auto_cluster(engine) -> None:
             node = engine.file_store.load(node_id)
             if node:
                 node.space = most_common
+                node.touch_updated()
                 engine.file_store.save(node)
 
             assigned += 1
--- a/src/ormah/background/auto_linker.py
+++ b/src/ormah/background/auto_linker.py
@@ -306,6 +306,7 @@ def _apply_edge(
                     weight=round(similarity, 2),
                 )
                 mem_node.connections.append(md_conn)
+                mem_node.touch_updated()
                 engine.file_store.save(mem_node)
         except Exception as e:
             logger.debug("Failed to persist connection to markdown for %s: %s", node_a_id[:8], e)
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -280,6 +280,7 @@ def run_conflict_detection(engine) -> None:
                 if mem_node is None:
                     continue
                 mem_node.connections.extend(new_connections)
+                mem_node.touch_updated()
                 engine.file_store.save(mem_node)
             except Exception as e:
                 logger.debug("Failed to persist conflict edge to markdown for %s: %s", nid[:8], e)
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -153,6 +153,7 @@ def _apply_consolidation(
             new_node = engine.file_store.load(new_id)
             if new_node and "about_self" not in new_node.tags:
                 new_node.tags.append("about_self")
+                new_node.touch_updated()
                 engine.file_store.save(new_node)
                 with engine.db.transaction() as tx_conn:
                     tx_conn.execute(
--- a/src/ormah/background/importance_scorer.py
+++ b/src/ormah/background/importance_scorer.py
@@ -106,6 +106,7 @@ def run_importance_scoring(engine) -> None:
         node = engine.file_store.load(nid)
         if node is not None:
             node.importance = round(importance, 4)
+            node.touch_updated()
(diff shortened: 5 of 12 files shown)


r-spade:
Addressed all four review findings in the latest commit:

1. **Identity-tier migration stamping** — accepted; the original "deterministic on every machine" rationale was wrong because the one-time completion flag means the migration never reruns after losing to a stale remote. Both phase-1 and phase-2 tier changes now `touch_updated()`. (`_migrate_fsrs` remains unstamped deliberately — FSRS stability/last_review are review-state metadata, same class as access tracking.)
2. **Phase-2 edge repair** — now appends the repaired `defines` connections to the self node's markdown (stamped) in addition to the DB insert. Regression test proves the edge survives `full_rebuild()`.
3. **Atomic tombstone stamp** — `soft_delete` now stamps via `FileStore.save()` (mkstemp + `os.replace`) instead of `write_text`.
4. **No-op update guard** — `update_node` compares the post-application snapshot to the pre-application snapshot and returns without stamping/saving when nothing changed.

Also correcting the description: the audited inventory is **22** save sites, not 23. Suite: 1204 passed, ruff clean, whisper tests untouched. Three new regression tests in `test_mutation_stamping.py` (19 total).

Issue #107, opened 2026-07-12
ormah uninstall deletes ~/.config/ormah — would destroy cloud.key and the recovery kit


r-spade:
## What

`ormah uninstall` removes `~/.config/ormah` as part of its XDG cleanup (`src/ormah/setup.py` xdg_dirs list, ~1784-1788). Once the paid cloud tier ships, that directory holds `cloud.key` (every age identity the user has ever generated) and `ormah-recovery-kit.md`. Deleting them makes all encrypted cloud backups **permanently unreadable** — zero-knowledge means there is no server-side recovery.

## Discovered

During E02 (encrypted snapshot engine) implementation: the new key lifecycle writes to `~/.config/ormah/cloud.key`, which sits directly in uninstall's deletion path.

## Possible fixes

- When `cloud.key` exists, uninstall should either (a) refuse to delete it and leave `~/.config/ormah` with only key material behind, or (b) require an explicit extra confirmation that names the consequence ("your cloud backups will become unreadable forever").
- At minimum, print the recovery-kit warning and pause.
- Consider moving key material to a dedicated dir treated as user data, not app config.

PR #108, opened 2026-07-12
feat(cloud): encrypted snapshot engine (E02)

r-spade:
## What

**E02 — Encrypted Snapshot Engine**, the zero-knowledge trust core of the paid tier (per E08 §1/§2/§5). Turns a local backup into an encrypted, uploadable, restorable blob with client-side keys.

New `src/ormah/cloud/` package:

- **`crypto.py`** — pyrage (age) wrappers: `encrypt_bytes`/`decrypt_bytes` over recipient/identity lists. Decryption always takes **all** known identities so bundles from before a key rotation stay readable forever.
- **`bundle.py`** — `build_bundle` produces `age( gzip-tar( snapshot ) )` with a `bundle-manifest.json` carrying per-file `{path, size, sha256}` plus sync fields (nullable until E06). `open_bundle` decrypts and extracts through a strict allowlist (`nodes/*.md`, `deleted/*.md`, `backup.json`, manifest only; regular files only; ≤100k members; ≤2GB expanded; duplicate + case-collision rejection) and **hard-fails on any hash/size mismatch, missing, or extra file**. Extraction writes bytes manually from `extractfile()` — tar metadata is never applied, which is strictly stronger than `filter="data"` (noted deviation from E08 §2 wording; also sidesteps the 3.11.0–3.11.3 data_filter availability question).
- **`keys.py`** — `~/.config/ormah/cloud.key` (0600 via the existing `_atomic_write` primitive): every identity ever generated, current first. `store_id` (UUIDv4) at `<memory_dir>/.store_id`. Recovery kit regeneration with all identities, restore instructions, and the "without it, nobody can — including us" warning (no passphrase exists or is mentioned). Rotation is non-destructive; **no `--force` exists**.

CLI: `ormah cloud init [--json] [--import-key PATH]` (refuses re-init, names `rotate-key`) and `ormah cloud rotate-key [--yes] [--json]`, mirroring the backup group's lazy-import/`--json`/confirmation patterns. Config: `cloud_backup_enabled` (default false), `cloud_api_url`; documented in `.env.example`. `.store_id` added to the backup manifest's excluded list.

Deps: `pyrage` (runtime), `boto3` (dev group only — `grep -rn boto3 src/` is empty; the dev-only `scripts/cloud_spike.py` is the sole consumer).

## Non-goals honored (E02 contract)

No service code, no scheduler job, no `--cloud` restore UX, no passphrase anywhere, no change to local backup behavior.

## Verification

- **1234 tests pass** (51 new), ruff clean, whisper pipeline suite untouched.
- Tamper suite: flipped byte / missing file / extra file / missing manifest / unknown format_version all fail loudly; malicious tars (traversal, absolute paths, symlink, hardlink, device node, directory member, duplicates, case collisions, member/expansion limits, non-tar payload) all rejected.
- Key lifecycle: 0600 asserted; init-refuses-overwrite; rotation retains old identities and **pre-rotation bundles still decrypt**; `--import-key` round-trip from a real recovery kit opens a real bundle.
- **Live spike against a real dev R2 bucket passed** (transcript below): the bucket only ever held ciphertext; counts and per-file hashes matched byte-identically after restore; spike object cleaned up.

```
[1] Create a scratch memory store with real nodes         5 nodes written
[2] Local backup (BackupService.create)                   memory_2026-07-12_18-46-12 (5 nodes)
[3] Generate age identity and build encrypted bundle      spike.age (1580 bytes)
[4] Presign PUT and upload with httpx                     HTTP 200
[5] Download; verify object is ciphertext                 starts with age header; no plaintext
[6] Decrypt + hash-verify + extract (open_bundle)         manifest verified: 6 files
[7] Restore into fresh memory dir; rebuild index          5 nodes
[8] Counts and per-file hashes match source               5/5 byte-identical ✓
[9] Clean up spike object                                 deleted
SPIKE OK
```

Also live on the dev machine: `ormah cloud init --json` created key/kit/store_id (0600), re-init refused, live `rotate-key` + post-rotation decryption of pre-rotation data.

## Related

- Filed #107: `ormah uninstall` deletes `~/.config/ormah`, which would destroy `cloud.key` — needs a guard before E05 ships.
- Part of the Sync v1/paid-tier chain: #105 (Step 0 mutation stamping) → **this** → E03 service → E04 accounts → E05 cloud backup → E06 sync.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,12 @@ ORMAH_PORT=8787
 # ORMAH_BACKUP_INTERVAL_HOURS=24
 # ORMAH_BACKUP_RETENTION_COUNT=10
 
+# Encrypted cloud backups (paid tier). Bundles are encrypted on your machine
+# with keys only you hold (see `ormah cloud init`); the service never sees
+# plaintext. Disabled until an account is configured.
+# ORMAH_CLOUD_BACKUP_ENABLED=false
+# ORMAH_CLOUD_API_URL=https://api.ormah.dev
+
 # Embeddings model
 ORMAH_EMBEDDING_MODEL=all-MiniLM-L6-v2
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "websockets>=14.0",
     "python-slugify>=8.0.0",
     "pyyaml>=6.0.0",
+    "pyrage>=1.3.0",
 ]
 
 [project.optional-dependencies]
@@ -48,6 +49,7 @@ litellm = ["litellm>=1.0.0"]
 
 [dependency-groups]
 dev = [
+    "boto3>=1.43.46",
     "pytest>=8.0.0",
     "pytest-asyncio>=0.24.0",
     "pytest-httpx>=0.35.0",
--- a/scripts/cloud_spike.py
+++ b/scripts/cloud_spike.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+"""E02 round-trip spike: snapshot → encrypt → R2 → decrypt → verify → restore.
+
+Dev-only script (boto3 lives in the dev dependency group; it never enters the
+shipped package). Requires a dev R2 bucket:
+
+    export ORMAH_DEV_R2_ENDPOINT="https://<account-id>.r2.cloudflarestorage.com"
+    export ORMAH_DEV_R2_ACCESS_KEY_ID="..."
+    export ORMAH_DEV_R2_SECRET="..."
+    export ORMAH_DEV_R2_BUCKET="ormah-dev"
+
+    uv run python scripts/cloud_spike.py
+
+The transcript this prints is the founding-member demo: the object in the
+bucket is ciphertext, and the restored graph is byte-identical to the source.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import os
+import sys
+import tempfile
+import uuid
+from pathlib import Path
+
+import boto3
+import httpx
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "src"))
+
+from ormah.backup import BackupService  # noqa: E402
+from ormah.cloud.bundle import open_bundle, build_bundle  # noqa: E402
+from ormah.cloud.crypto import AGE_HEADER, generate_identity, recipient_for  # noqa: E402
+from ormah.index.builder import IndexBuilder  # noqa: E402
+from ormah.index.db import Database  # noqa: E402
+from ormah.store.file_store import FileStore  # noqa: E402
+from ormah.models.node import MemoryNode, NodeType  # noqa: E402
+
+STEP = 0
+
+
+def step(msg: str) -> None:
+    global STEP
+    STEP += 1
+    print(f"\n[{STEP}] {msg}")
+
+
+def sha_map(root: Path) -> dict[str, str]:
+    return {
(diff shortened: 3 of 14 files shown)


r-spade:
Addressed all five review findings in the latest commit:

1. **Destination symlink following** — `open_bundle` restructured: members stream into a fresh `mkdtemp` staging directory (nothing pre-existing to follow), hashes verify there, and files move into the destination only afterwards. The destination must now be a new or empty directory — a pre-populated dest (including the planted-symlink repro) is refused outright. Regression tests cover planted symlink, dest-is-a-file, and empty-dir cases.
2. **Missing spike script** — root cause: the repo's `/scripts/` gitignore rule silently dropped it from `git add -A`; it existed and ran locally. Force-added. Also re-ran the live R2 spike against the rewritten extraction path — full round-trip passed again.
3. **Lifecycle atomicity** — new `ormah cloud kit` subcommand regenerates the recovery kit from the existing key file (idempotent recovery for both stranded states); `init`/`rotate-key` failure paths now name it explicitly, with rotate's message flagging that the stored kit is missing the new key.
4. **Memory bound** — extraction is now chunked streaming to disk (1 MiB chunks, hash-while-write); only the compressed tar stays in memory (graphs are MBs per E00). The expansion limit is enforced on **actual streamed bytes**, so a member lying about its size can't bypass it (new test).
5. **Unicode collisions** — `NFKC`-normalize before `casefold`; composed-vs-decomposed `é` test added.

Suite: 1242 passed (8 new tests), ruff clean, `grep -rn boto3 src/` still empty. Live `ormah cloud kit --json` verified on the dev machine.

r-spade:
Addressed all three re-review findings:

1. **Symlinked destination** — `_check_dest` rejects a symlinked `dest_dir` via `is_symlink()` (lstat, not followed) and is re-invoked immediately before the staged files are moved, closing the swap-after-validation race. Regression test plants a symlink to an empty external dir and asserts nothing lands there.
2. **store_id preservation on import** — this was the important one: `cloud init --import-key` now parses the kit's store id (`store_id: <uuid>` line; legacy bare-UUID kits also supported) and installs it via `install_store_id` **before any key material is written** — a conflicting existing `.store_id` aborts the whole import with an explanation, and a fresh machine adopts the kit's id exactly. CLI regression tests cover both the preservation and the conflict-refusal (key file verified untouched after abort).
3. **Encrypted input bound** — new `max_bundle_bytes` cap (512 MB default) checked against the file size before `read_bytes()`; the in-memory ciphertext+plaintext window under that cap is documented in the module as the accepted constraint per E00's "graphs are MBs".

Suite: 1251 passed (9 new tests), ruff clean.

r-spade:
Final finding addressed:

- **Fail-closed parsing** — an explicit `store_id:` line that isn't a valid UUIDv4 now raises before anything is written: no key material lands, no `.store_id` is minted. CLI regression test imports a deliberately damaged kit and asserts both. `None` is returned only when the material genuinely carries no store identifier (bare key file).
- **Strict UUIDv4** — single `_validate_store_id` (RFC 4122 variant + version 4, per E08 §1) now backs `extract_store_id`, `install_store_id`, and `get_or_create_store_id`. The tests' example UUIDs were indeed non-v4 (v5/v6/v7) and only passed because validation was version-blind — corrected to genuine v4 values, plus explicit non-v4-rejection tests.

Suite: 1255 passed (4 new tests), ruff clean. Also verified the strict validator accepts the real store id generated on the dev machine.

Issue #84, opened 2026-07-05
bug(index): full_rebuild is non-atomic (DELETE + reinsert in separate txns) → can persist an empty nodes table; startup rebuild-if-empty then loops


AndreLFSMartins:
## Symptom

After a mid-rebuild failure (in the observed incident, following fd exhaustion — see the companion fd-leak issue), the `nodes`, `nodes_fts` and `edges` tables were found **empty** (`nodes=0`) while the markdown source-of-truth in `nodes/` was fully intact (7395 files). On the next restart the server re-empties: `startup` sees `count==0` and calls `full_rebuild` again.

## Root cause

`IndexBuilder.full_rebuild` (`src/ormah/index/builder.py`) deletes and re-inserts in **two separate transactions**:

```python
with self.db.transaction() as conn:
    conn.execute("DELETE FROM node_tags"); ...; conn.execute("DELETE FROM nodes")
# <-- COMMIT here: the empty tables are already persisted
...
with self.db.transaction():                 # separate transaction
    for path in paths:
        try: self._index_file_nodes_only(path)   # per-file failures are caught + logged
        except Exception as e: logger.warning(...)
```

If the re-insert pass fails for every file (e.g. `path.read_text` raising under fd exhaustion — each caught per-file and logged, so the surrounding transaction still commits with 0 rows), the empty state is persisted. `memory_engine.startup` then re-triggers it, self-perpetuating across restarts:

```python
count = self.db.conn.execute("SELECT COUNT(*) FROM nodes").fetchone()[0]
if count == 0:
    self.builder.full_rebuild()   # empties again if the fault persists
```

Note: the 1-minute Index updater uses `incremental_update`, which **is** atomic (single transaction, rolls back on error) — it is not the culprit here.

## Reproduction

```python
builder.full_rebuild()              # seed the index from a fixture store
before = count_nodes()              # > 0
def boom(_p): raise OSError(24, "Too many open files")
monkeypatch.setattr(builder, "_index_file_nodes_only", boom)
builder.full_rebuild()             # today: commits empty tables, no rollback, no raise
assert count_nodes() == before     # fails today: 0
```

## Proposed fix

- Wrap DELETE + both insert passes in **one** transaction so any failure rolls back to the prior committed state.
- Guard: if 0 nodes are produced while source files exist, `raise` (→ rollback) rather than commit an empty index.
- Make `startup` tolerate that raise (log at ERROR + continue) instead of crash-looping under a KeepAlive supervisor.

[...]

PR #116, opened 2026-07-13
fix(index): refuse to silently DROP a populated vector store on a dimension mismatch (#115)

### LINKED ISSUE #115: bug(index): init_vec_table silently DROPs a populated vector store on a dim mismatch — the 768 default wipes a 1024 store
## Summary

`Database.init_vec_table(dim)` **silently DROPs the entire `node_vectors` table** whenever the configured embedding dimension differs from the stored one. The parameter defaults to `dim=768`, so any boot in which `ORMAH_EMBEDDING_DIM` fails to resolve — a partially-applied `.env`, a missing line, a wrapper that does not export it — falls back to 768 and wipes a 1024-dim store (e.g. `bge-m3`) with no confirmation, no backup, and only an `INFO` log line. Recovery is asynchronous, so the store stays empty for the duration of a full re-embed while search and maintenance run against nothing.

This happened twice in one day on a real 12.1k-node store.

## Evidence (live store, 12,164 nodes)

```
09:53:15  Embedding dimension changed (1024 → 768), recreating vec table   ← DROP
09:53:26  Embedding backfill (delta): embedding 12138 missing nodes
09:58:12  Embedding backfill (delta): embedded=12138 failed=0 missing=0 vec=12138/12138
10:28:52  Embedding dimension changed (768 → 1024), recreating vec table   ← DROP again
10:29:03  Embedding backfill (delta): embedding 12138 missing nodes
```

Between 09:58 (a full, healthy vector store) and 10:29, the only event was a restart. `full_rebuild` did not run (`fts_needs_rebuild` unset, nothing logged). The wipe came entirely from `init_vec_table`'s dimension check.

Downstream, with `node_vectors` empty, `auto_linker` fail-closed-aborts on the first vectorless node above its watermark, so the cursor froze (watermark stuck, 12,135-node backlog, `pairs_attempted=0` in 15 ms). Maintenance "succeeded" daily while doing nothing.

## Root cause

`src/ormah/index/db.py::init_vec_table`:

```python
def init_vec_table(self, dim: int = 768) -> None:
    ...
    if existing:
        try:
            row = self.conn.execute("SELECT embedding FROM node_vectors LIMIT 1").fetchone()
            if row is not None:
                existing_dim = len(row[0]) // struct.calcsize("f")
                if existing_dim != dim:
                    logger.info("Embedding dimension changed (%d → %d), recreating vec table",
                                existing_dim, dim)
                    with self.transaction() as conn:
                        conn.execute("DROP TABLE node_vectors")     # <-- no authorization, no guard
        except Exception:
            pass                                                    # <-- swallows everything
    ...
```

Three compounding defects:

1. **A config default (768) can destroy data.** The dimension is taken on faith from `settings.embedding_dim`; there is no way to distinguish "the operator deliberately changed the embedding model" from "the config failed to load". The destructive branch treats both identically.
2. **The probe cannot see an EMPTY mismatched table.** `SELECT embedding ... LIMIT 1` returns `None` when the table has no rows, so an empty table with the wrong declared dimension keeps its wrong schema and every later insert fails.
3. **`except Exception: pass` swallows any error raised inside the check** — including any guard one might add there.

## Proposed fix

1. **Read the stored dimension from the table DDL** (`SELECT sql FROM sqlite_master WHERE name='node_vectors'`, regex `FLOAT\[(\d+)\]`) instead of probing a row, so an empty mismatched table is detected too. Raise on unparseable DDL (fail closed) rather than booting into a broken vector index.
2. **Refuse to DROP a POPULATED table.** Raise a `RuntimeError` naming the configured dim, the stored dim, and the vector count, and telling the operator either to fix `ORMAH_EMBEDDING_DIM` or to authorize the migration explicitly. An EMPTY table may still be recreated freely (no data to lose).
3. **Authorize a deliberate model migration with an explicit, dim-valued, one-shot token** — e.g. `ORMAH_REINDEX_ON_DIM_CHANGE=<new dim>`, honored only when it equals the configured dim, and **consumed in the database** so a flag left behind in `.env` cannot silently re-authorize a future drop. The consumed dimensions must **accumulate**, otherwise a later migration overwrites the record of an earlier one and revives a spent token.
4. **Do the whole read → decide → drop → consume → recreate sequence inside one `BEGIN IMMEDIATE`**, or two concurrent initializers (e.g. two launchd agents both with `RunAtLoad=true`) can both pass the check and both spend the same authorization.
5. Restrict the broad `except Exception: pass` to the `sqlite_vec` import only, so the guard's `raise` can never be swallowed.

## Relation to other issues

- **#32 / #38** — delta backfill. The backfill is what has to repair the damage after each wipe; this issue is why the damage keeps happening.
- **#109** — the backfill's own durability (progress lost on restart) and the `auto_linker` fail-closed abort that turns an empty store into a frozen cursor. #109 is the *consequence*; this issue is the *trigger*.
- **#84** — non-atomic `full_rebuild` is a second, independent way `node_vectors` can be emptied.

## Deferred follow-ups (surfaced while fixing this)

- A **permanently** un-embeddable ("poison") node parks the `auto_linker` watermark forever. Because each run selects only `auto_link_max_nodes_per_run` nodes after the watermark, everything beyond that window is never processed. Reproduced: two consecutive runs rescanning the same 500 nodes with a 501-node backlog; node 501 never reached. Proper fix is a retry/dead-letter set decoupled from the monotonic cursor.
- **Schema-mode backfill has no per-row progress tracking**, so an interrupted schema pass re-encodes every row on the next run (idempotent — wasted compute, no data loss).

Found while bringing a ~12.1k-node real store back online after it was wiped twice in one morning.

AndreLFSMartins:
Fixes #115.

`init_vec_table` silently DROPs a populated `node_vectors` table whenever the configured embedding dimension differs from the stored one. Since the parameter defaults to `dim=768`, any boot where `ORMAH_EMBEDDING_DIM` fails to resolve falls back to 768 and wipes a 1024-dim store (e.g. `bge-m3`). This happened **twice in one morning** on a real 12,164-node store:

```
09:53:15  Embedding dimension changed (1024 → 768), recreating vec table   ← DROP
09:58:12  Embedding backfill (delta): embedded=12138 missing=0 vec=12138/12138
10:28:52  Embedding dimension changed (768 → 1024), recreating vec table   ← DROP again
```

Downstream, an empty vector store froze the `auto_linker` watermark (12,135-node backlog, `pairs_attempted=0` in 15 ms), so maintenance "succeeded" daily while doing nothing.

## What changed

**1. Read the stored dim from the table DDL, not from a row probe.**
`SELECT sql FROM sqlite_master` + `FLOAT\[(\d+)\]` replaces `SELECT embedding ... LIMIT 1`. The old probe returns `None` on an *empty* table, so an empty table with the wrong declared dimension silently kept its wrong schema and every later insert failed. Unparseable DDL now raises (fail closed) instead of booting into a broken vector index.

**2. Refuse to DROP a POPULATED table.**
A mismatch on a table holding vectors raises `RuntimeError` naming the configured dim, the stored dim and the vector count, and telling the operator what to do. An **empty** table is still recreated freely — there is no data to lose.

**3. The `raise` can no longer be swallowed.**
The old `try: ... except Exception: pass` wrapped the whole check; a guard raised inside it would have been silently discarded. The `try` now wraps **only** the `sqlite_vec` import, with `except ImportError: return`.

**4. Explicit, dim-valued, one-shot authorization for a deliberate model migration.**
`ORMAH_REINDEX_ON_DIM_CHANGE=<new dim>` (new setting, default `0`) authorizes the drop, and **only** when it equals the configured dim — a stale flag naming a different dim cannot authorize anything. The authorization is then **consumed in the database**, so a flag left behind in `.env` cannot silently re-authorize a future drop.

**5. Consumed authorizations accumulate.**
The consumed dims live in `meta.reindex_consumed_dims` as a set. Storing only the last one would let a later migration erase the record of an earlier one and revive a spent token (`4→8` consumes 8; `8→16` overwrites the marker; restoring `dim=8` with the stale flag would then drop a populated 16-dim store).

**6. The whole sequence runs in one `BEGIN IMMEDIATE`.**
Reading the count and the marker *before* the transaction left a check-then-drop race: two concurrent initializers (e.g. two launchd agents both with `RunAtLoad=true` — the exact deployment where this was found) could both observe the marker as absent and both spend the same one-shot authorization. Reproduced with a two-thread probe: two `DROP TABLE` statements executed, and the losing thread hit `OperationalError: no such table`. Now read → decide → drop → consume → recreate is a single serialized transaction.

## Tests

New `tests/test_index/test_init_vec_table_guard.py` (8 tests), all written failing-first:

| test | proves |
|---|---|
| `test_dim_mismatch_on_populated_store_raises_not_drops` | populated + unauthorized → raises, rows preserved |
| `test_dim_mismatch_on_empty_store_recreates` | empty + mismatch → recreated at the new dim (the row-probe blind spot) |
| `test_allow_drop_authorizes_reindex` | explicit authorization performs the migration |
| `test_unparseable_ddl_raises_and_preserves_rows` | corrupt/foreign schema fails closed, rows kept |
| `test_startup_wiring_respects_reindex_flag` | a stale flag naming another dim is still refused |
| `test_authorization_is_one_shot` | the authorization is consumed and cannot be reused |
| `test_consumed_authorization_survives_a_later_migration` | a later migration cannot revive a spent token |
| `test_concurrent_init_vec_table_race_consumes_authorization_once` | two concurrent initializers → exactly **one** `DROP` (counted via a `sqlite3` trace callback synchronized with a `threading.Barrier`; fails on the pre-fix code with 2 drops) |

`tests/test_embeddings/test_adapters.py::TestDimensionMismatch` pinned the old silent-drop behaviour and is rewritten to the new contract.

**Suite:** `22 failed, 1287 passed` on this branch vs `22 failed, 1278 passed` on pristine `main` — the same 22 pre-existing environment-dependent failures (`test_setup*.py`), **zero regressions**, +9 new passing tests. `ruff check src/ tests/`: clean.

## Migration note

Existing deployments are unaffected: a store whose dim already matches the config never enters the mismatch branch. An operator who genuinely changes embedding model now sets `ORMAH_REINDEX_ON_DIM_CHANGE=<new dim>` for one boot and removes it afterwards; the error message spells this out, including how to clear the consumed marker if a second deliberate reindex to the same dim is ever needed.

## Related

- #109 — the backfill's own durability and the `auto_linker` fail-closed abort. That issue is the *consequence* of an empty store; this one is the *trigger*.
- #32 / #38 — delta backfill, which is what has to repair the damage after each wipe.
- #84 — non-atomic `full_rebuild`, a second independent way `node_vectors` can be emptied.

Reviewed by two independent adversarial reviewers over three rounds; findings 4, 5 and 6 above were each found by review and fixed with a failing test first.

### DIFF
--- a/src/ormah/config.py
+++ b/src/ormah/config.py
@@ -47,6 +47,9 @@ class Settings(BaseSettings):
     embedding_provider: str = "local"  # "local", "ollama", "litellm"
     embedding_model: str = "BAAI/bge-base-en-v1.5"
     embedding_dim: int = 768
+    # Set to the NEW dim to authorize ONE deliberate destructive reindex of a
+    # populated vector store (embedding-model migration). Remove after the boot.
+    reindex_on_dim_change: int = 0
 
     # LLM for extraction. Disabled by default; setup can opt into cloud/local providers.
     llm_provider: str = "none"
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -82,7 +82,10 @@ class MemoryEngine:
         self.file_store = FileStore(settings.nodes_dir)
         self.db = Database(settings.db_path)
         self.db.init_schema()
-        self.db.init_vec_table(settings.embedding_dim)
+        self.db.init_vec_table(
+            settings.embedding_dim,
+            allow_drop=settings.reindex_on_dim_change == settings.embedding_dim,
+        )
 
         self.graph = GraphIndex(self.db)
         self.builder = IndexBuilder(self.db, self.file_store)
--- a/src/ormah/index/db.py
+++ b/src/ormah/index/db.py
@@ -537,22 +537,46 @@ class Database:
                 count = self.conn.execute(
                     "SELECT count(*) FROM node_vectors"
                 ).fetchone()[0]
-                if count > 0 and not allow_drop:
-                    raise RuntimeError(
-                        f"Embedding dimension mismatch: configured "
-                        f"ORMAH_EMBEDDING_DIM={dim} but node_vectors holds "
-                        f"{count} vectors of dim {existing_dim}. Refusing to "
-                        f"DROP them. If this is a deliberate embedding-model "
-                        f"change, set ORMAH_REINDEX_ON_DIM_CHANGE={dim} for one "
-                        f"boot (remove it afterwards); otherwise fix "
-                        f"ORMAH_EMBEDDING_DIM — the code default is 768."
+                if count > 0:
+                    consumed = self.conn.execute(
+                        "SELECT value FROM meta WHERE key = 'reindex_consumed_dim'"
+                    ).fetchone()
+                    already_consumed = consumed is not None and consumed["value"] == str(dim)
+                    if not allow_drop:
+                        raise RuntimeError(
+                            f"Embedding dimension mismatch: configured "
+                            f"ORMAH_EMBEDDING_DIM={dim} but node_vectors holds "
+                            f"{count} vectors of dim {existing_dim}. Refusing to "
+                            f"DROP them. If this is a deliberate embedding-model "
+                            f"change, set ORMAH_REINDEX_ON_DIM_CHANGE={dim} for one "
+                            f"boot (remove it afterwards); otherwise fix "
+                            f"ORMAH_EMBEDDING_DIM — the code default is 768."
+                        )
+                    if already_consumed:
+                        raise RuntimeError(
+                            f"Embedding dimension mismatch: a reindex to dim {dim} "
+                            f"was already performed and the ORMAH_REINDEX_ON_DIM_CHANGE "
+                            f"authorization is spent. node_vectors now holds {count} "
+                            f"vectors of dim {existing_dim} — remove the stale "
+                            f"ORMAH_REINDEX_ON_DIM_CHANGE from your config. If a second "
+                            f"deliberate reindex to dim {dim} is genuinely needed, clear "
+                            f"the marker first: DELETE FROM meta WHERE "
+                            f"key='reindex_consumed_dim'."
+                        )
+                    logger.info(
+                        "Recreating vec table (%d → %d, %d vectors dropped)",
+                        existing_dim, dim, count,
                     )
-                logger.info(
-                    "Recreating vec table (%d → %d, %d vectors dropped)",
-                    existing_dim, dim, count,
-                )
-                with self.transaction() as conn:
-                    conn.execute("DROP TABLE node_vectors")
+                    with self.transaction() as conn:
+                        conn.execute("DROP TABLE node_vectors")
+                        conn.execute(
+                            "INSERT OR REPLACE INTO meta (key, value) VALUES "
+                            "('reindex_consumed_dim', ?)",
+                            (str(dim),),
+                        )
+                else:
+                    with self.transaction() as conn:
+                        conn.execute("DROP TABLE node_vectors")
 
         with self.transaction() as conn:
             conn.execute(
--- a/src/ormah/index/db.py
+++ b/src/ormah/index/db.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import re
 import sqlite3
 import threading
 from contextlib import contextmanager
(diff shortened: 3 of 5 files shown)


AndreLFSMartins:

I have to correct the verification table I posted above. The conclusion survives, but the evidence I gave for it did not actually test what I claimed it tested.

That means both sides of my "base vs. patched" comparison executed the **same** source. The identical failure sets I presented as evidence of zero regressions were an artifact of the bug — they were guaranteed to be identical, and would have been identical even if this branch were badly broken. I caught it only because a test from a newer branch failed with an `AttributeError` on a model field that plainly existed in the file I was looking at.

| | failed | passed |
|---|---|---|
| base `4f66abc` (`main`) | 12 | 1288 |
| + this branch | 12 | 1297 |

Failure sets identical in both directions (`comm` on the sorted `FAILED` lines returns nothing either way), delta is exactly the 9 new tests, and all 12 pre-existing failures are environment-dependent (`test_setup*`, `test_setup_json`, plus `test_config`/`test_session_watcher` reading a global `~/.config/ormah/.env`). **Zero regressions** — same conclusion as before, now actually measured.

The earlier figures (22 failed / 1278 vs 1287) should be disregarded.

PR #127, opened 2026-07-14
fix(index): requeue a node only when its content fingerprint changes (#126)

### LINKED ISSUE #126: perf(index): writing an edge re-queues the node — auto_linker feeds itself and the backlog never drains (backlog ≈ store)
## Symptom

The auto_linker backlog sits at ~95% of the store and does not drain, no matter how much throughput you give the job. Measured on a live 14,942-node store: backlog **14,178**. Tuning K (1→10), `claude_cli_max_concurrency` (1→2), `auto_link_max_nodes_per_run` (100/200/500) and the interval changed the *rate* but never the *backlog*.

This is very likely the `backlog ≈ store` finding from the 2026-07-13 deep evaluation, which recorded the symptom without a cause.

## Root cause

`seq` is a **change-sequence**: `IndexBuilder.index_node` allocates a fresh `seq` from `meta.node_seq_next` on **every** index pass (builder.py:176-187). The auto_linker enqueues by `seq > watermark` (auto_linker.py:36), so **any node that gets reindexed goes back to the end of the queue**.

The code states the intended contract itself (builder.py:178-180):

> *Every **content** (re)write lands the node at the head, so reindex/import/restore re-enter the delta regardless of frontmatter timestamps. **Metadata-only UPDATEs elsewhere do not pass through here.***

Writing an edge violates that contract. `_apply_edge` persists the connection into the node's markdown:

```python
mem_node.connections.append(Connection(...))
mem_node.touch_updated()
engine.file_store.save(mem_node)      # rewrites the .md file
```

The file changes → `file_hash` changes → `index_updater` (every 60s) reindexes it → `seq` is bumped → **the node re-enters the auto_linker queue**. But nothing relevant to linking changed: same content, same embedding, same 6 vector neighbours, and the pairs are already in `auto_link_checked`. The reprocessing cannot discover anything.

**The auto_linker re-queues the nodes it just processed.**

`conflict_detector` writes edges the same way, so it has the same effect.

## Evidence (live store, 2026-07-14)

`seq` does not correlate with `created` — nodes from the same day are scattered across the whole range:

```
5 oldest nodes by created (all 2026-06-27):
  seq=766207   <- the highest seq in the store
  seq=439329
  seq=175154
```

The head of the queue is old nodes touched today:

```
nodes with seq > 700,000:                                    13,103
  of those, created < today but updated = today:             11,833   (90%)

seq=766230 | created=2026-07-07 | updated=2026-07-14T20:03  <- while auto_linker was running
seq=766229 | created=2026-07-08 | updated=2026-07-14T20:03
seq=766228 | created=2026-07-09 | updated=2026-07-14T20:03
```

Re-queue rate: **~352 nodes/hour** put back by edge writes (849 edges over 213 distinct source nodes in the sample hour).

## Impact

Not a correctness bug — it converges (a re-queued node whose pairs are all in `auto_link_checked` produces no new edge, so it is not touched again). It is a **throughput tax**: every run spends part of its budget re-scanning nodes that have nothing to gain, while virgin nodes wait. In a sample of the next 300 queued nodes, 50 (17%) had already been judged.

Over a store's lifetime this is what pins the backlog at ≈ store size.

## Fix options

1. **Do not bump `seq` when only connections changed** (recommended). `index_node` compares the incoming content/title against the stored row and preserves `seq` when only the connection block differs. Fixes it at the root and covers `conflict_detector` too.
2. Have the auto_linker not `touch_updated()` when persisting an edge. Smaller, but treats the symptom at one caller and leaves `conflict_detector` broken.
3. Split `content_seq` from `change_seq`, with the auto_linker enqueuing on the former. Most correct, most expensive.

Related: #117 (the edge-write path), #109 (watermark freeze, different cause), #81.

AndreLFSMartins:
Closes #126.

## The bug

`seq` is a change-sequence: `IndexBuilder` reallocates it from `meta.node_seq_next` on **every** index pass, and the auto_linker enqueues by `seq > watermark`. So any reindexed node goes back to the end of the queue.

Writing an edge does exactly that. `_apply_edge` persists the connection into the node's markdown (`touch_updated()` + `file_store.save()`), the `file_hash` changes, `index_updater` reindexes within 60s, and the `seq` is bumped — **the auto_linker requeues the node it just processed**, with nothing to learn, since the pairs are already in `auto_link_checked`.

The code already stated the contract it was violating:

> Every **content** (re)write lands the node at the head... **Metadata-only UPDATEs elsewhere do not pass through here.**

## Evidence (live store, 14,942 nodes)

- backlog pinned at **14,178** (~95% of the store), regardless of throughput
- of the 13,103 nodes with `seq > 700,000`, **11,833 (90%)** were created before today but updated today — old nodes requeued by an edge write
- re-queue rate **~352 nodes/hour**
- `seq` does not correlate with `created`: nodes from the same day carry seq 175,154 / 439,329 / 766,207
- a 64-minute measurement with the linker running flat out: the cursor advanced 20,247 seq positions and the backlog **went up** (14,178 → 14,232). Only 207 new nodes arrived in that window.

## The fix

A **content fingerprint** — `sha256(title, content, type, space)` — is persisted on the node and written **only by the builder**. `seq` is reallocated only when it changes.

Those four fields are exactly what a linking decision depends on: `title`/`content` feed the embedding and the LLM judge prompt, `type` is shown to the judge, and `space` is shown to the judge *and* drives `cross_space_penalty` during candidate selection. They are also precisely the columns `auto_linker` selects per queued node. Connections, tags, tier, importance and access counts cannot change what the linker would decide, so they no longer requeue the node.

Two subtleties, both caught in peer review:

**The baseline must be the persisted fingerprint, never the stored row.** `auto_cluster` writes `space` straight into SQLite *and* into the markdown, bypassing the builder (`auto_cluster.py:64-86`). Comparing the incoming markdown against the stored *row* would find them already equal at reindex time, preserve the `seq`, and **freeze that node out of relinking forever** — strictly worse than the unconditional bump it replaces. A direct `UPDATE` does not touch the fingerprint column, so the mismatch is still caught. Pinned by `test_direct_db_space_update_still_requeues`, which fails if the comparison is done against the row.

**A requeue must invalidate cached verdicts.** The auto_linker skips any pair already in `auto_link_checked` *before* it looks at the edge, so a fresh `seq` alone changes nothing. `memory_engine.update_node` clears that table only for content/title edits — a type/space edit does not, and a direct markdown edit clears nothing at all. The invalidation now happens in the builder, where the change is detected, covering every path into the index (disk edits and sync included). It is gated on `prior is not None`, so `full_rebuild` does not fire one delete per node against a table whose PK has no index on `node_b`.

`full_rebuild` is otherwise unchanged: it empties the table and clears the watermark, so every node is requeued and a restored store is fully relinked.

Fixing it in the builder also covers `conflict_detector`, which writes edges the same way.

## Migration

`content_fingerprint` is added by `_migrate()` and backfilled from the existing rows, so upgrading does not requeue the whole store once.

The backfill **skips rows whose `file_hash` no longer matches the file on disk**. Those have a pending reindex, and if `auto_cluster` already wrote a new `space` into both the row and the markdown, stamping them would bake in a change that was never indexed — the next reindex would see "no change" and the relink would be lost silently. They are left `NULL`, which never equals the incoming fingerprint, so the first reindex requeues them. Pinned by `test_migration_leaves_stale_rows_unstamped`.

## Tests

12 tests in `tests/test_index/test_seq_fingerprint.py`:

- `test_edge_write_does_not_bump_seq` — the bug; fails on `main` (`assert 4 == 2`)
- `test_direct_db_space_update_still_requeues` — the auto_cluster dual-write freeze
- `test_fingerprint_change_invalidates_cached_pairs` / `test_edge_write_keeps_cached_pairs`
- `test_content_/title_/type_change_bumps_seq`, `test_tags_only_change_does_not_bump_seq`
- `test_full_rebuild_allocates_new_seq`
- `test_migration_leaves_stale_rows_unstamped` / `test_migration_stamps_clean_rows`
- `test_indexing_persists_a_content_fingerprint`

The two load-bearing tests were verified by mutation: reverting the implementation to compare the stored row makes `test_direct_db_space_update_still_requeues` fail, and making the backfill stamp every row makes `test_migration_leaves_stale_rows_unstamped` fail.

Suite: `1422 passed`, with the same 6 pre-existing failures as `main` (5 in `test_setup.py`, 1 in `test_space_detect.py` — all environment-dependent). Ruff clean.

## Follow-ups (not in this PR)

- `memory_engine.update_node` still runs its own `DELETE FROM auto_link_checked`, now redundant with (and narrower than) the builder's. Worth removing.
- Removing a never-judged pair's imported edge does not requeue the node. Rejudging a pair whose edge a user deliberately deleted would recreate it, so this is a deliberate trade-off.

### DIFF
--- a/src/ormah/background/auto_linker.py
+++ b/src/ormah/background/auto_linker.py
@@ -328,16 +328,26 @@ def run_auto_linker(engine) -> None:
         threshold = settings.auto_link_similarity_threshold
         cross_space_penalty = settings.auto_link_cross_space_penalty
         max_edges = settings.auto_link_max_edges_per_run
+        # #126: created only counts POSITIVE edges — a 'none'/'error' verdict still costs an
+        # LLM call but was free of budget. Before this PR that was harmless (cached pairs were
+        # skipped), but invalidating verdicts on content change makes those pairs eligible
+        # again: a bulk space/content change can now issue thousands of judging calls that
+        # create zero edges, with nothing to stop it. pairs_judged hard-caps every judgement
+        # call, fail-closed exactly like max_edges below (0 = unlimited).
+        max_pairs = settings.auto_link_max_pairs_per_run
 
         watermark = _get_watermark(conn)
         nodes = _select_nodes_after(conn, watermark, settings.auto_link_max_nodes_per_run)
 
         created = 0
+        pairs_judged = 0
         last_complete: int | None = None
 
         for node in nodes:
             if created >= max_edges:
                 break  # batch budget spent; do not advance past this node
+            if max_pairs > 0 and pairs_judged >= max_pairs:
+                break  # judgement budget spent; do not advance past this node
 
             node_resolved = True
             text = f"{node['title'] or ''} {node['content']}".strip()
@@ -358,6 +368,9 @@ def run_auto_linker(engine) -> None:
 
                 for match in similar:
                     if created >= max_edges:
+                        node_resolved = False  # interrupted mid-node
+                        break
+                    if max_pairs > 0 and pairs_judged >= max_pairs:
                         node_resolved = False  # interrupted mid-node
                         break
                     if match["id"] == node["id"]:
@@ -391,6 +404,7 @@ def run_auto_linker(engine) -> None:
                         continue
 
                     llm_result = _llm_classify_link(settings, node, other)
+                    pairs_judged += 1
                     if llm_result is None:
                         # LLM UNAVAILABLE (raw None) — transient. Leave node unresolved so the
                         # watermark does not pass it. Not a poison: if the LLM is down, EVERY node
--- a/src/ormah/config.py
+++ b/src/ormah/config.py
@@ -131,6 +131,7 @@ class Settings(BaseSettings):
     auto_link_cross_space_penalty: float = 0.1  # subtracted from similarity for cross-space pairs
     auto_link_max_edges_per_run: int = 500
     auto_link_max_nodes_per_run: int = 500  # cursor batch: nodes scanned per run
+    auto_link_max_pairs_per_run: int = 200  # hard cap on LLM judgements per run (0 = unlimited)
 
     # Auto-merge
     auto_merge_threshold: float = 0.85
--- a/src/ormah/index/builder.py
+++ b/src/ormah/index/builder.py
@@ -6,6 +6,7 @@ import logging
 from pathlib import Path
 
 from ormah.index.db import Database
+from ormah.index.fingerprint import content_fingerprint
 from ormah.store.file_store import FileStore
 from ormah.store.markdown import parse_node
 
@@ -115,9 +116,10 @@ class IndexBuilder:
             INSERT OR REPLACE INTO nodes
             (id, type, tier, source, space, title, content, created, updated,
              last_accessed, access_count, confidence, importance,
-             valid_until, stability, last_review, file_path, file_hash)
+             valid_until, stability, last_review, file_path, file_hash,
+             content_fingerprint)
             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,
-                    ?, ?, ?, ?, ?)
+                    ?, ?, ?, ?, ?, ?)
             """,
             (
                 node.id,
@@ -138,6 +140,7 @@ class IndexBuilder:
                 node.last_review.isoformat() if node.last_review else None,
                 str(path),
                 file_hash,
+                content_fingerprint(node.title, node.content, node.type.value, node.space),
             ),
         )
 
--- a/src/ormah/index/builder.py
+++ b/src/ormah/index/builder.py
@@ -99,7 +99,16 @@ class IndexBuilder:
         node = parse_node(path.read_text(encoding="utf-8"))
         with self.db.transaction():
             prior = self._prior_row(node.id)  # read BEFORE the delete (#126)
-            self._remove_node(node.id)
+            # The vector is still valid exactly when the content fingerprint is: title and
+            # content are what feed the embedding. Dropping it on an unchanged-content
+            # reindex is permanent loss — nothing re-embeds it — and the node would sit
+            # behind the watermark with no vector, invisible to the linker and unable to be
+            # anyone else's semantic candidate. mark_outdated() (valid_until only) walks
+            # exactly this path.
+            unchanged = prior is not None and prior["content_fingerprint"] == content_fingerprint(
+                node.title, node.content, node.type.value, node.space
+            )
+            self._remove_node(node.id, keep_vectors=unchanged)
             self._index_file(path, prior)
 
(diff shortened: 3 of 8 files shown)


AndreLFSMartins:
## Peer review round — four fixes pushed

Two independent reviewers (Cursor, Codex) attacked this with the brief: *find a path where `seq` is preserved but a linking input changed, freezing a node out of relinking, or where invalidating cached pairs causes an LLM re-judgement storm.* Both barrels landed. Fixes:

### 1. An unchanged-content reindex destroyed the node's vector (HIGH — real)

`index_single()` called `_remove_node()` without `keep_vectors`, and **nothing re-embeds** — `_reindex_all_embeddings` only runs at startup/rebuild, there is no periodic backfill. Combined with the preserved `seq`, `mark_outdated()` (which only sets `valid_until`) left the node **behind the watermark AND with no vector** — invisible to the linker *and* unable to be any other node's semantic candidate. Before this PR the bumped `seq` at least made the auto_linker park the watermark on it, a visible fail-closed signal; the fix had silently removed that signal without fixing the cause.

The vector is valid exactly when the content fingerprint is, so it is now kept under the same condition. `test_unchanged_content_reindex_keeps_the_vector` fails without it.

### 2. `full_rebuild` carried stale verdicts across restored content (HIGH — real)

The rebuild clears the watermark so everything is requeued, but left `auto_link_checked` intact — and the auto_linker skips any cached pair *before* it looks at the edge. So a restore over **changed** content silently carried the old content's link decisions forward, and the fresh `seq` bought nothing.

It now drops the verdicts of nodes whose fingerprint actually changed, and keeps them for untouched nodes — rebuilding an unchanged store must not re-judge it with the LLM.

### 3. The run had no LLM-call budget (HIGH — real, and this PR created the exposure)

`auto_link_max_edges_per_run` only increments on a **positive** edge, so `none`/`error` judgements were free. Harmless before — cached pairs were skipped. But **this PR makes those pairs eligible again**, so a bulk `space` change could make one run issue ~500 nodes × 6 candidates ≈ 3,000 LLM calls while creating **zero** edges, with nothing to stop it.

New `auto_link_max_pairs_per_run` (default 200) hard-caps judgements per run, fail-closed like `max_edges`: the node interrupted mid-judgement stays unresolved so the watermark never advances past it. The test pins this with **all verdicts `none`** — `created` stays 0 the whole run, proving the cap is independent of edge creation.

### 4. Index on `auto_link_checked(node_b)`

`_invalidate_checked_pairs` deletes on `node_a = ? OR node_b = ?`; the PK only covers `node_a`.

---

### Rebutted, with reasoning

**"`incremental_update` keeps a stale vector when the content changed."** True, and pre-existing — but deleting it there would be *worse*: with no periodic re-embed, the node becomes vectorless and **parks the watermark forever** (the auto_linker's own comment calls this out). The real fix is re-embedding on reindex, which is a separate change.

---

Suite: **1427 passed**, same 6 pre-existing failures as `main`. Ruff clean.

PR #128, opened 2026-07-15
test: the suite reads the developer's real environment — make it hermetic (#106)

### LINKED ISSUE #106: Config tests read the developer's real ~/.config/ormah/.env (test isolation)
## What

`tests/test_config.py::test_llm_provider_defaults_to_none` fails on any machine that has a real ormah config: `Settings.model_config` resolves `env_file` to `~/.config/ormah/.env` at import time (`src/ormah/config.py:20`), so the developer's live `ORMAH_LLM_PROVIDER=litellm` overrides the tested default.

## How it was discovered

Running the full suite for the Sync v1 Step 0 PR (#105): 1201 passed, this one test failed — on the feature branch **and** on clean main. No shell env vars were set; the leak comes from the config file discovery in `_EXISTING_ENV_FILES`.

## Root cause

Config tests construct `Settings(...)` directly without isolating the env-file source. Same family as the recently fixed uninstall-test isolation (2e76b5b).

## Possible fixes

- Fixture that patches `Settings.model_config["env_file"]` (or `_EXISTING_ENV_FILES`) to a nonexistent path for config tests.
- Or a `Settings(_env_file=None)` override in the `_settings()` test helper — probably the one-line fix.

AndreLFSMartins:
Fixes #106.

On a machine that actually runs ormah, the test suite does not describe the tree under test — it describes the developer's install. Three separate leaks, one of which stops collection entirely.

## 1. `Settings()` is built at import time

`ormah/config.py` instantiates its singleton when the module loads, reading `~/.config/ormah/.env` and every `ORMAH_*` variable. `tests/conftest.py` imports it at line 12, so no fixture can get in front of it — a provider this branch does not accept aborts the run before a single test executes.

Two layers fix it: a root `conftest.py`, which loads before `tests/conftest.py` and is therefore the only place that can act before that import (it drops `ORMAH_*` and points `HOME` at an empty directory), and the autouse fixture in `tests/conftest.py` for `Settings()` instances built later during a test.

Pointing `HOME` away has a second benefit: `Settings.memory_dir` defaults under `Path.home()`, so the whole run stays off the developer's real memory store.

## 2. `shutil.which` is not where the binary is found

`TestConfigureCodexMcp` arranges "no codex binary" by patching `shutil.which`, but `configure_codex_mcp` calls `_find_binary` — and `shutil.which` is only its first lookup, followed by the mise shims, nvm bins, `~/.local/bin`, `/usr/local/bin` and `/opt/homebrew/bin`. Anyone with the Codex CLI installed fails these three tests on `main` today. `TestEnsureCodexHooksEnabled` in the same file already patches `_find_binary`; this makes the three consistent with it.

## 3. `$HOME` compared unresolved (one production line)

`detect_space_from_dir` realpaths its argument, then compares it against a raw `expanduser("~")`. When `$HOME` traverses a symlink the two never match, the home-directory guard does not fire, and the caller gets the home's basename as a space instead of `None`.

## Verification

From a worktree cut off `main`, on a machine with the Codex CLI installed and a populated `~/.config/ormah/.env`:

| | before | after |
|---|---|---|
| developer's real environment | collection aborts (exit 4) | **2008 passed**, exit 0 |
| `HOME=$(mktemp -d)` | 4 failed | **2008 passed**, exit 0 |

`ruff check src/ tests/ conftest.py` clean. Everything is test scaffolding except the single `realpath` line in `space_detect.py`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,24 @@
+"""Pytest bootstrap — runs before tests/conftest.py, and before anything
+imports ormah.
+
+`ormah.config` builds its `Settings()` singleton at module import time, and
+`Settings` reads `~/.config/ormah/.env` plus every `ORMAH_*` variable in the
+environment. On a developer machine that runs ormah, those describe the live
+install, not the test tree: a provider this branch does not accept aborts
+collection outright, and a non-default embedding model silently changes what
+the cache tests look for.
+
+Point HOME at an empty directory and drop the ORMAH_* variables here, before
+that import happens. This also keeps `Settings.memory_dir` — which defaults
+under `Path.home()` — away from the developer's real store.
+"""
+
+from __future__ import annotations
+
+import os
+import tempfile
+
+for _key in [k for k in os.environ if k.startswith("ORMAH_")]:
+    del os.environ[_key]
+
+os.environ["HOME"] = tempfile.mkdtemp(prefix="ormah-test-home-")
--- a/src/ormah/adapters/space_detect.py
+++ b/src/ormah/adapters/space_detect.py
@@ -14,8 +14,11 @@ def detect_space_from_dir(path: str) -> str | None:
     """
     path = os.path.realpath(path)
 
-    # Don't auto-detect space for home directory or root
-    home = os.path.expanduser("~")
+    # Don't auto-detect space for home directory or root. realpath both sides:
+    # `path` is already resolved above, so comparing it against an unresolved
+    # $HOME misses the home directory whenever $HOME traverses a symlink, and
+    # the caller silently gets the home's basename as a space instead of None.
+    home = os.path.realpath(os.path.expanduser("~"))
     if path == home or path == "/":
         return None
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import os
 import shutil
-import tempfile
 from pathlib import Path
 
 import pytest
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import os
 import shutil
+import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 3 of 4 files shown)


AndreLFSMartins:
Rebased onto current `main` and widened, because the original fixture turned out to close only part of the leak.

**The part it could not reach.** `ormah/config.py` builds its `Settings()` singleton at module import time, and `tests/conftest.py` imports it at line 12 — before any fixture runs. On a machine that actually runs ormah, collection aborts outright:

```
ImportError while loading conftest 'tests/conftest.py'
  ...
  src/ormah/config.py:663: settings = Settings()
E pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
E llm_provider
E   Value error, llm_provider must be one of {'none', 'ollama', 'litellm'}, got 'claude_cli'
```

A root `conftest.py` loads before `tests/conftest.py`, which makes it the only place that can act before that import. It drops the `ORMAH_*` variables and points `HOME` at an empty directory, so `~/.config/ormah/.env` is not there to be read. That also keeps `Settings.memory_dir` — which defaults under `Path.home()` — away from the developer's real memory store for the whole run.

**A second leak, unrelated to the env file.** The three `TestConfigureCodexMcp` cases arrange "no codex binary" by patching `shutil.which`, but `configure_codex_mcp` calls `_find_binary`, where `shutil.which` is only the first lookup — it then walks the mise shims, nvm bins, `~/.local/bin`, `/usr/local/bin` and `/opt/homebrew/bin`. **Anyone with the Codex CLI installed fails these three tests on `main` today**, with `Expected 'run' to not have been called. Called 1 times.` Patching `_find_binary` is what `TestEnsureCodexHooksEnabled` in the same file already does.

**One production line.** `detect_space_from_dir` realpaths its argument and compares it against a raw `expanduser("~")`. When `$HOME` traverses a symlink the two never match, the home guard does not fire, and the caller gets the home's basename as a space instead of `None`. `test_detect_returns_none_for_home` fails the moment `$HOME` is symlinked (on macOS, `/var/...` → `/private/var/...` suffices).

**Verification.** On a machine with the Codex CLI installed and a populated `~/.config/ormah/.env`, from a worktree cut off `main`:

| | before | after |
|---|---|---|
| suite, developer's real environment | collection aborts (exit 4) | **2008 passed**, exit 0 |
| suite, `HOME=$(mktemp -d)` | 4 failed | **2008 passed**, exit 0 |

`ruff check src/ tests/ conftest.py` clean. Every change is in test scaffolding except the one `realpath` line.

PR #130, opened 2026-07-15
fix(db): retire per-thread SQLite connections on thread death — FD leak (#83)

### LINKED ISSUE #83: bug(db): per-thread SQLite connections never released → fd leak → [Errno 24] Too many open files under long-running server
## Symptom

A long-running server accumulates open file descriptors on `index.db` (each WAL connection holds ~3: `.db` / `-wal` / `-shm`) until `[Errno 24] Too many open files`, after which the server stops responding (connection reset). In one observed incident a single server process held **2641 fds** on `index.db` before wedging.

## Root cause

`Database._new_connection` (`src/ormah/index/db.py`) registers every per-thread connection in `self._all_conns` and **never closes a connection when its owning thread dies** — the only close path is `Database.close()` at server shutdown:

```python
# db.py
with self._conns_lock:
    self._all_conns.append(conn)   # appended, never popped/closed per-thread
return conn
```

Any thread that touches `db.conn` (via `threading.local`) creates a connection that outlives the thread. The dominant driver is agent-backed maintenance: each phase spawns a **fresh daemon thread** per cycle (`src/ormah/background/maintenance_manager.py:61,88`), leaking ~2 connections per maintenance cycle. Pool-based threads (APScheduler `ThreadPoolExecutor`, the anyio limiter) reuse their connection, so only these ephemeral threads leak.

Introduced by #19 (thread-local connections, which replaced the shared-connection global-lock freeze from #18). #48 later made the thread-local connection test deterministic but did not surface the fd leak.

## Reproduction

Spawn N ephemeral threads that each touch `db.conn`, join them, force a GC:

```python
db = Database(tmp / "t.db"); db.init_schema()
threads = [threading.Thread(target=lambda: db.conn.execute("SELECT 1").fetchone()) for _ in range(20)]
for t in threads: t.start()
for t in threads: t.join()
del threads; gc.collect()
assert len(db._all_conns) <= 1   # fails today: stays at 21, fds never released
```

## Proposed fix

Register a `weakref.finalize` on the owning thread object so the connection is closed (and dropped from `_all_conns`) when the thread is garbage-collected — root-cause and driver-agnostic, no need to hunt every call site:

```python
weakref.finalize(threading.current_thread(), self._retire_connection, conn)
```

`Database.close()` still handles any still-live threads' connections at shutdown.

## Notes on evidence

- The 2641-fd count and the server wedge were **observed** in a live incident.
- The ~2 connections/cycle leak rate is **inferred** from the connection lifecycle, not directly measured.
- Version: reproduced against the `0.13.2` (#72) base; the culprit lines are unchanged through `origin/main` @ `dc49714`.

AndreLFSMartins:
## Problem

Per-thread SQLite connections were created but **never released**. Under a long-running server, worker threads come and go while their connections stay open, leaking file descriptors until `[Errno 24] Too many open files`.

## Fix

Register `weakref.finalize(threading.current_thread(), self._retire_connection, conn)` on every new per-thread connection. When the owning thread is garbage-collected, `_retire_connection` closes the connection and drops it from `_all_conns`. Bounded FD usage regardless of thread churn.

## Verification

`pytest tests/test_index/test_db_fd_lifecycle.py` → 1 passed (`test_ephemeral_thread_connection_is_retired`).

Fixes #83

### DIFF
--- a/src/ormah/index/db.py
+++ b/src/ormah/index/db.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import sqlite3
 import threading
+import weakref
 from contextlib import contextmanager
 from pathlib import Path
 
@@ -48,6 +49,11 @@ class Database:
             pass  # sqlite-vec not installed — vector search disabled
         with self._conns_lock:
             self._all_conns.append(conn)
+        # Auto-close this connection when its OWNING thread is garbage-collected. Without this,
+        # an ephemeral thread (e.g. a maintenance phase thread) leaks its connection — and its
+        # ~3 WAL fds — for the whole process lifetime, until [Errno 24]. finalize fires at most
+        # once; close() (shutdown) also handles any still-live threads' connections.
+        weakref.finalize(threading.current_thread(), self._retire_connection, conn)
         return conn
 
     @property
@@ -542,6 +548,18 @@ class Database:
                 )
         except ImportError:
             pass  # sqlite-vec not available, vector search disabled
+
+    def _retire_connection(self, conn: sqlite3.Connection) -> None:
+        """Drop a dead thread's connection from the registry and close it."""
+        with self._conns_lock:
+            try:
+                self._all_conns.remove(conn)
+            except ValueError:
+                pass  # already retired by close()
+        try:
+            conn.close()
+        except Exception:  # noqa: BLE001
+            pass
 
     def close(self) -> None:
         with self._conns_lock:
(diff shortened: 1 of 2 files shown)


AndreLFSMartins:
Closing as a duplicate of #275, which landed today.

Same issue (#83), same mechanism — and #275 is the better version. It carries everything that was here (`weakref.finalize(threading.current_thread(), self._retire_connection, conn)` and `_retire_connection`) plus finalizer tracking, a re-entrant lock for the case where the GC fires a connection finalizer inside a section that already holds it, and `finalizer.detach()` on close — none of which this PR had.

My fault for cutting the branch twice; I only spotted it while inventorying what my Beta diverges by. Nothing is lost here.

Issue #81, opened 2026-07-04
perf(background): duplicate_merger + conflict_detector re-scan the whole store (ORDER BY RANDOM) every run — no incremental watermark like #26 gave auto_linker


AndreLFSMartins:
# perf(background): duplicate_merger + conflict_detector re-scan the whole store (`ORDER BY RANDOM()`) every run — no incremental watermark like #26 gave auto_linker

## Summary

`#26` fixed `auto_linker` by adding an incremental **seq watermark**: it now processes only nodes newer than the last run, so its cost is O(Δ) instead of O(n²). Its two O(n²) siblings — `run_duplicate_detection` and `run_conflict_detection` — were **not** given the same treatment. Their candidate finders still `SELECT ... FROM nodes ORDER BY RANDOM()` over the **whole table** every run and rely solely on the persisted skip tables (`duplicate_checked` / `conflict_checked`, added in PR #79) to filter out already-resolved pairs.

The LLM-call *cost* per run is bounded (`duplicate_check_max_llm_calls_per_run` / `conflict_check_max_llm_calls_per_run`, default 100). The problem is **selection**, not budget:

- **No prioritization of new work.** A genuinely-new duplicate/conflict isn't checked first — it waits until `RANDOM()` happens to surface its node, which gets *less* likely as `n` grows.
- **Degrading efficiency as the skip tables fill.** Once most candidate pairs are resolved, each run must scan progressively further into the randomly-ordered table to collect a batch of still-unresolved candidates — wasting effort re-probing resolved neighborhoods.

Net effect (the concern that prompted this issue): on a growing store, a fixed per-run cap under **random** selection means maintenance never reliably converges on the *new* work — it becomes a smaller and less relevant fraction of the graph over time, even though the persistence half of the fix is already in place.

> Identified by code inspection during the first real `claude_cli` sleep-cycle (~5900 nodes). Not yet measured against wall-clock on a large store.

## Evidence

**Dedup** — `src/ormah/background/duplicate_merger.py` (`_find_merge_candidates`, ~L175):

```python
nodes = engine.db.conn.execute(
    "SELECT id, content, title, type FROM nodes ORDER BY RANDOM()"
).fetchall()                                   # O(n) full-table, random order, no watermark
for node in nodes:
    query_vec = encoder.encode(text)
    similar = vec_store.search(query_vec, limit=6)   # O(n) brute-force per node
    # ... skip via duplicate_checked WHERE result='not_duplicate' ...
```

[...]

Issue #122, opened 2026-07-14
bug(background): a deterministically failing pair parks the auto_linker watermark forever — the monotonic cursor has no retry/dead-letter set


AndreLFSMartins:
Deferred follow-up from #117 (see the closing comment there).

The auto_linker advances a monotonic watermark (`seq > watermark`, `_select_nodes_after`). A node is only allowed to advance the cursor once it is fully collected, drained and `resolved`. #119 made a failing pair *fail closed* — the node stays unresolved and the watermark does not move past it — which is correct for a transient failure.

But there is no escape hatch for a **deterministic** failure. If one pair fails on every single run (poison content, a permanently unparseable LLM answer, a node whose markdown cannot be written), the watermark parks at that node forever and **every later node is starved** — the exact freeze #117 was about, just reached by a different route.

What's missing: a per-pair attempt counter and a dead-letter set, so a pair that has failed N times is recorded, skipped, and surfaced, letting the cursor move on instead of blocking the whole store.

Related: #109 (same freeze symptom, different cause).

PR #133, opened 2026-07-15
perf(background): delta-select dedup + conflict seeds by seq watermark (#81)

AndreLFSMartins:
## What & why

Part of #81. `duplicate_merger` and `conflict_detector` selected candidates with
`ORDER BY RANDOM()` over the whole store every run, so on a growing store a genuinely-new
duplicate/conflict waited until chance surfaced its node — coverage never converged
(measured: 9,453 / 10,287 nodes never appeared in any checked pair). This applies the same
incremental **seq-watermark** pattern #26 gave `auto_linker` to its two O(n²) siblings, so
each background run processes only nodes newer than the last (O(Δ) instead of O(n²)).

## Design

- **Shared `background/watermark.py`** — key-parametrized `get`/`set` on `meta`
  (`duplicate_check_watermark`, `conflict_check_watermark`). `auto_linker` is left on its own
  private copy to keep this diff small; migrating it is a follow-up.
- **Opt-in `delta` flag on both finders.** `delta=False` (default) is today's `ORDER BY
  RANDOM()` selection **byte-for-byte** and a plain-list return — the agent / two-call
  maintenance path (`memory_engine.py`) is unchanged and never advances a cursor. Only the
  background `run_*` jobs pass `delta=True`. This is load-bearing: with the default
  `llm_provider="none"`, the `run_*` jobs never fire, so an agent-only deployment must keep
  the random path to make progress.
- **Vector neighbours stay age-unfiltered** — a new seed pairs against neighbours of any age,
  so every new duplicate/conflict (which always involves ≥1 new node) is reachable.
- **Watermark advances only past the contiguous drained prefix.** A seed with no persisted
  vector is a fail-closed **barrier** (the vector index can be empty/backfilling right after
  `full_rebuild`): later seeds are still processed (liveness, mirroring `auto_linker`) but no
  seed drains past the barrier, so the cursor can't jump the hole. An LLM-unavailable pair
  likewise stops the advance at its seed.
- **`run_duplicate_detection` re-fetches full rows** before the LLM/merge — the finder's
  candidate previews truncate content to 400 chars, and merging from a preview would corrupt
  node content.
- **`full_rebuild` clears all three cursors** (+ the conflict scope stamp), since mass reindex
  re-allocates `seq`.
- Config: `duplicate_check_max_nodes_per_run` / `conflict_check_max_nodes_per_run` (default 500),
  mirroring `auto_link_max_nodes_per_run`.

## Peer review

This went through the repo's peer-review pipeline, which caught three real correctness bugs,
each fixed with a regression test first:

1. **Vectorless watermark hole** — the advance considered only drained seeds, so a drained
   higher-seq seed advanced the cursor *past* a vectorless-skipped lower-seq seed, permanently
   stranding its pairs. Fixed by making the vectorless seed a drain barrier.
2. **Ephemeral scope reset** — the `conflict_check_all_spaces` reset lived only as a local var
   in the finder while the run wrote the scope stamp unconditionally; an empty/barrier batch
   therefore lost the reset. Fixed by persisting the reset in the run before selection.
3. **Background dedup inheriting `auto_link_checked`** — routing the run through the finder
   made background dedup skip every pair `auto_linker` had already examined (that table is
   auto_linker's alone). Fixed by having the background run ignore `auto_link_checked`
   (`respect_checked=False`); the agent path keeps it.

A once-per-run warning now logs when a vectorless barrier parks a cursor.

## Known limitations (follow-ups, out of scope here)

- **Embedding-corpus reindex does not reset the cursors** — `_reindex_all_embeddings` (an
  embedding schema/model upgrade) replaces vectors without bumping `seq` or clearing a
  watermark, so already-drained nodes aren't re-examined with the new embeddings. This is a
  pre-existing property of the #26 pattern (it affects `auto_link_watermark` identically) and a
  correct fix must distinguish a schema change from a missing-vector backfill (to avoid
  amplifying #32) — better as its own cross-cutting issue. A content *edit* is unaffected
  (`update_node` re-embeds and bumps `seq`).
- **A permanently-unembeddable node parks its cursor** and replays its batch's LLM work each
  run — the same fail-closed cost `auto_linker` carries; bounded-retry / dead-letter is #122.
- **Pending-proposal suppression** — a duplicate pair suppressed because one node sits in an
  unrelated pending merge proposal isn't re-judged after the seed passes (the OR-match is
  intentional — it serializes overlapping merges). Recovery on proposal rejection is a
  follow-up.

## Tests

New coverage for the watermark helper, delta vs legacy selection, new×old pairing, the
vectorless barrier (finder + run level, incl. liveness), scope-toggle reset (finder + run),
LLM-failure parking, the full-content re-fetch guard, and the `auto_link_checked` regression.
Full `tests/test_background/` passes (excluding two pre-existing environment-leak failures
unrelated to this change).

### DIFF
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -182,12 +182,13 @@ def _find_conflict_candidates(
                 drained_seeds.append((node["id"], node["seq"]))
                 continue
 
-            # FAIL-CLOSED (overview invariant, mirrors upstream auto_linker.py):
-            # a seed with text but no persisted vector must NOT drain — an
-            # empty/backfilling index would return zero neighbors and the
-            # cursor would pass pairs that were never derived.
+            # FAIL-CLOSED BARRIER (overview invariant, mirrors upstream
+            # auto_linker.py): a seed with text but no persisted vector stops
+            # the batch here. `break`, not `continue` — a later seed must not
+            # drain and advance the watermark past this hole, or this seed's
+            # pairs are permanently skipped once its vector is backfilled.
             if delta and vec_store.get(node["id"]) is None:
-                continue  # NOT appended to drained_seeds
+                break
 
             query_vec = stored_or_encoded(
                 vec_store,
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -191,6 +191,12 @@ def _find_conflict_candidates(
             # not `break`: later seeds are still PROCESSED (liveness, mirrors
             # auto_linker) but no further seed drains once the barrier is hit.
             if delta and vec_store.get(node["id"]) is None:
+                if not barrier_hit:
+                    logger.warning(
+                        "conflict delta stalled: node %s has no persisted vector (embedding "
+                        "backfill pending?); cursor parked at seq %s until it embeds",
+                        node["id"][:8], node["seq"],
+                    )
                 barrier_hit = True
                 continue
 
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -281,6 +281,18 @@ def run_conflict_detection(engine) -> None:
             CONFLICT_WATERMARK_KEY, get_watermark, set_watermark,
         )
 
+        # Durably reset the cursor on a scope change BEFORE selection — so an
+        # expanded scope re-examines older nodes even if THIS run cannot
+        # drain the first seed (vectorless barrier) or the finder returns
+        # empty on a transient error. Persisting here (not just ephemerally
+        # in the finder) is what makes the reset survive to the next run:
+        # the advance loop below reloads the persisted watermark.
+        _stamp = engine.db.conn.execute(
+            "SELECT value FROM meta WHERE key = ?", (CONFLICT_SCOPE_STAMP_KEY,)
+        ).fetchone()
+        if _stamp is not None and _stamp["value"] != _conflict_scope_value(settings):
+            set_watermark(engine, CONFLICT_WATERMARK_KEY, 0)
+
         candidates, drained_seeds = _find_conflict_candidates(
             engine, limit=10_000, delta=True,
         )
--- a/src/ormah/background/conflict_detector.py
+++ b/src/ormah/background/conflict_detector.py
@@ -276,7 +276,14 @@ def run_conflict_detection(engine) -> None:
             logger.debug("Conflict detection skipped: LLM not enabled")
             return
 
-        candidates = _find_conflict_candidates(engine, limit=10000)
+        from ormah.background.watermark import (
+            CONFLICT_WATERMARK_KEY, get_watermark, set_watermark,
+        )
+
+        candidates, drained_seeds = _find_conflict_candidates(
+            engine, limit=10_000, delta=True,
+        )
+        failed_seed_seqs: set[int] = set()
         edges_created = 0
         dirty_nodes: dict[str, list[Connection]] = {}
 
@@ -286,6 +293,7 @@ def run_conflict_detection(engine) -> None:
 
             llm_result = _llm_check_conflict(settings, node_a, node_b)
             if llm_result is None:
+                failed_seed_seqs.add(candidate["seed_seq"])
                 continue
             if not llm_result.get("conflict"):
                 continue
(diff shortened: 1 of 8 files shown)


AndreLFSMartins:
Flagging a sequencing risk I found while re-verifying #81 today — not a defect in this PR, but a reason not to land it alone. Detail in https://github.com/r-spade/ormah/issues/81#issuecomment-5159377608.

**#81 has two halves and this PR is one of them.** I wrote that issue assuming the persistence half was already done via `duplicate_checked` / `conflict_checked`. It isn't: those tables were introduced on #79, which is still open, and they have zero occurrences on `main`. What `main` actually has is one shared table — all three jobs **read** `auto_link_checked`, and only `auto_linker` **writes** to it (`auto_linker.py:287`; conflict reads at `:166`, dedup at `:191`, neither writes).

**Why that interacts with delta-selection specifically.** This PR is careful about the obvious objection — "vector neighbours stay age-unfiltered, so every new duplicate/conflict is reachable" — and that's correct as far as *selection* goes. But it narrows the conflict/dedup seed set to Δ, which is the same set `auto_linker` is working through, and the linker is the only one of the three that records a verdict. So a pair's fate depends on who reaches it first, and both outcomes are bad:

- linker first → the pair is in `auto_link_checked`, conflict/dedup skip it, **and it never comes back**, because with a watermark there's no second pass over old seeds to catch it;
- conflict first → it judges the pair, writes nothing, and re-judges it on a later run at full LLM cost.

Today's `ORDER BY RANDOM()` is what accidentally limits the damage: it keeps re-drawing old seeds, so pairs the linker hasn't registered still get picked up eventually. Replacing that with a watermark removes the safety net before the thing it compensates for is fixed.

**I'm not claiming this as measured** — I haven't run a cycle with this branch applied. The defaults make all three jobs 1440 min and they share the scheduler's thread pool, so I can't tell you who wins the race, only that the loser's work is discarded either way. The check that would settle it: apply this branch to a populated store, run one maintenance cycle, and count pairs actually judged by `conflict_detector` versus the pre-PR baseline. If that number collapses, the ordering matters; if it doesn't, my concern is wrong and this is clear.

**What I'd suggest**, none of which requires changing the code here:

1. Retitle the closing keyword to **"part of #81"** rather than "Closes #81", so the issue stays open for the persistence half.
2. Land a per-job memo (own table, or a `job` column on the shared one) before or together with this — recording a terminal result for *every* judged pair, including "no conflict", with errors explicitly non-terminal.
3. If you'd rather have it as one unit, I can add the memo to this branch; I kept them separate because the memo is arguably its own fix and this PR is already sizeable.

Happy to do whichever you prefer, or to drop this if you read the interaction differently — the code here is sound and I'd rather it landed in the right order than not at all.

r-spade:
Thanks Andre — I reviewed and tested #133. The watermark direction looks right, and the relevant background tests pass.

I reproduced the sequencing risk you identified: because the three jobs currently rely on `auto_link_checked`, auto-linking can cause a conflict or duplicate pair to be permanently skipped after the watermark advances.

I also found a related case: a pending merge proposal can suppress another duplicate pair, and if that proposal is later rejected or removed, the watermark may already have passed the pair permanently.

I think the safest order is:

1. Add a small, focused prerequisite that gives each job its own persistent pair verdicts, including terminal negative results while keeping transient errors retryable.
2. Make proposal rejection/removal return affected pairs to eligibility.
3. Rebase #133 onto that foundation, validate and document the new per-run limits, and describe it as `Part of #81` unless the agent maintenance path is bounded too.

Does that sequencing and division of responsibility sound right to you?

Since you are already taking on #220–#223, I’m happy to own the prerequisite and remaining integration work here rather than add more to your queue. I can prepare the focused PR, then we can coordinate the smallest update needed for #133.

This is a request to tighten the sequencing and correctness, not a rejection of the approach in #133.

AndreLFSMartins:
Yes to all three — the sequencing is right, and I'd rather you own the prerequisite than have it sit behind #220–#223. Thanks for actually running it.

**On the proposal-suppression case you found.** That's the stronger of the two, and I under-rated it. I had it in the PR description as a follow-up ("Pending-proposal suppression"), reasoning that the OR-match is intentional because it serializes overlapping merges. What I missed is that today it self-heals *by accident*: `ORDER BY RANDOM()` redraws the seed on some later run, after the proposal is resolved. The watermark removes the redraw, so a rejected or removed proposal permanently strands its pair. Same shape as the `auto_link_checked` interaction — the watermark isn't introducing the hole, it's removing the thing that was papering over it. Promoting it to a prerequisite is correct.

**One boundary worth pinning down in the prerequisite's design.** The per-job memo has to treat "no conflict" / "not a duplicate" as terminal while an LLM or embedding failure stays retryable — otherwise it either drops the negatives (and we re-pay the full LLM cost for them every run, which is most of what these jobs do) or it records transient errors as verdicts (and the pair is never re-judged). That's the same split #122 wants for the auto_linker cursor's dead-letter set, so it's worth deciding once: a *deterministically failing* pair and a *pair judged negative* have to land in different buckets, even though both look like "we're done with this one" from the run's perspective.

**Acceptance check I'd suggest for your PR.** The measurement I flagged as unrun still applies, and it's the one that settles whether the memo actually closed the hole: apply the prerequisite plus this branch to a populated store, run one maintenance cycle, and count pairs actually judged by `conflict_detector` against the pre-PR baseline. Pre-memo that number should collapse; post-memo it shouldn't. Happy to run it on my store if that's useful — I have a populated one and it's the same setup that produced the 9,453 / 10,287 figure in the description.

**Housekeeping I'll do on #133 now, no need to wait on you:** retitle the closing keyword to `Part of #81`. Your caveat holds either way — the agent maintenance path is still `delta=False` / `ORDER BY RANDOM()` byte-for-byte, so #81 isn't fully closed by this PR regardless of the memo.

**Two scheduling questions:**

1. **#95** (batch K pairs per LLM call, #87) is parked on your earlier "don't rebase until #133 lands". #133 now waits on your prerequisite, so that's two hops. Does #95 need to stay behind #133, or can it rebase onto `main` independently? It touches the same `run_*` functions so I'd expect a textual conflict, but batching is orthogonal to selection — I don't see a semantic dependency. Your call; I'd rather it wait than create merge pain for you.

2. Does the prerequisite's second half (proposal rejection returning pairs to eligibility) overlap with **#209** (unbounded merge-proposal queue)? Both touch the proposal lifecycle, and if the same code path is getting a TTL/cap and a re-eligibility hook, it may be cheaper as one change than two.

PR #147, opened 2026-07-16
fix(setup): skip the Claude Code hooks and MCP the ormah plugin already provides

### LINKED ISSUE #145: bug(setup): `ormah setup` re-wires hooks the Claude Code plugin already registers — every prompt whispers twice
## Summary

When the Claude Code plugin is installed, `ormah setup` wires the same three hooks a **second** time into `~/.claude/settings.json`. Both wirings fire, so every human turn runs the whisper twice.

The plugin registers `UserPromptSubmit`, `PreCompact` and `SessionEnd` in `integrations/claude-plugin/hooks/hooks.json`. `_claude_code_wire()` (`src/ormah/setup.py:2388`) calls `configure_claude_hooks()` (`src/ormah/setup.py:337`), which writes the same three events into `~/.claude/settings.json` with absolute paths. `_install_hooks` merges carefully **within** settings.json, but it cannot see the plugin's — those live in a different file, so no merge can dedupe across the two.

## Evidence

On my install (plugin 0.13.3 + settings.json wiring), one prompt produced **2** rows in `retrieval_events`. After removing the settings.json entry: **1**.

```
1 prompt, both wirings present -> 2 retrieval_events rows
1 prompt, plugin wiring only   -> 1 row
```

Both commands resolve to the same binary — the plugin shim is `exec ormah whisper inject` and `which ormah` is the same path `configure_claude_hooks` hardcodes — so this is pure duplicate work, not two different behaviours:

- 2× encode + hybrid search + cross-encoder rerank per human turn (the expensive path, on every prompt).
- 2 rows per prompt in `retrieval_events`, which biases any per-event metric — `whisper_health` counts each prompt twice.
- doubled nudge/extraction counters.
- `PreCompact`/`SessionEnd`: `ormah whisper store` runs twice per compaction and per session end.

## Why the existing guard doesn't hold

`--skip-client-setup` exists (`src/ormah/cli.py:790`, `run_setup(skip_client_setup=...)` at `src/ormah/setup.py:2683`) and `integrations/claude-plugin/SETUP.md:38` is explicit: *"Do not run `ormah setup` without `--skip-client-setup` in this plugin flow."* So the project already knows. But the guard is **opt-in**, and the paths a plugin user is most likely to hit don't pass it:

- `install.sh:196` — plain `ormah setup`.
- `install.sh:191` — `ormah setup --update` on the re-run/upgrade path, under the banner *"Reapplying integrations — updating hooks"*. `--update` does not imply `--skip-client-setup` (only `skip_client_setup` gates the wire loop at `setup.py:2683`), and `SETUP.md:35` warns against conflating them — but that warning only reaches a human reading the doc, not the installer.
- `desktop/ui/src/InstallPanel.tsx:38` → `setup_agents` → `ormah setup --json` (`desktop/src-tauri/src/commands.rs:69`), no flag.

Detection is the root of it: `_claude_code_detected()` (`setup.py:2293`) is just "is the `claude` binary present", and `_claude_code_is_wired()` (`setup.py:2297`) only inspects `settings.json`. With the plugin installed and working, `is_wired` still reports False — so setup dutifully wires again. Nothing in the chain can observe that the plugin already owns these hooks.

This is also why manual cleanup doesn't stick: removing the entries is a one-off human action, re-adding them is a side effect of a command run for an unrelated reason (install, upgrade, reconfigure the LLM). I removed these same three entries on 2026-07-15 and found them back on 2026-07-16.

## Proposed fix

Make the guard automatic instead of opt-in: have the Claude Code wire path detect an installed ormah plugin (its `hooks/hooks.json` under `~/.claude/plugins`, or the `ormah@ormah` entry in `enabledPlugins`) and skip the hook wiring, reporting *"plugin already provides the hooks"* rather than writing. That fixes `install.sh`, the `--update` path and the desktop button in one place, since all three route through `_claude_code_wire`.

Two details worth deciding, hence no PR yet:

1. **Scope of the skip.** The plugin also ships MCP + skills, so the redundancy may cover all of `_claude_code_wire` (`configure_claude_hooks`, `configure_claude_code_mcp`, `install_claude_md`, ...), not only the hooks. Skipping just the hooks is the minimal, safest cut.
2. **`_claude_code_is_wired` should agree.** If it stays blind to the plugin it will keep reporting "not wired" on a perfectly wired install, which is confusing in `ormah status` output regardless of what setup does.

Happy to send a PR if you agree on the shape.

## Environment

ormah 0.13.6, plugin 0.13.3, macOS, Claude Code 2.1.x.

AndreLFSMartins:
Closes #145.

With the Claude Code plugin installed, `ormah setup` wires the same three hooks a **second** time into `~/.claude/settings.json`. Both wirings fire, so every human turn runs the whisper twice — two encodes, two hybrid searches, two cross-encoder reranks. The MCP server is registered twice for the same reason.

## Evidence

Live `retrieval_events` on my install (plugin 0.13.3 + settings.json wiring), last 7 days, grouped by `(session_id, prompt_hash)`:

| rows per prompt | prompts |
| --- | --- |
| 1 | 198 |
| **2** | **319** |
| 3–5 | 7 |

Both wirings resolve to the same binary — the plugin shim is `exec ormah whisper inject` and `which ormah` is the path `configure_claude_hooks` hardcodes. Pure duplicate work. Beyond cost, every per-event metric in `whisper_health` is double-counted.

`_install_hooks` merges carefully *within* `settings.json`, but the plugin's hooks live in a **different file**, so no merge can ever dedupe across the two.

## Why the existing guard didn't hold — and a finding that widens the issue

`--skip-client-setup` exists and `integrations/claude-plugin/SETUP.md:38` is explicit about it, but it is **opt-in** and the paths a plugin user actually hits don't pass it: `install.sh` (plain and `--update`) and the desktop button via `run_setup_json`.

**New finding, not in the issue:** `configure_claude_hooks` and `configure_claude_code_mcp` use `os.path.expanduser("~/...")`, which reads `$HOME` — not `Path.home()`. Two tests in `tests/test_setup_json.py` reach `_claude_code_wire()` without isolating HOME, so **running the test suite re-wires the developer's real config**. That is very likely the real reason manual cleanup never sticks — more frequent than `ormah setup` itself. Fixed here (`a478221`), structurally, by pinning `HOME`.

## The fix

A guard at the top of `_claude_code_wire()` — the single funnel all three entry points pass through. When a user-scoped plugin is **verified working**, setup strips the redundant CLI hooks + MCP and skips re-writing them, so existing installs self-heal.

`_claude_code_plugin_provides_hooks()` licenses deleting the user's working config, so it is deliberately strict. It requires **all** of:

1. the plugin **enabled** in `~/.claude/settings.json`;
2. a **user-scoped** entry in `~/.claude/plugins/installed_plugins.json` under the **same fully-qualified key** that is enabled;
3. that entry's `installPath/hooks/hooks.json` **parsed**, declaring at least one hook `_is_ormah_hook()` accepts;
4. `installPath/.mcp.json` **parsed**, declaring `mcpServers.ormah`.

Any failure → `False` → setup wires exactly as today. A wrong `False` costs a duplicate (loud, measurable); a wrong `True` costs the user the whisper (silent). The default takes the loud side.

Each of those four is a separate peer-review correction — every earlier revision mistook one signal for another: enabled ≠ installed, registry entry ≠ manifests on disk, manifest exists ≠ manifest valid, some `ormah@*` key ≠ *the* enabled key.

**Scope of the strip:** hooks + MCP only. The agent and slash command are namespaced by the plugin (`/ormah:maintenance` vs `/ormah-maintenance`) — different names, not duplicate registrations, and they never auto-execute, so they contribute nothing to the double-whisper. They keep being installed, as does `CLAUDE.md`, which no plugin can write.

## Also fixed here (both needed by the above)

- `_claude_code_is_wired` never detected hooks: it read `entry.get("command")` off the **matcher** dict, whose only key is `"hooks"`, so the branch always saw `""` and the function could only return True via the `.claude.json` MCP fallback. It now inspects `matcher["hooks"][*]` via `_is_ormah_hook`, and treats a working plugin as wired — without that, stripping the MCP entry would make the desktop agent panel report a working install as "not wired".
- `_remove_mcp_from_json` wrote `~/.claude.json` with a bare `write_text` while its sibling used `_atomic_write`. This change makes that path run on every `setup --update` for plugin users, against the file holding the whole Claude Code config — a crash mid-write would truncate it.

## Known limitation (deliberate, tested)

A **project**- or **local**-scoped plugin does not trigger the guard, so the duplication persists there. `configure_claude_hooks` writes the **global** settings, which serve every project; removing them because one project enabled the plugin would break the whisper everywhere else. Global-active ↔ global-wired is the only safe symmetry. A test pins this so the behaviour is deliberate rather than accidental.

## Follow-up (not in this PR)

`_codex_is_wired` has the identical dead-branch bug this PR fixes for Claude Code. Left alone — Codex's hooks schema wasn't verified to match.

## Verification

- `pytest tests/` (HOME isolated): **1429 passed**. Same failures as the base commit — `TestConfigureCodexMcp` (the real `codex` binary breaks its mock) — plus two caused by the isolation harness itself. **Zero regressions**; +30 tests.
- `ruff check src/ tests/`: clean.
- Reviewed by a two-peer council (Cursor + Codex) at the plan stage and again on this diff; both rounds rejected an earlier revision, and every accepted finding is folded in.

### DIFF
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -2346,6 +2346,10 @@ def _claude_code_detected() -> bool:
 
 
 def _claude_code_is_wired() -> bool:
+    # The plugin ships the hooks and the MCP server — an install with a working
+    # plugin is wired even when settings.json holds nothing of ours.
+    if _claude_code_plugin_provides_hooks():
+        return True
     # Check for ormah whisper hooks in settings.json and ormah MCP in .claude.json
     settings_path = Path.home() / ".claude" / "settings.json"
     try:
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -684,25 +684,29 @@ def _install_markdown_block(
     target.write_text(updated)
 
 
-def _plugin_enabled_in_settings(settings_path: Path, plugin_name: str) -> bool:
-    """Return True when the plugin is enabled in a Claude settings file."""
+def _enabled_plugin_keys(settings_path: Path, plugin_name: str) -> list[str]:
+    """Return the fully-qualified enabledPlugins keys for a plugin, in file order."""
     if not settings_path.exists():
-        return False
+        return []
     try:
         data = json.loads(settings_path.read_text())
     except (json.JSONDecodeError, ValueError):
-        return False
+        return []
 
     enabled_plugins = data.get("enabledPlugins", {})
     if not isinstance(enabled_plugins, dict):
-        return False
+        return []
 
-    for key, enabled in enabled_plugins.items():
-        if enabled is not True:
-            continue
-        if key == plugin_name or key.startswith(f"{plugin_name}@"):
-            return True
-    return False
+    return [
+        key
+        for key, enabled in enabled_plugins.items()
+        if enabled is True and (key == plugin_name or key.startswith(f"{plugin_name}@"))
+    ]
+
+
+def _plugin_enabled_in_settings(settings_path: Path, plugin_name: str) -> bool:
+    """Return True when the plugin is enabled in a Claude settings file."""
+    return bool(_enabled_plugin_keys(settings_path, plugin_name))
 
 
 def _candidate_project_roots(cwd: Path | None = None) -> list[Path]:
@@ -2299,6 +2303,12 @@ def _claude_code_plugin_provides_hooks() -> bool:
       - installed: ``plugins[]`` in ~/.claude/plugins/installed_plugins.json,
                    carrying the scope and the resolved installPath.
 
+    The two files must agree on the SAME fully-qualified key (e.g.
+    ``ormah@some-market``), not merely both mention an "ormah"-prefixed key:
+    a stale, still-installed marketplace entry must never stand in for the
+    marketplace that is actually enabled, or a broken active install could
+    hide behind a healthy but abandoned one.
+
     This predicate licenses deleting the user's own wiring, so it requires both,
     plus hooks/hooks.json AND .mcp.json under that installPath — the wire guard
     strips both the CLI hooks and the CLI's MCP entry, so proving only the
@@ -2317,7 +2327,8 @@ def _claude_code_plugin_provides_hooks() -> bool:
     Fails open: any unreadable or unparseable config returns False, so setup
     wires exactly as it does today.
     """
-    if not _plugin_enabled_in_settings(Path.home() / ".claude" / "settings.json", "ormah"):
+    enabled_keys = _enabled_plugin_keys(Path.home() / ".claude" / "settings.json", "ormah")
+    if not enabled_keys:
         return False
 
     registry_path = Path.home() / ".claude" / "plugins" / "installed_plugins.json"
@@ -2330,9 +2341,8 @@ def _claude_code_plugin_provides_hooks() -> bool:
     if not isinstance(plugins, dict):
         return False
 
-    for key, entries in plugins.items():
-        if not isinstance(key, str) or key.split("@")[0] != "ormah":
-            continue
+    for key in enabled_keys:
+        entries = plugins.get(key)
         if not isinstance(entries, list):
             continue
         for entry in entries:
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -2290,6 +2290,57 @@ class AgentDescriptor:
     platform: list[str] | None = field(default=None)
 
 
+def _claude_code_plugin_provides_hooks() -> bool:
+    """True when a user-scoped ormah plugin is enabled AND actually installed.
+
+    Claude Code keeps the two states in two different files, and an enabled flag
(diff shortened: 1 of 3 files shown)


r-spade:
Thanks Andre. The duplicate-wiring issue is real and the overall shape here is right: setup should trust a verified user-scoped plugin and avoid writing the same hooks/MCP twice. I found one safety gap in the MCP manifest guard: an mcpServers.ormah key with an empty or non-plugin command could still license removing the working CLI MCP entry. I pushed a small follow-up commit on top that requires the plugin MCP entry to use the ormah-mcp wrapper command and adds regressions for empty/wrong MCP entries. Verified locally with the setup-focused tests: tests/test_setup.py tests/test_setup_json.py passed.

PR #153, opened 2026-07-21
fix(ingest): gate leading_orphan recovery on forward progress (ADR guard, fixes #149)

### LINKED ISSUE #149: bug(session-watcher): false-positive leading_orphan re-ingests the same transcript forever and strands its tail
## Symptom

The session watcher re-extracted the **same 788,113-char payload 36 times in 14 hours** (one transcript, cycle ≈ 27 min), writing ~773 near-duplicate memories. Each cycle is announced by:

```
[session_watcher] INFO: Session watcher recovering legacy mid-response cursor for <file>
```

followed by a full re-drain of the file from offset 0. Additionally, the ~530 KB tail of the transcript after the trigger point is **never ingested** — it is discarded on every cycle.

## Root cause (verified by deterministic replay against the real parser)

Two parser rules are mutually inconsistent:

1. `parse_transcript` commits a safe boundary right after an assistant record with a terminal `stop_reason` (`_assistant_is_terminal` → `_safe_end = f.tell()`), even with no following user record.
2. The leading-orphan check flags any text-bearing assistant record seen before the first user record when `start_offset > 0`:

```python
# transcript/parser.py
if text and not _saw_user_record and start_offset > 0:
    _leading_orphan = True
```

A legitimate Claude Code transcript can contain `assistant(end_turn)` **immediately followed by another assistant record** — in the observed case an error/continuation record:

```
assistant stop_sequence "API Error: Connection closed mid-response. The response above may be incomplete."
```

with the next `user` record ("continue") arriving 3 records later. Rule 1 parks the cursor exactly on that boundary; on the next tick, rule 2 misclassifies the healthy cursor as a legacy mid-response cursor.

The recovery path then rewinds to 0:

```python
# background/session_watcher.py
if result.leading_orphan:
    # ... A one-time re-ingest of this file; ...
    logger.info("Session watcher recovering legacy mid-response cursor for %s", rel)
    prev_offset = 0
    result = parse_transcript(path, start_offset=0, max_bytes=flush_bytes)
```

"One-time" exists only in the comment — **no `legacy_recovered` marker is ever persisted**, and the unchanged-file skip is deliberately bypassed while `end_offset < size`. Since the trigger is a permanent property of the file's bytes, the loop is eternal and deterministic: re-drain all slices, land on the same boundary, trigger again.

The orphan is a false positive even by outcome: parsing from the flagged offset returns a perfectly usable result (`safe_end` advances past the error record, 1 user turn, ~7.8K chars) — but the recovery branch checks `leading_orphan` first and throws that result away, which is why the tail is never ingested.

Replay evidence: driving `parse_transcript` slice-by-slice from offset 0 with default `flush_bytes` reproduces the exact per-slice `turns=` signature logged in every one of the 26 recovery cycles, ending in `orphan=True` at the same offset every time. The ~27 min period is self-driven (sum of the per-slice LLM extraction calls), not a scheduler interval.

Note: the same `leading_orphan` → rewind pattern also exists in the hook path (`cli_adapter` whisper-store), so the hook side is exposed to the same class of bug.

## Ruled out

- Slice failure pinning the cursor — completed cycles do advance the cursor up to the boundary; the reset is the explicit rewind.
- Racing workers — a single-threaded replay reproduces the loop in full.
- State file reverted by another process — the persisted entry matches the watcher's own mid-cycle accumulation byte-for-byte.
- Exception between ingest and state save — none logged.

## Fix direction

Either alone breaks the loop; the first also un-strands the tail:

1. Don't rewind when the orphan-flagged parse still produced a usable boundary (`safe_end_offset > start_offset`): drop the orphan fragment and advance.
2. Persist a one-shot `legacy_recovered` marker in the state entry so the recovery can fire at most once per file.

## Repro test sketch

JSONL fixture: `user → assistant(end_turn) → assistant(stop_sequence, "API Error…") → user("continue") → assistant(end_turn)`, sized so the first `end_turn` boundary caps a slice. Drive the watcher ingest repeatedly with a fake engine; assert the cursor is strictly monotonic, "recovering legacy" is logged at most once, and no slice is ingested twice.

AndreLFSMartins:
Fixes #149.

## What

A false-positive `leading_orphan` — an `assistant(stop_reason=end_turn)` safe boundary followed by an `assistant("API Error…")` record before the next user turn — made the recovery rewind re-ingest the same transcript **forever** (observed: the same 788K-char payload extracted 36× in 14h, ~773 near-duplicate memories) while the ~530KB tail past the boundary was never ingested.

This PR gates the rewind on **forward progress** with a single shared predicate, so the watcher and the hook path cannot diverge:

- `should_rewind(result, start_offset)` in `transcript/parser.py`: rewind only when `result.leading_orphan and result.safe_end_offset <= start_offset` (no forward progress). An orphan **with** progress is a false positive: the fragment is dropped and the cursor advances — which also un-strands the tail.
- `background/session_watcher.py` and `adapters/cli_adapter.py` (whisper store) both call it instead of checking `leading_orphan` directly.
- Post-rewind park guard (found in review): when the whole-file re-parse still lands `safe_end_offset <=` the original cursor (orphan still in-flight), both consumers now **park** (`NO_PROGRESS` / silent exit) instead of re-ingesting the closed prefix on every tick.

Genuine legacy mid-response cursors (orphan **and** no progress, with recoverable content) still rewind once, as before.

## Tests (TDD, red-first)

- Parser: 5 new tests — the #149 byte pattern (orphan flagged, progress, no rewind, `API Error` excluded from safe conversation), no-progress legacy cursor still rewinds, no-orphan-no-rewind, large-orphan variant, and a pinning test documenting the deliberate bounded drop of a legacy fragment tail when later closed turns exist.
- Watcher: 5 new tests — cursor monotonic across two ticks with a single extraction, cheap parse-only no-op below `min_turns` on an active file, accepted-drop pinning, legacy recovery preserved, and in-flight orphan parking with zero re-ingest and untouched state.
- Whisper hook: 2 new tests — cursor advances to EOF sending only the tail (no full re-extract, orphan excluded), and in-flight orphan parking with cursor untouched. The existing legacy-recovery test is untouched and still passes.

Full suite: 1449 passed; the 11 remaining failures are environment-dependent (global `~/.config/ormah/.env` leaking into bare `Settings()`) and fail identically on the merge base `f8ff9d7` — verified on a clean worktree. `ruff check src/ tests/`: clean.

## Known residual (deliberate, filed separately)

A session whose **final** record is a *terminal* API-error orphan still pays one whole-file re-ingest (the pre-existing legacy-recovery path; cursor then reaches EOF and parks — no loop). Fixing it requires classifying orphan provenance, which this design deliberately avoids; follow-up issue to be filed.

### DIFF
--- a/src/ormah/adapters/cli_adapter.py
+++ b/src/ormah/adapters/cli_adapter.py
@@ -450,8 +450,14 @@ def cmd_whisper_store(args):
             # prompt. With forward progress the orphan is a false positive (ADR-0003,
             # #149): drop the fragment and advance, or every hook fire re-extracts the
             # whole transcript.
+            original_offset = start_offset
             start_offset = 0
             result = parse_transcript(path, start_offset=0)
+            if result.safe_end_offset <= original_offset:
+                # The rewind made no progress: the "orphan" tail is a still-open in-flight
+                # response, not a recoverable one. ADR-0003: a no-progress transcript
+                # parks, it does not re-extract the closed prefix on every hook fire.
+                sys.exit(0)
     except Exception:
         sys.exit(0)
 
--- a/src/ormah/adapters/cli_adapter.py
+++ b/src/ormah/adapters/cli_adapter.py
@@ -440,13 +440,16 @@ def cmd_whisper_store(args):
     if start_offset >= path.stat().st_size:
         sys.exit(0)
 
-    from ormah.transcript.parser import parse_transcript
+    from ormah.transcript.parser import parse_transcript, should_rewind
 
     try:
         result = parse_transcript(path, start_offset=start_offset)
-        if result.leading_orphan:
-            # Cursor left mid-response by an older version: re-parse from the start to
-            # recover the dropped tail with its prompt (one-time full re-extract).
+        if should_rewind(result, start_offset):
+            # Orphan with NO forward progress: a genuine cursor left mid-response by an
+            # older version — re-parse from the start to recover the dropped tail with its
+            # prompt. With forward progress the orphan is a false positive (ADR-0003,
+            # #149): drop the fragment and advance, or every hook fire re-extracts the
+            # whole transcript.
             start_offset = 0
             result = parse_transcript(path, start_offset=0)
     except Exception:
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -18,7 +18,7 @@ from watchdog.observers import Observer
 
 from ormah.engine.memory_engine import MemoryEngine
 from ormah.text.tokens import distinctive_tokens
-from ormah.transcript.parser import TranscriptResult, TranscriptTurn, parse_transcript
+from ormah.transcript.parser import TranscriptResult, TranscriptTurn, parse_transcript, should_rewind
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


r-spade (approved):
Thanks Andre. This is a useful fix and I agree the need is real: a false-positive leading_orphan after a safe assistant boundary can otherwise force repeated whole-file re-ingestion while stranding the tail.

I reviewed the parser predicate plus the watcher and whisper hook call sites, rebased the branch onto current main, and kept the newer session-watcher reconcile/startup-shutdown behavior intact. Verification: focused #149 regressions passed locally, the relevant parser/session-watcher/whisper-out files passed locally, and the GitHub test job is green. The remaining CI lint failure is the known repo-wide Ruff blocker, not this fix. This looks good to merge.

PR #208, opened 2026-08-11
fix(lock-order): hoist the index sweep's FileStore calls out of its write txn (fixes #207)

### LINKED ISSUE #207: Deadlock: lock-order inversion in IndexBuilder.incremental_update freezes all writes ~2 min after start (0.14.8)
## Summary

On 0.14.8, a lock-order inversion between the restore-exclusion RLock (`L_mem`) and `Database._lock` (`L_db`) deadlocks the server roughly **2 minutes after every start**. Reads keep serving (`/admin/health` answers in 2ms throughout), but every write — `remember`, whisper stores, `/agent/recall` (which writes via `_log_decision`) — hangs forever. The server looks healthy while being completely write-dead.

## Root cause

Two lock orders coexist:

- `IndexBuilder.incremental_update` opens its write transaction (**`L_db`**, held across `Database.transaction`'s yield) and only then calls `file_store.list_paths()` and `file_store.file_hash()`. Both are decorated with the restore-exclusion lock, because the engine hands `FileStore` its own lock (`FileStore(nodes_dir, self._memory_operation_lock)`). That is **`L_db` → `L_mem`**.
- Every background job decorated with `@serialized_memory_job` takes **`L_mem` → `L_db`** (the decorator acquires `L_mem`, then the job's writers open transactions).

Opposite acquisition orders on two locks form a deadlock cycle, and `index_updater` runs every **60 seconds** — so any background job overlapping with the index sweep locks the whole write side permanently.

## Evidence (live 0.14.8 install, ~36k nodes)

- `py-spy dump` captured the cycle directly: `auto_linker` holding `L_mem` (`acquired: True`) waiting on `L_db` at `auto_linker.py:312`, while `index_updater` held `L_db` waiting on `L_mem` at `builder.py:119` — both stamped with an identical start time.
- An external `BEGIN IMMEDIATE` probe returned `database is locked` on 10 consecutive attempts over 60s while read-only connections passed — the write lock was held, not leaked.
- `/admin/health` answered in 2ms the whole time; `/agent/whisper` and `/agent/recall` never answered.

## Reproduction

The linked PR includes a regression test that drives the exact interleaving (builder inside its write txn reaching for `L_mem` while a memory job holds `L_mem` and reaches for `L_db`). It hangs past a 10s join on current `main`.

## Fix

PR: hoist the two `FileStore` calls above the transaction, removing the inversion at the root — the same pattern `full_rebuild` already uses in the same file. Decorating the job with `@serialized_memory_job` instead was considered and rejected: it would hold `L_mem` across the entire sweep (36k files here) every 60s, trading the deadlock for sustained write starvation.

AndreLFSMartins:
Fixes #207

## Problem

`incremental_update` opened the write txn (`L_db`) and then called `file_store.list_paths` and `file_hash`, both decorated with the engine's restore-exclusion RLock (`L_mem`, handed to `FileStore` by the engine): `L_db → L_mem`. Every `@serialized_memory_job` background job takes `L_mem → L_db`. Opposite orders on two locks deadlock, and `index_updater` runs every 60s — a live 0.14.x server froze all writes ~2 minutes after each start while reads kept serving.

Verified live by py-spy on the reporting install: `auto_linker` held `L_mem` (`acquired: True`) waiting on `L_db` at `auto_linker.py:312` while `index_updater` held `L_db` waiting on `L_mem` at `builder.py:119`, both jobs stamped with an identical t0. An external `BEGIN IMMEDIATE` returned `database is locked` on 10 consecutive probes over 60s while read-only connections passed. `/admin/health` answered in 2ms throughout; `/agent/whisper` and `/agent/recall` never answered at all.

## Fix

Hoisting both calls above the txn removes the inversion at the root without retaining any new lock. Adding `@serialized_memory_job` to the job was rejected: it would hold `L_mem` across the whole loop (36k files on the reporting store) every 60s, trading the deadlock for sustained write starvation. `full_rebuild` already hoists `list_paths` — this aligns `incremental_update` with the pattern in its own file, and moves half the loop's I/O out from under `L_db`.

The hashes are pre-computed in an explicit loop rather than a comprehension so that a file removed between listing and hashing stays non-fatal: the loop's own try/except used to absorb exactly that, and a bare comprehension would let it kill the job.

## Test

`test(index): reproduce the incremental_update lock-order deadlock` drives the exact interleaving: the builder holds `L_db` inside its write txn and reaches for `L_mem` via `file_store.list_paths`, while a memory job holds `L_mem` and reaches for `L_db`. It hung past a 10s join before this change and passes after it.

### DIFF
--- a/src/ormah/index/builder.py
+++ b/src/ormah/index/builder.py
@@ -37,11 +37,20 @@ class IndexBuilder:
 
         # Two-pass: nodes first, then edges (to satisfy FK constraints)
         paths = list(self.file_store.list_paths())
+        hashes: dict[Path, str] = {}
+        for path in paths:
+            try:
+                hashes[path] = self.file_store.file_hash(path)
+            except Exception as e:
+                logger.warning("Failed to hash %s: %s", path, e)
+
         count = 0
         with self.db.transaction():
             for path in paths:
+                if path not in hashes:
+                    continue
                 try:
-                    self._index_file_nodes_only(path)
+                    self._index_file_nodes_only(path, hashes[path])
                     count += 1
                 except Exception as e:
                     logger.warning("Failed to index %s: %s", path, e)
@@ -67,12 +76,8 @@ class IndexBuilder:
         indexed_ids = set(indexed.keys())
         disk_ids: set[str] = set()
 
-        # Both FileStore calls take L_mem -- the engine hands FileStore its own lock
-        # (FileStore(nodes_dir, self._memory_operation_lock)). Calling them inside the write txn
-        # would request L_mem while holding L_db, the reverse of the L_mem -> L_db order every
-        # @serialized_memory_job background job takes: a deadlock cycle, and this job runs every
-        # 60s. Hoisting also halves the I/O done under L_db. full_rebuild already hoists
-        # list_paths for the same reason.
+        # FileStore calls take L_mem. Complete them before the write transaction so no builder
+        # path requests L_mem while holding L_db, the reverse of the order used by memory jobs.
         paths = self.file_store.list_paths()
         hashes: dict[Path, str] = {}
         for path in paths:
@@ -93,11 +98,11 @@ class IndexBuilder:
                     disk_ids.add(node.id)
 
                     if node.id not in indexed:
-                        self._index_file(path)
+                        self._index_file(path, file_hash)
                         added += 1
                     elif indexed[node.id] != file_hash:
                         self._remove_node(node.id, keep_vectors=True)
-                        self._index_file(path)
+                        self._index_file(path, file_hash)
                         updated += 1
                 except Exception as e:
                     logger.warning("Failed to process %s: %s", path, e)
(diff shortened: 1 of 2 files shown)


r-spade:
Thanks Andre : the diagnosis was right, but one nested FileStore call was missed: new/changed files still reached `_index_file_nodes_only() -> file_hash()` while holding `L_db`; `full_rebuild()` and `index_single()` had the same path.

I added `1c65fc4` to precompute and pass hashes before transactions, plus an invariant test covering every builder entry path. The same A/B probe deadlocks on `ba3a359` and completes with the fix; locally, 1,839 tests and Ruff pass.

My tests should have caught a breaking change like this. Apologies it did not. 
Adding the lock to the FileStore had a bigger blast radius than I had thought. Thanks for catching and fixing. Really appreciate it.

r-spade (approved):
Thanks Andre. The deadlock diagnosis is correct, and the completed fix now removes every FileStore acquisition from IndexBuilder transaction bodies. The A/B deadlock probe, invariant test, full local suite, Ruff, and CI all pass. This is ready to merge.

PR #234, opened 2026-08-16
fix: surfacing must not be confirmed use (#220)

AndreLFSMartins:
## Supersedes #229 for issue #220

#229 is a draft covering #220–#223 together on `feat/issue-220-223-memory-lifecycle`. This PR
supersedes it **for #220's scope only** — it does not implement #221 (bounded reinforcement),
#222 (retrievability-driven decay), or #223 (seven-day lease/promotion/provenance), which remain
open and unaddressed here. If this PR lands, #229 should be narrowed to #221–#223 or closed in
favor of separate PRs for each.

## What this does (#220 — surfacing must not be confirmed use)

- Removes lifecycle writes from every search/surface path (hybrid recall, FTS fallback,
  structured recall, formatted recall, UI search route, whisper/context loading).
- Adds one serialized `_record_confirmed_use` lifecycle operation.
- `recall_node(id)` confirms exactly the requested node; returned neighbours stay non-mutating.
- Positive `explicit`/`implicit` feedback confirms exactly its resolved node; `auto_heuristic`
  is excluded pending #218.
- `recall_search_structured`'s tuning parameters (`min_relevance`, `auto_temporal`,
  `spread_activation`, `query_vec`) are now keyword-only, closing a silent-corruption trap left
  behind when `touch_access` was removed: a stale positional call used to have its `False`
  reinterpreted as `min_relevance=0`, dropping the relevance floor with no error. All in-repo
  call sites already pass keyword args, so nothing else changes.
- Guards `_record_confirmed_use` against `stability == 0` (matches the same guard already
  present in `decay_manager.py` and `importance_scorer.py`) and isolates `recall_node`'s
  reinforcement behind `try/except`, matching `submit_feedback` and the session watcher.

## Review history

An internal two-peer review (Cursor + Codex) found the stability/isolation gap above and one
finding this PR does not fix: `touch_access`'s removal narrows `recall_search_structured`'s
signature, and the same review's "keyword-only" recommendation is what the last commit
implements. Full contract suite: `tests/test_engine/test_confirmed_use_contract.py`, 30 passing
tests covering non-mutation, claim exclusivity, and the fixes above.

## Known pre-existing failures (not from this branch)

12 test IDs fail locally regardless of this branch — 4 environmental root causes (fastembed
cache path assumptions, Codex MCP config fixtures, and two flaky background-job timing tests),
none reproduced in clean CI. None touch the code this PR changes.

## Out of scope, filed separately

[...]

Issue #220, opened 2026-08-13
bug(lifecycle): separate surfaced results from confirmed memory use


r-spade:
## Problem

Ormah currently treats most returned search results as memory use. `recall_search` and the default structured-search path call `_touch_access` for each direct result, and that one helper updates `access_count`, `last_accessed`, `last_review`, and stability together.

A result appearing in a list does not prove that the caller read or used it. This creates a self-reinforcing loop: frequently surfaced memories accumulate access history and lifetime even when ignored.

## Decision from #191

Separate **surfacing** from **confirmed use**.

Surfacing events include:

- appearing in a recall result set;
- bare whisper injection;
- UI search results;
- graph activation or conflict expansion.

They may be logged for retrieval evaluation, but must not:

- increment `access_count`;
- advance the confirmed-use/decay anchor;
- update stability;
- promote an archival node.

Confirmed-use events are:

- deliberately fetching one node through `recall_node(id)`;
- source-qualified positive feedback (`explicit`, `implicit`, or `auto_llm_judge`).

`auto_heuristic` remains excluded until #218 provides meaningful signal calibration. Negative feedback remains prompt-specific affinity evidence and does not change stability.

## Suggested implementation

- Replace the overloaded `_touch_access` contract with explicit surfacing/logging and confirmed-use paths.
- Ensure all recall implementations and fallbacks follow the same semantics.
- Keep retrieval/whisper event logging intact so ignored appearances remain observable.
- Make the confirmed-use operation reusable by the reinforcement and promotion work.

## Acceptance criteria

- Broad recall returning N nodes does not change lifecycle fields on any returned node.
- UI search does not change lifecycle fields.
- Bare whisper and graph-added results do not change lifecycle fields.
- `recall_node(id)` records confirmed use for exactly the requested node, not its neighbours.
- Qualified positive feedback records confirmed use for exactly its resolved node.
- Tests cover hybrid and FTS fallback paths.

## Related

- #191 — design decision
- #218 — signal-strength calibration; not required for the source-qualified rule

AndreLFSMartins:
Per Discord (2026-08-14): #220–#223 are assigned to me, implemented as separate PRs, with preferred landing order #220 → #222 → #221 → #223. Draft PR #229 will be closed as superseded; its branch remains temporarily as optional reference. Starting here with #220.

Issue #191, opened 2026-08-02
design debate: memory lifecycle — what should "archival" mean, and should a whisper count as a use?


AndreLFSMartins:
First design debate issue, as we discussed in #144. Keeping it deliberately short: my write-up on #31 was ~18k characters, which is a big part of why it's been sitting — that was a proposal asking for approval, and this is three questions asking for your opinion. No code attached, nothing to review.

Context is the memory lifecycle (#28 / #31). Since that write-up I measured its claims against my live store (31,648 nodes) and against the current local model, and **three of my own claims didn't survive** — so before I write another line of it, I'd rather agree on the shape with you.

## The one fact that frames all three questions

`fsrs_initial_stability` is defined and validated in `config.py:110` — and has **no reader anywhere outside `config.py`**. New nodes get `stability = 1.0` from the Pydantic default in `models/node.py:74` instead. Both defaults are `1.0`, so the knob looks wired and setting `ORMAH_FSRS_INITIAL_STABILITY=7` changes nothing at all.

With `fsrs_decay_threshold = 0.3`, `R = exp(-t/1.0)` crosses the line at **~29 hours**. On my store: **99.4% of nodes are archival, and 99.1% still sit at `stability = 1.0` with `access_count = 0`.** So "archival" currently means "older than a day", not "cold" — and every gate keyed to retrievability or importance is selecting on noise. You'd barely see this at 3k nodes; it dominates at 31k.

That's a one-line fix I could send today. I don't think I should, because it changes what the whole tier system *means*, and that's a design call, not a bug.

## Q1 — What should "archival" mean?

If the knob gets wired, we pick the working window: ~7 days of initial stability puts demotion at ~8–9 unsignaled days; ~12–17 days gives a 2–3 week window. But the real question is what the tier is *for*. Whisper only reads core + working, so the window is effectively "how long a memory stays eligible to be whispered without anyone touching it." Weeks? A month? Something adaptive?

## Q2 — Should a whisper injection count as using a memory?

Today it doesn't: only explicit `recall` bumps stability. That's why 99.1% of my nodes never move — the system whispers a memory into context hundreds of times and still calls it unused.

The obvious fix (injection reinforces stability) has a real failure mode: injected nodes rank better, so they get injected more, and the rich get richer until the whisper pool is a handful of nodes. A weak, capped signal (say 1×/day/node) plus negative feedback lowering stability is the version I'd argue for — but I'd like your read, since this directly shapes what users experience day to day.

[...]

PR #235, opened 2026-08-16
fix(lifecycle): retrievability alone drives working→archival decay (#222)

### LINKED ISSUE #222: bug(lifecycle): stop importance from permanently blocking working-tier decay
## Problem

Working-to-archival decay currently requires both:

```text
R < fsrs_decay_threshold
importance < decay_importance_threshold
```

Importance combines cumulative access count, graph edge count, and an FSRS-derived recency signal. Cumulative signals can permanently exceed the default `0.5` gate even when recency falls to zero.

With current defaults, 50 accesses plus four edges produce a permanent non-recency contribution of approximately `0.51445`. Such a node can never leave `working`, however long it remains unused.

The configured `importance_recency_half_life_days=14` is currently unread; importance instead reuses FSRS retrievability, coupling two different concepts.

## Decision from #191

- Normal `working → archival` transition is controlled by retrievability alone.
- Remove importance as a pre-gate for tier demotion.
- Wire importance's own recency signal using `importance_recency_half_life_days`.
- Keep importance for ranking, display, and core-cap prioritization.
- Use `core` as the explicit mechanism for permanent whisper eligibility.

Archival is reversible dormancy, not deletion, so an important but unused node remains available to deliberate recall and can return through confirmed use.

## Acceptance criteria

- A working node below the retrievability threshold becomes archival regardless of historical access or edge counts, unless protected by an explicit lifecycle rule such as `core`/identity handling.
- A node with 50 accesses and four edges does not remain working forever solely because of importance.
- Importance recency follows its configured 14-day half-life and no longer depends on stability.
- Existing importance uses outside decay continue to work.
- Tests cover high-importance stale nodes, custom importance weights, and independent recency behaviour.
- User-facing configuration/background-job documentation is updated.

## Related

- #191 — design decision
- #28/#31 bounded forgetting should remain gated on corrected lifecycle signals

AndreLFSMartins:
Closes #222. Implements the #191 decision: retrievability alone controls the `working → archival` transition, and importance gets its own recency clock.

## What changed

**`decay_manager.run_decay`** no longer reads `importance` at all — the query dropped the column (`SELECT id, stability, last_review, last_accessed`) and the pre-gate is gone. Demotion is decided by FSRS retrievability against `fsrs_decay_threshold`.

**`importance_scorer`** got `_recency_signal(days_ago, half_life_days)`, a half-life decay on importance's own clock. Previously importance reused FSRS retrievability (`exp(-days/stability)`), so a high-stability node read as permanently recent — the coupling #222 describes. The scorer no longer selects `stability`.

**`config.py`** gained a `field_validator` for `importance_recency_half_life_days`, rejecting non-finite and non-positive values. Without it, `0.0` raised `ZeroDivisionError` and aborted the whole 120-minute scoring job rather than skipping one node, and `NaN`/`inf` silently saturated the signal to `1.0`. `_recency_signal` carries the same guard as defence in depth.

**Docs** (`05 - Background Jobs`, `12 - Configuration Reference`) and the admin-panel task description in `routes_admin.py` now describe what the job actually does.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Stale working node demotes regardless of access/edge history | `test_high_importance_stale_node_is_decayed` (importance 0.9) |
| 50 accesses + 4 edges does not pin a node to `working` | `test_accumulated_access_and_edges_do_not_pin_a_node_to_working` — drives the real `run_importance_scoring` end-to-end, asserts the computed importance (**0.5892**) clears the old 0.5 gate *before* asserting demotion |
| Recency follows the configured half-life, independent of stability | `test_recency_signal_*` + the scorer no longer selecting `stability` |
| Importance uses outside decay still work | `TierManager.enforce_core_cap()` is the only remaining production reader |
| Tests cover stale high-importance, custom weights, independent recency | `test_decay_manager.py`, `test_importance_scorer.py`, `test_config.py` |
| User-facing config/background-job docs updated | `docs/05`, `docs/12`, `routes_admin.py` |

## Blast radius

The formula change makes old nodes score materially higher — a 30-day never-reviewed node goes from recency ≈ 9e-14 to ≈ 0.227. Sweeping every consumer of `nodes.importance` found **exactly one production reader left**: `TierManager.enforce_core_cap()` (`tier_manager.py:56-60`). Not search ranking (`docs/03:113` says so explicitly), not whisper, not context building, not the API, not the UI. The effect is a relative ordering among `core` nodes, bounded by the 0.33 recency weight.

Surviving protections after the gate removal: the self node (`decay_manager.py:47-48`), the `core` tier (never enters the `WHERE tier = 'working'` query), and `R < fsrs_decay_threshold` (0.3).

Decay timescales, derived from stability being access-derived:

| Node | Before | After |
|---|---|---|
| 50 accesses, stability 30 | pinned to `working` forever | demotes after ~36 days idle |
| stability at the 365-day cap | — | holds ~439 days |
| never accessed | ~1.2 days | ~1.2 days (unchanged) |

Archival is reversible dormancy, not deletion.

## Notes

`decay_importance_threshold` is deliberately left in `config.py` with no production reader on this branch, with a comment saying so. Bounded forgetting (#28/#31) reintroduces a reader as the deletion-protection gate, so removing it here and re-adding it there would be churn. Flagging it explicitly since a reviewer will notice config with no reader.

This branch is cut from `main` and does not depend on #234 (#220).

## Verification

- Full suite: **13 failed, 1836 passed** vs the same baseline measured on `main` before the branch: **13 failed, 1826 passed**. Same 13 failures, name for name — they are pre-existing environment leakage, unrelated to this change. Zero regressions, +10 passing.
- `ruff check src/ tests/` → `All checks passed!`

## Known follow-up

Two test-coverage gaps found in review, not fixed here, both about pinning invariants rather than production defects:

1. No decay test sets `stability`, so with the default `S = 1.0` at 30 days both FSRS (`≈9e-14`) and the half-life (`≈0.23`) fall under the 0.3 threshold — swapping `run_decay`'s formula for the half-life one would leave the suite green. A pair with `stability = 100` (stays `working`, `R ≈ 0.74`) and `stability = 1.0` (goes `archival`) would pin it.
2. `run_importance_scoring` is never driven with a modified `importance_recency_half_life_days`; the setting is only exercised in the pure helper and in config validation. Hardcoding `14.0` in the job would pass every test.

Happy to add both here if you would rather they land with this change.

### DIFF
--- a/docs/05 - Background Jobs.md
+++ b/docs/05 - Background Jobs.md
@@ -129,7 +129,9 @@ It does not do embedding clustering or hierarchical clustering despite the name.
 Importance does not protect a node from decay. Cumulative signals such as access
 and edge counts could otherwise push a stale node permanently above any threshold.
 Archival is reversible dormancy, not deletion — a demoted node stays reachable by
-deliberate recall. Use the `core` tier for permanent whisper eligibility.
+deliberate recall. Use the `core` tier for permanent whisper eligibility. Importance
+still decides which nodes survive when `core` is over its cap — see
+[01 - Data Model](<./01 - Data Model.md>) for how `enforce_core_cap()` uses it.
 
 ### Importance Scorer
 
--- a/docs/05 - Background Jobs.md
+++ b/docs/05 - Background Jobs.md
@@ -124,9 +124,12 @@ It does not do embedding clustering or hierarchical clustering despite the name.
 ## Importance and Decay
 
 - `importance_scorer` computes a normalized multi-signal importance value
-- `decay_manager` uses FSRS-style retrievability and importance to decide whether to demote working memories
+- `decay_manager` uses FSRS-style retrievability alone to decide whether to demote working memories
 
-High-importance nodes are protected from decay.
+Importance does not protect a node from decay. Cumulative signals such as access
+and edge counts could otherwise push a stale node permanently above any threshold.
+Archival is reversible dormancy, not deletion — a demoted node stays reachable by
+deliberate recall. Use the `core` tier for permanent whisper eligibility.
 
 ### Importance Scorer
 
@@ -134,7 +137,7 @@ High-importance nodes are protected from decay.
 
 1. **Access signal**: how often the node has been recalled, based on `access_count`
 2. **Edge signal**: how connected the node is in the graph, based on total edge count
-3. **Recency signal**: how retrievable the node currently is, using FSRS-style `exp(-days_ago / stability)`
+3. **Recency signal**: how recently the node was touched, using half-life decay `exp(-ln(2) * days_ago / importance_recency_half_life_days)` — independent of FSRS stability
 
 With the default configuration, those signals are weighted almost evenly:
 
@@ -150,8 +153,6 @@ Access and edge counts are log-normalized against reference values, then the wei
 The scorer runs every `120` minutes by default and only writes a new value when the change is meaningful.
 
 This score is not static. Recall and search hits update `access_count`, `last_accessed`, `last_review`, and `stability`, so a memory's importance changes over time as it is used, connected, or left untouched.
-
-Important current nuance: `importance_recency_half_life_days` exists in configuration, but the current scorer implementation uses FSRS stability for the recency term instead.
 
 ## Walkthrough Example
 
--- a/docs/12 - Configuration Reference.md
+++ b/docs/12 - Configuration Reference.md
@@ -183,6 +183,10 @@ Important note: current search applies tier as a multiplicative factor after con
 | `importance_recency_half_life_days` | `14.0` |
 | `decay_importance_threshold` | `0.5` |
 
+`decay_importance_threshold` no longer affects `working -> archival` decay, which
+depends on retrievability alone (#222). It is retained for configuration
+compatibility and currently has no effect.
+
 ## Whisper-Out and Nudge
 
 | Setting | Default |
--- a/docs/12 - Configuration Reference.md
+++ b/docs/12 - Configuration Reference.md
@@ -184,8 +184,8 @@ Important note: current search applies tier as a multiplicative factor after con
 | `decay_importance_threshold` | `0.5` |
 
 `decay_importance_threshold` no longer affects `working -> archival` decay, which
-depends on retrievability alone (#222). It is retained for configuration
-compatibility and currently has no effect.
+depends on retrievability alone (#222). It remains a documented setting with no
+effect in this version; bounded forgetting reintroduces a consumer.
 
 ## Whisper-Out and Nudge
 
--- a/src/ormah/api/routes_admin.py
+++ b/src/ormah/api/routes_admin.py
@@ -38,7 +38,7 @@ _TASK_DESCRIPTIONS = {
     "auto_linker": "Discovers and creates edges between semantically related memories.",
     "auto_cluster": "Groups memories into clusters based on semantic similarity and tags.",
     "consolidator": "Merges or summarizes redundant working-tier memories into consolidated entries.",
-    "decay_manager": "Applies time-based decay to memory importance, demoting stale unused memories.",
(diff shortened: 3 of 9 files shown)


r-spade:
Thanks Andre — I reviewed the decay/importance changes and the core implementation is sound. The half-life formula is correct, the configured value is wired into the full scorer, and the retrievability-only decay behavior matches #222.

I reproduced one small batch-isolation gap: after the importance pre-gate is removed, a high-importance row with a valid but timezone-naive timestamp can reach `now - anchor`, raise `TypeError`, and stop the whole decay run before later valid nodes are processed. The same profile passes on current `main` only because that high-importance row is skipped. Normal Ormah writers produce aware UTC timestamps and the measured store has no naive values, so this is defensive hardening for edited/imported data rather than evidence of current corruption.

I also agree with the two discriminator tests you offered in the PR description: `S=1` versus `S=100` at the same age, and an end-to-end scorer test using a non-default importance half-life.

Since this is small and maintainer edits are enabled, I’m going to add a focused follow-up commit with per-row timestamp isolation and those three regression tests, then ask you to review that commit before we merge. Please do flag any concern with the handling or test shape — I’d value your read on it.

r-spade:
I’ve added the focused maintainer follow-up as commit `8bd8192` (`fix(lifecycle): isolate invalid decay timestamps`). It is a fast-forward addition on your branch; none of your commits were rewritten.

The commit:

- keeps timestamp parsing and aware/naive subtraction inside the per-node error boundary, logs the affected short node ID, and continues the decay batch;
- proves one malformed high-importance row cannot prevent a later valid stale row from archiving;
- proves equal-age `S=1` and `S=100` nodes take different decay outcomes through the real job;
- proves the full importance scorer honors a configured seven-day half-life rather than hardcoding the default.

Focused verification is green: 83 tests passed, Ruff passed, and `git diff --check` is clean. Fresh PR CI is now running against current `main`.

When you have a moment, could you review this commit and let me know whether the failure handling and test shape match your intent? I’ll hold the merge for your feedback.

AndreLFSMartins:
This looks good. The fix is exactly the right shape: keeping the arithmetic inside the same try/except boundary instead of adding a second one, and the three tests are the ones I flagged in the PR description. Nice turnaround.

One thing worth calling out explicitly before we merge: the fix is fail-open. A node with a naive or otherwise unparseable timestamp now never decays. It just logs a warning on every run, forever. I think that's the right default: we shouldn't silently touch data we can't interpret. But right now it's implicit in the code, not a decision anyone made on purpose. If a writer ever produces naive timestamps at scale, this is how we'd find out: a warning line in a log nobody's watching.

Want your read: is "skip forever, warn every run" the intended behavior, or does it deserve a tracking issue or a counter somewhere so it doesn't stay silent? Either way, I'm fine merging as-is. This isn't a blocker, just want the policy on record.

r-spade:
Thanks Andre — agreed, and I think it is worth making the policy explicit.

At the decay layer, `skip this node and warn` is the intended behavior. If the timestamp does not identify an unambiguous instant, we should not guess a timezone and perform a tier mutation. Keeping that one node working is the safer failure direction; the rest of the batch must continue.

I would not define “skip forever” as the permanent system policy, though. #249 now tracks the root invariant: normalize every supported Markdown/model timestamp to an aware datetime at the boundary (interpreting naive values as UTC for compatibility), while retaining this per-row isolation for direct-index corruption or unsupported external writes. The measured 1,943-node store currently has zero naive `last_accessed` or populated `last_review` values, so this is defensive rather than an observed backlog.

For now, the per-node warning is adequate observability and is not a merge blocker. If #249 or production data finds non-zero volume, we can add a surfaced maintenance counter rather than expanding this PR pre-emptively.

Thank you for checking the failure policy explicitly. With your review and the fresh green CI, I consider #235 ready from my side and will merge now

Issue #238, opened 2026-08-17
bug(engine): memory-operation exclusion is intra-process only; a CLI and the server race on one memory dir


AndreLFSMartins:
`MemoryEngine._memory_operation_lock` is an in-process `RLock`. It serializes memory operations within one Python process and provides **no exclusion at all** between processes sharing a memory directory.

That configuration is normal, not exotic: the server runs under launchd while the user invokes the `ormah` CLI, and both open the same memory dir.

## Surface

This is not specific to one operation. Everything guarded by `@_serialized_memory_operation` / `memory_operation()` has the same gap — `remember`, `update_node`, restore, and the reinforcement path that #221 added.

## Concrete case, from #221

#221 made reinforcement a check-then-write pair: read `last_review`, decide whether the per-day cooldown has elapsed, write a new `stability`. That is a TOCTOU, and it was correctly serialized against threads inside one process (`@_serialized_memory_operation`, with a regression test that fails `assert 4 == 1` when the decorator is removed).

Across two processes the same interleaving is unguarded: a CLI recall and a server recall can both read the pre-cooldown `last_review`, both conclude the node is due, and both reinforce — the exact double-bump the cooldown exists to prevent.

The window is small and the damage is bounded (one extra bounded increment, not the unbounded growth #221 fixed), which is why it is filed as its own issue rather than as a #221 blocker.

## Directions

Two that fit a local-first, SQLite-backed design:

1. A file lock (`flock`) in the memory directory, acquired by `memory_operation()`. Simple, portable enough, and covers the Markdown writes too — which SQLite-level locking would not.
2. Move the check-then-write pairs into `BEGIN IMMEDIATE` transactions and let SQLite arbitrate. Narrower, and does not help operations whose critical section spans a `file_store.save`.

Whichever is chosen, the deciding question is whether the critical section includes Markdown I/O. Most of these do, which argues for (1).

## Worth documenting either way

Even before a fix, the intra-process-only scope of the guarantee should be stated where someone would look for it — the `_serialized_memory_operation` docstring currently reads as though it establishes exclusion generally.

---

Found during the final whole-branch review of #221.

AndreLFSMartins:
A second, independent review (Codex, during the `/council-pr` peer round on the #221 branch) reached the same finding from a different direction, and added a detail this issue was missing.

**`StoreLock` already exists in this repository, specifically for cross-process store exclusion.**

[...]

Issue #236, opened 2026-08-16
bug(lifecycle): a restore onto an existing index skips the FSRS migration


AndreLFSMartins:
## Problem

Restoring a backup onto an installation whose SQLite index still exists never re-evaluates the
FSRS migration, so the restored nodes keep a default `stability` instead of being seeded from
`access_count`.

Three links in the chain, each verified against `a28837b`:

1. `IndexBuilder.full_rebuild` (`src/ormah/index/builder.py:24-36`) clears `node_tags`, `edges`,
   `nodes_fts`, `nodes` and `node_vectors`, but from `meta` it deletes only `auto_link_watermark`.
   Every other key survives the rebuild.
2. `MemoryEngine.reload_restored_graph` (`src/ormah/engine/memory_engine.py:1237-1244`) calls
   `rebuild_index()` and never calls `_migrate_fsrs`.
3. `_migrate_fsrs` (`src/ormah/engine/memory_engine.py:159`) is invoked only from `__init__`
   (line 151), and returns early on `if fsrs_migrated: return`.

So a pre-FSRS backup restored onto an already-migrated install finds `fsrs_migrated = '1'` still
in `meta`, skips the seed, and leaves those nodes at the default `stability` rather than
`min(30, access_count * 2)`.

## Impact

Restored nodes look maximally forgettable regardless of their real usage history. Decay reads
them as `S = 1` and can demote memories with a long access history to archival. The blast radius
grows once #222 (PR #235) makes tier demotion depend on retrievability alone.

`backup.py:331-334` excludes `index.db`, `index.db-shm` and `index.db-wal` from every backup, so
the index is always derived state on the restore path — but the lifecycle keys inside it are
treated as durable.

## Suggested fix

Treat the lifecycle keys the same way `auto_link_watermark` is already treated — as derived state
that a rebuild invalidates:

- Delete `fsrs_migrated` (and `lifecycle_model_version`, once #221 lands) inside `full_rebuild`.
- Call `_migrate_fsrs()` at the end of `rebuild_index` / `reload_restored_graph`.

The `last_review` guard #221 introduces then does the actual work of deciding whether the store is
genuinely pre-FSRS, because it finally gets to run on this path.

## Test that must fail first

Start from an engine that has already migrated, replace the Markdown with pre-FSRS nodes
(`last_review` NULL, `access_count = 5`, `stability = 1.0`), call `reload_restored_graph()`, and
assert `stability == 10.0`. On `main` today it stays at `1.0`.

[...]

PR #239, opened 2026-08-17
fix(lifecycle): bound stability reinforcement and add a per-day cooldown (#221)

### LINKED ISSUE #221: bug(lifecycle): bound stability reinforcement and add a per-day cooldown
## Problem

The current stability update is unbounded and has no spacing cooldown:

```text
S' = min(S × 1.5 × R^-0.2, 365)
```

Under the exponential curve, `R` approaches zero for old memories, so `R^-0.2` approaches infinity. An `S=1` node recalled after 30 days receives a spacing factor of about `403.43`; even with the proposed stability saturation alone it would jump to about `S=202.7` from one use.

Repeated same-session calls also compound: the current `days_since >= 0.001` floor allows ten immediate recalls to produce approximately `1.5^10 = 57×` growth.

## Decision from #191

Keep the exponential retrievability curve:

```text
R = exp(-t / S)
```

Use bounded, diminishing reinforcement on **confirmed use only**:

```text
spacing = min(R^-0.2, 2.0)

S' = min(
    S × (1 + g × S^-w × spacing),
    fsrs_max_stability
)

g = 0.5
w = 0.5
```

Allow at most one numeric stability increase per node per day. Every confirmed use still advances its recency/decay anchor, so this requires separate timestamps for:

1. most recent confirmed use;
2. most recent numeric stability update.

The constants are initial policy choices, not fitted values, and should remain configurable.

## Future curve changes

Centralize lifecycle math rather than duplicating formulas across the engine and background jobs. Replace boolean-only migration semantics with an integer lifecycle-model version capable of representing future migrations.

A future curve migration should preserve each node's archival deadline, not attempt a constant rescale of the full curve.

## Acceptance criteria

- An `S=1` node confirmed-used after 30 days gets a bounded update of `1 → 2`, before any promotion floor.
- The spacing calculation remains finite for extremely old nodes and cannot fail through retrievability underflow.
- Growth diminishes as stability rises; approximately 74 closely spaced eligible updates are required to reach the default cap from `S=1`.
- Ten confirmed uses in one day produce one numeric stability update but record the latest confirmed-use time.
- The decay manager and reinforcement path use one centralized exponential retrievability implementation.
- Lifecycle-model versioning can represent more than migrated/not-migrated.
- Config validation, tests, and configuration documentation cover all new knobs/state.

## Related

- #191 — design decision
- Depends on the confirmed-use semantics issue split from #191

AndreLFSMartins:
Closes #221.

## The problem

Reinforcement was unbounded. Every recall multiplied `stability` with no ceiling on how often it could happen, so ten uses in one afternoon compounded ten times. A memory used in a single burst could reach the cap and stop decaying altogether — pinned in `working` forever, which is the opposite of what a forgetting curve is for.

## What changes

**A new pure module, `src/ormah/lifecycle.py`,** owns the formulas: retrievability, spacing, reinforced stability, and cooldown eligibility. No I/O, no DB, no settings object — callers pass knobs in. `memory_engine._touch_access` and `background/decay_manager.run_decay` become thin callers of it.

**Bounded, diminishing growth,** with the curve fixed by #191 and not tuned here:

```text
spacing = min(R^-0.2, fsrs_spacing_cap)
S'      = min(S * (1 + fsrs_growth_factor * S^-fsrs_growth_exponent * spacing),
              fsrs_max_stability)
```

Each step shrinks as stability rises. From `S = 1.0`, reaching `fsrs_max_stability` takes 74 eligible updates.

**A per-node cooldown.** `stability` and `last_review` move at most once per `fsrs_reinforcement_cooldown_days`; `access_count` and `last_accessed` still move on every event. Combined with the default 1-day cooldown, those 74 updates become 74 days of daily use — the AC asked for a bound, and the composition gives a stronger one.

The check-then-write pair is a TOCTOU, so `_touch_access` is now serialized. The regression test counts reinforcement calls rather than asserting final stability: all racing threads read `S = 1.0` and write `1.5`, so the final value cannot distinguish one bump from four. Remove the decorator and the test fails `assert 4 == 1`.

**One retrievability implementation, anchored on use.** `decay_manager` calls `lifecycle.retrievability` instead of its own `math.exp`, and both decay and the importance scorer now anchor on `last_accessed` rather than `last_review` — the cooldown can leave `last_review` a full window behind real use, and an actively used memory must not read as stale.

`importance_scorer.py` gets the anchor flip and nothing else. It keeps its own inline retrievability, because #222 rewrites that line and reading `r["stability"]` there afterwards would raise an `IndexError` the surrounding handler does not catch.

**Four new config knobs** (`fsrs_growth_factor` 0.5, `fsrs_growth_exponent` 0.5, `fsrs_spacing_cap` 2.0, `fsrs_reinforcement_cooldown_days` 1.0) replace `fsrs_stability_growth`, which was a base multiplier with different semantics. Validators reject non-finite values at the config boundary — NaN compares `False` against every `<=` check downstream and would otherwise be serialized straight into frontmatter.

**`meta.fsrs_migrated` becomes `meta.lifecycle_model_version`,** an integer, so a future curve migration can tell which model produced the stored values. Every fallback in reading it resolves to "already migrated": skipping a seed is inert, running one overwrites `stability` and rewrites the Markdown.

## Migration — what happens to an existing store

On first startup, `_migrate_fsrs` reads `meta`. An existing store carries `fsrs_migrated='1'`, so it maps to version 1, the seed does **not** run, and two `meta` rows are written. **No node is touched.** Existing `stability` values are not rescaled, per this issue's own constraint.

The user-visible consequence is documented rather than silently absorbed: a memory the old unbounded formula pushed to `fsrs_max_stability` stays there and needs roughly `-ln(fsrs_decay_threshold) × fsrs_max_stability` days of disuse — about 440 at the defaults — before it decays. Only future growth is bounded. Correcting existing values would need a rescaling migration, deliberately not done here.

### The empty-`meta` case

SQLite is derived state and `backup.py` excludes `index.db` from every backup, so a fresh-device restore arrives with no `meta` at all. Mapping that to version 0 would run the seed over stability the user actually earned.

The durable signal already existed and was simply never consulted: `last_review` lives in the Markdown frontmatter and is restored on rebuild. If any node carries it, the store is not pre-FSRS and the seed is skipped.

**This covers the empty-`meta` case only.** It does nothing for a restore onto an installation whose index is still present: `full_rebuild` preserves `meta` and `reload_restored_graph` never calls `_migrate_fsrs`. That path is a pre-existing defect on `main` today — `_migrate_fsrs` already returns early on `if fsrs_migrated: return` with the same result — and it is tracked separately as #236, not fixed here.

### Downgrade

**Downgrading below this lifecycle model after upgrading is not supported.** Both `lifecycle_model_version` and `fsrs_migrated` are written together, which stops a binary that predates the new key from *reseeding*. That is all it does. The old binary keeps writing `stability` with the unbounded formula while the version marker sits at the new value, and the next upgrade trusts a marker that no longer describes the data. Roll the binary back only together with a backup taken from before the upgrade.

## Review trail

Six task-scoped reviews (one per task, each on its own diff), then a whole-branch review, then an adversarial peer round. Two findings from that round are worth surfacing rather than burying:

**Cross-process races (medium).** The reinforcement lock is an in-process `RLock`. A CLI and a server sharing one memory directory are not excluded from each other, so the cooldown's check-then-write can still interleave across processes. This is not specific to this branch — the whole engine has intra-process-only exclusion — and it is filed as #238. The repository already contains `StoreLock` for exactly this, which makes the fix a matter of applying an existing primitive.

**The `last_review` probe can be defeated by hand-authored Markdown (high).** `store/markdown.py` parses `stability` and `last_review` independently, so a hand-written file carrying `stability: 30` with no `last_review` key produces meaningful stability with no durable signal. If such a store is restored with no index, and no node anywhere in it carries `last_review`, the seed runs and overwrites it.

Real, and narrower than it sounds: it needs a store where not one node has ever been reinforced or seeded. It is also **not a regression** — on `main` today the same empty-`meta` restore seeds unconditionally, destroying the data in every case, while this branch destroys it only in the never-recalled subset. The branch narrows a pre-existing defect without closing it. Appended to #236, whose shape (durable versioning, per-node migration) is what would close it properly.

## Testing

New: `tests/test_lifecycle.py`, `tests/test_config_fsrs.py`, `tests/test_engine/test_reinforcement_cooldown.py`, `tests/test_engine/test_lifecycle_model_version.py`, plus additions to the decay and importance suites.

The tests were checked for discrimination, not just for green: removing the serialization decorator fails the concurrency test; removing the `last_review` guard makes the migration test report `14.0` instead of `1.0`; reverting the decay hardening makes a naive timestamp abort the whole run again. Each of those was run.

Full suite on this branch: 1889–1891 passing. The 11–12 failures are pre-existing and environmental on the development machine — local configuration overriding defaults in `test_config`/`test_consolidator`/`test_session_watcher`, plus `test_setup` cases around the Codex MCP config and the fastembed cache. None touch lifecycle, decay, or stability. `ruff check src/ tests/` is clean.

## Merge order

This should land **after** #234 (#220) and #235 (#222). Without #220 the branch does not literally implement "confirmed use", and #222 rewrites the same `recency_signal` line touched here.

### DIFF
--- a/src/ormah/lifecycle.py
+++ b/src/ormah/lifecycle.py
@@ -1,0 +1,89 @@
+"""Centralized memory-lifecycle math (#221, decided in #191).
+
+The FSRS curve used to be written out in the engine and in every background job
+that needed it, so a curve change meant editing three call sites that could
+silently diverge. Everything numeric about retrievability, spacing, and
+reinforcement lives here instead: pure functions, no I/O, no settings object, no
+database. Callers pass the knobs in.
+"""
+
+from __future__ import annotations
+
+import math
+from datetime import datetime, timedelta
+
+# Shape of the spacing curve, fixed by #191. Deliberately not configurable — the
+# knobs are the bounds around it (spacing cap, growth factor/exponent).
+_SPACING_EXPONENT = 0.2
+
+
+def retrievability(
+    days_since: float, stability: float, *, fallback_stability: float = 1.0
+) -> float:
+    """``R = exp(-t / S)`` — the probability the memory is still retrievable.
+
+    ``Node.stability`` is ``Field(ge=0.0)``, so ``S = 0`` is a representable value
+    that would make the exponent infinite; it falls back to *fallback_stability*.
+    """
+    effective_stability = stability if stability else fallback_stability
+    return math.exp(-max(days_since, 0.0) / effective_stability)
+
+
+def spacing_factor(
+    days_since: float, stability: float, cap: float, *, fallback_stability: float = 1.0
+) -> float:
+    """``min(R^-0.2, cap)``, computed without ever materializing ``R``.
+
+    ``R`` underflows to ``0.0`` once ``t / S`` passes ~745, and ``0.0 ** -0.2``
+    raises ``ZeroDivisionError``. Since ``R^-0.2 == exp(0.2 * t / S)``, working on
+    the exponent keeps the result finite at any age, and returning the cap as soon
+    as the exponent reaches ``log(cap)`` keeps ``exp`` away from its own overflow.
+    """
+    effective_stability = stability if stability else fallback_stability
+    exponent = _SPACING_EXPONENT * max(days_since, 0.0) / effective_stability
+    if exponent >= math.log(cap):
+        return cap
+    return min(math.exp(exponent), cap)
+
+
+def reinforced_stability(
+    stability: float,
+    days_since: float,
+    *,
+    growth_factor: float,
+    growth_exponent: float,
+    spacing_cap: float,
+    max_stability: float,
+    initial_stability: float,
+) -> float:
+    """``S' = min(S * (1 + g * S^-w * spacing), max_stability)``.
+
+    Bounded and diminishing: the ``S^-w`` term shrinks each step as stability
+    rises, and the capped spacing factor stops a single very old node from
+    reaching the ceiling in one use.
+    """
+    effective_stability = stability if stability else initial_stability
+    spacing = spacing_factor(
+        days_since,
+        effective_stability,
+        spacing_cap,
+        fallback_stability=initial_stability,
+    )
+    grown = effective_stability * (
+        1 + growth_factor * effective_stability**-growth_exponent * spacing
+    )
+    return round(min(grown, max_stability), 2)
+
+
+def reinforcement_due(
+    last_review: datetime | None, now: datetime, cooldown_days: float
+) -> bool:
+    """True when a node's numeric stability update is off cooldown.
+
+    ``last_review`` tracks the last numeric update, which is distinct from
+    ``last_accessed``: use advances the decay anchor on every event, while the
+    stability number moves at most once per cooldown window.
+    """
+    if last_review is None:
+        return True
+    return (now - last_review) >= timedelta(days=cooldown_days)
--- a/src/ormah/lifecycle.py
+++ b/src/ormah/lifecycle.py
@@ -1,10 +1,13 @@
-"""Centralized memory-lifecycle math (#221, decided in #191).
+"""The single definition of Ormah's memory-lifecycle math.
 
-The FSRS curve used to be written out in the engine and in every background job
-that needed it, so a curve change meant editing three call sites that could
-silently diverge. Everything numeric about retrievability, spacing, and
-reinforcement lives here instead: pure functions, no I/O, no settings object, no
-database. Callers pass the knobs in.
+Every numeric rule about retrievability, spacing, and reinforcement lives here,
+so the engine and the background jobs share one curve rather than each carrying
+a copy that can drift. Changing how memory decays or strengthens means editing
+this module and nothing else.
+
+Pure functions only: no I/O, no settings object, no database. Callers own the
+state and pass the knobs in, which keeps the curve independently testable and
+free of any caller's lifecycle.
 """
 
 from __future__ import annotations
--- a/docs/01 - Data Model.md
+++ b/docs/01 - Data Model.md
@@ -30,13 +30,13 @@ connections: list[Connection]  # Outgoing edges to other nodes
 # Scoring
 confidence: float           # [0.0-1.0] How certain we are (default: 1.0)
 importance: float           # [0.0-1.0] Computed by importance_scorer job; see 05 - Background Jobs
-access_count: int           # Times this node has been accessed/recalled
+access_count: int           # Number of confirmed uses
 
 # Temporal
 created: datetime           # When first stored (UTC)
 updated: datetime           # Last modification (UTC)
-last_accessed: datetime     # Last confirmed use (UTC) — not a search hit
-last_review: datetime | None # Last FSRS review (spaced repetition)
+last_accessed: datetime     # Last confirmed use (UTC), not a search hit; the decay/spacing anchor
+last_review: datetime | None # Last numeric stability update; gated by the reinforcement cooldown, so it can lag last_accessed
 valid_until: datetime | None # Expiry date (set by mark_outdated)
 
 # FSRS (Spaced Repetition)
--- a/docs/05 - Background Jobs.md
+++ b/docs/05 - Background Jobs.md
@@ -124,7 +124,7 @@ It does not do embedding clustering or hierarchical clustering despite the name.
 ## Importance and Decay
 
 - `importance_scorer` computes a normalized multi-signal importance value
-- `decay_manager` uses FSRS-style retrievability alone to decide whether to demote working memories
+- `decay_manager` uses FSRS-style retrievability alone to decide whether to demote working memories, anchored on `last_accessed` (the last confirmed use) rather than on `last_review` (the last numeric stability update), which the reinforcement cooldown can leave a window behind
 
 Importance does not protect a node from decay. Cumulative signals such as access
 and edge counts could otherwise push a stale node permanently above any threshold.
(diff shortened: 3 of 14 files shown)


src/ormah/lifecycle.py
```diff
@@ -0,0 +1,89 @@
+"""Centralized memory-lifecycle math (#221, decided in #191).
```
r-spade: I think this is class ! Thanks for extracting and consolidating this in one place. 

A note : The doc string here contains information about how we came to this decision. Having this as the central location where one can change the lifecycle math without having to touch many places is great, but I think the doc string does not have to reflect the historical stages through this class evolved into its state. Guess the best thing is to state its purpose and reason for existence. 

Open to feedback on this :)

r-spade:
Thanks Andre — one clarification before the eventual rebase onto #235.

I read the two timestamps introduced by the cooldown as having separate responsibilities:

- `last_review`: when numeric stability was last increased, so it gates whether the cooldown has elapsed;
- `last_accessed`: the latest confirmed use, so it measures how difficult/spaced the next use actually was.

The current reinforcement path correctly checks cooldown eligibility with `last_review`, but then calculates `days_since` from `last_review or last_accessed` as well.

Concrete case: stability was increased 25 hours ago, the memory was confirmed-used again one hour ago inside the cooldown, and it is used now. The current code measures 25 hours of spacing and moves `S=1` to about `1.62`; measuring from the latest confirmed use gives about `1.50`. With a configured seven-day cooldown, the same distinction can become `1→2` versus `1→1.5`.

Is using the last numeric update as the spacing anchor intentional? My reading of #221 is that `last_review` should gate the numeric update, while `last_accessed` should supply `days_since` to the spacing term—otherwise recent uses inside the cooldown are ignored when estimating recall difficulty.

I reproduced this with a focused test. If that interpretation matches yours, we can add the regression while resolving the post-#235 rebase. If you intended the cooldown window to collapse all intervening uses into one daily review, could you share the reasoning so we preserve the intended model?

AndreLFSMartins:
Confirmed — your reading is correct, and it's a real bug, not intended behavior.

I reproduced your numbers directly against `reinforced_stability()` with the default knobs (`g=0.5`, `w=0.5`, `spacing_cap=2.0`, `S=1`):

- anchored on `last_review` (25h): `days_since≈1.0417` → `S≈1.61585` → rounds to `1.62`
- anchored on `last_accessed` (1h): `days_since≈0.04167` → `S≈1.50418` → rounds to `1.50`

Both match what you posted.

You're also right that this is distinct from the `last_accessed`-anchoring already done for decay/importance in this PR: `_touch_access` (the reinforcement path itself, `memory_engine.py`) still computed `days_since` from `node.last_review or node.last_accessed` — so a confirmed use that lands inside the cooldown window is invisible to the spacing calculation once the cooldown clears, exactly as your concrete case shows. `last_review` should only gate *whether* the numeric update runs (`reinforcement_due`); `last_accessed` is the actual last confirmed use and should be the spacing anchor.

Pushed a one-line fix (`anchor = node.last_accessed or node.last_review`) plus a regression test that seeds the two timestamps apart and asserts the anchor is `last_accessed` — it fails against the old order (`1.62`) and passes with the fix (`1.50`). None of the existing cooldown tests caught this because their `_backdate_review` helper always moved both timestamps together, so they never exercised the case where they diverge.

Will fold this in before the rebase onto #235.

src/ormah/lifecycle.py
AndreLFSMartins: Agreed — and it's a good distinction. The docstring was explaining the refactor rather than the module: where the curve used to live and why it moved is #191/#221 material, and it's already recorded there. Someone opening `lifecycle.py` needs to know what it is for and what it guarantees, not which change produced it.

Rewritten to state the two things that hold regardless of how we got here — this is the single place the curve is defined, and it holds pure functions only (no I/O, no settings, no database), so the math stays independently testable:

```
"""The single definition of Ormah's memory-lifecycle math.

Every numeric rule about retrievability, spacing, and reinforcement lives here,
so the engine and the background jobs share one curve rather than each carrying
a copy that can drift. Changing how memory decays or strengthens means editing
this module and nothing else.

Pure functions only: no I/O, no settings object, no database. Callers own the
state and pass the knobs in, which keeps the curve independently testable and
free of any caller's lifecycle.
"""
```

Pushed as `bee9191`. Happy to trim it further if you'd rather it were shorter.

r-spade:
> Confirmed — your reading is correct, and it's a real bug, not intended behavior.
> 
> I reproduced your numbers directly against `reinforced_stability()` with the default knobs (`g=0.5`, `w=0.5`, `spacing_cap=2.0`, `S=1`):
> 
> * anchored on `last_review` (25h): `days_since≈1.0417` → `S≈1.61585` → rounds to `1.62`
> * anchored on `last_accessed` (1h): `days_since≈0.04167` → `S≈1.50418` → rounds to `1.50`
> 
> Both match what you posted.
> 
> You're also right that this is distinct from the `last_accessed`-anchoring already done for decay/importance in this PR: `_touch_access` (the reinforcement path itself, `memory_engine.py`) still computed `days_since` from `node.last_review or node.last_accessed` — so a confirmed use that lands inside the cooldown window is invisible to the spacing calculation once the cooldown clears, exactly as your concrete case shows. `last_review` should only gate _whether_ the numeric update runs (`reinforcement_due`); `last_accessed` is the actual last confirmed use and should be the spacing anchor.
> 
> Pushed a one-line fix (`anchor = node.last_accessed or node.last_review`) plus a regression test that seeds the two timestamps apart and asserts the anchor is `last_accessed` — it fails against the old order (`1.62`) and passes with the fix (`1.50`). None of the existing cooldown tests caught this because their `_backdate_review` helper always moved both timestamps together, so they never exercised the case where they diverge.
> 
> Will fold this in before the rebase onto #235.

@AndreLFSMartins 
Thanks!
To make things a bit easier for you I just took on the rebasing after I finished reviewing the PR. I will wait for your confirmation before I merge this, so as to make sure I havent missed something.

Its approved from my side

r-spade (approved):

AndreLFSMartins:
Confirmed — thanks for taking the rebase.

I re-ran the full suite against `497dc1c`: **1949 passed**, and the only 3 failures (`tests/test_setup.py::TestConfigureCodexMcp`) reproduce identically on the base `a521eaf`, so they are environmental on my machine (a local `codex` binary on `PATH`), not from this branch. `ruff check src/ tests/` is clean.

I also audited your rename commit specifically:

- The anchor fix survived the rebase — `anchor = node.last_accessed or node.last_review` is still in `_record_confirmed_use`.
- The regression test still catches the bug: reverting the anchor to the old order makes `test_reinforcement_anchors_on_last_accessed_not_a_lagging_last_review` fail (1.62 instead of 1.50), and restoring it turns the file green again (7 passed).
- Your edit to the concurrency docstring is accurate. Both call sites reach `_record_confirmed_use` *after* `db.transaction()` has returned, so there really is no additional serialization on that path — the interleaving the test constructs is still the production shape.

Good to merge from my side. Resolving the `lifecycle.py` thread — the docstring rewrite landed in `f7d510d`.

AndreLFSMartins:
One heads-up before you hit merge: the PR shows `BLOCKED` even with your approval, all three checks green and no open review threads.

The `main` ruleset has `require_extra_approval_for_unattributed_changes: true`. Since you pushed the rebase, the most recent push carries commits authored by me — changes not attributed to the pusher — so the rule wants an approval from someone who isn't the pusher. Your own approval can't satisfy that, and mine can't either as the PR author.

Your admin bypass ("merge without waiting for requirements") should clear it. Flagging it only so it doesn't read as a stale check.

For the record, current state: `mergeable: MERGEABLE`, `reviewDecision: APPROVED`, `test`/`lint`/`desktop` all SUCCESS, 0 unresolved threads.

Issue #242, opened 2026-08-19
bug(feedback): legacy fallback can silently miss confirmed-use reinforcement


r-spade:
## Problem

`submit_feedback(..., whisper_log_id=None)` keeps a legacy compatibility fallback that resolves the newest `whisper_log` row for the node, regardless of whether that row was actually injected.

Confirmed-use claiming correctly requires `was_injected = 1`. If the newest row is a rejected candidate (`was_injected = 0`), positive feedback still records affinity/signals and returns `Feedback recorded...`, but no confirmed-use claim is taken and the node lifecycle does not advance.

## Concrete scenario

1. Memory A is injected on turn 3.
2. A is considered but rejected on turn 4, producing a newer `was_injected = 0` row.
3. A client later calls `submit_feedback(A, +1, source="implicit")` without the exact event ID.
4. The fallback selects the rejected turn-4 row.
5. Feedback evidence is stored and success is returned, but confirmed use is silently lost.

This was highlighted during an independent review of PR #234. That PR deliberately tests and accepts the fail-closed result, which is safer than reinforcing the wrong event, so it should not block #234. The open question is whether this fallback case is rare enough to leave silent.

## Root cause

The legacy fallback serves two different jobs with one selected row:

- choosing where affinity/signals evidence attaches;
- proving that the memory was surfaced and is therefore eligible for confirmed use.

The newest row is not necessarily eligible for both jobs because rejected candidates are logged on later prompts.

## Possible directions

- Require `whisper_log_id` for lifecycle confirmation and return an explicit “feedback recorded, use not confirmed” result when it is absent or resolves to a rejected event.
- Resolve evidence attachment and confirmed-use provenance separately, with carefully defined recency/session bounds.
- Deprecate the fallback after clients consistently propagate exact event IDs.

Any fix should remain fail-closed and must not strengthen a memory merely because an older injected row exists somewhere in its history.

## Related

- #220 / PR #234 — confirmed-use semantics and the deliberately accepted fallback loss

r-spade:
@AndreLFSMartins : Came across this while reviewing the PR 220, would like to know your thoughts on this

AndreLFSMartins:
Thanks for flagging it. Short answer to the open question — **no, it is not rare enough to leave silent, but not for the reason the issue suggests: it does not measurably bite today, and the failure mode that matters most is not the silent miss.**

[...]

Issue #201, opened 2026-08-09
perf(instructions): the 707-word always-on block duplicates the MCP schemas and the runtime signal — half of it is reference, not instruction


AndreLFSMartins:
## Summary

`src/ormah/instructions.md` is 707 words and sits in the user's global `CLAUDE.md`, so it is re-read on **every message of every project**, forever. Most of it is not instruction the agent acts on — it is reference material that either (a) already exists verbatim in the MCP tool schemas the agent reads at call time, (b) is emitted by the runtime at the exact moment it becomes relevant, or (c) restates a single instruction four times.

The three variants carry the same weight: `instructions.md` 707 words, `codex_instructions.md` 740, `pi_instructions.md` 725.

This is a request to restructure it around progressive disclosure — keep the steps that fire every turn resident, and let the rest be reached when it is actually needed.

## Measurement

Per-item word count of `instructions.md` (`wc -w` on the split blocks, at 0cb7fc7):

| Item | Words |
|---|---:|
| header + `## Guidelines` | 55 |
| 1. Proactively remember | 41 |
| 2. Remember at natural save points | 62 |
| 3. Notice what stands out | 127 |
| 4. Check before assuming | 26 |
| 5. Memory supports the flow | 125 |
| 6. Keep memories atomic | 40 |
| 7. Use appropriate tiers | 24 |
| 8. Avoid broad startup loading | 31 |
| 9. Mark outdated info | 23 |
| 10. Set confidence | 25 |
| 11. Run maintenance in the background | 33 |
| 12. Submit implicit feedback | 95 |
| **Total** | **707** |

Items 3 and 5 alone are 252 words — 36% of the block.

I did not have a tokenizer available, so I am deliberately not quoting a token figure; word counts above are exact, any token number would be an estimate.

### 1. The tool schemas already say it — and the copy has already drifted

The MCP schemas are authoritative and the agent reads them when it calls the tool. `instructions.md` caches them, and a cache of a lookup that costs nothing to perform earns nothing:

| instructions.md | `remember` / `submit_feedback` schema (verbatim) |
|---|---|
| 7. `core` = always-relevant, `working` = actively relevant, `archival` = historical | `tier`: "'core' = always loaded, 'working' = searchable, 'archival' = deep storage." |
| 10. "set `confidence` below 1.0… affects how prominently the memory appears" | `confidence`: "Belief strength 0.0-1.0. Lower values mean less certain." |
[...]

PR #247, opened 2026-08-19
docs(pi): align implicit-feedback guidance with the exact whisper_log_id contract

AndreLFSMartins:
Aligns `pi_instructions.md` item 12 with the post-#101 contract. Documentation only, one line.

## What was wrong

The Pi instructions told clients to call `ormah_submit_feedback(node_id=<id>, signal=1, source="implicit")` with no `whisper_log_id`, so **every** implicit feedback call from a Pi client resolves through the legacy latest-event fallback instead of attaching to the exact whisper event. `instructions.md` and `codex_instructions.md` have carried the corrected wording since #101.

## How it drifted

```
2026-06-25  PR #46 opens; pi_instructions.md branches from instructions.md as it
            stood that day, renamed for the ormah_-prefixed tools
2026-07-11  #101 merges, fixing item 12 in the two instruction files on main
2026-07-12  PR #46 merges, 32 hours later, adding a third file with the old wording
```

No conflict was possible — the file was new, so nothing it touched had changed. #101 could not have caught it and no reviewer had reason to look.

## Why it is doc-only

`tool_schemas.py:217` already declares `whisper_log_id` and `mcp_adapter.py:383-384` already forwards it when present. The parameter was reachable from the Pi path the whole time; Pi was simply never asked for it. No engine, schema, or behaviour change here.

## How the replacement was produced

Derived from `instructions.md` by the same `ormah_` prefix substitution the rest of the file already uses, rather than retyped, so the two cannot drift apart in this edit:

```
$ diff <(sed -n '29p' src/ormah/instructions.md | sed 's/`submit_feedback/`ormah_submit_feedback/g') \
       <(sed -n '31p' src/ormah/pi_instructions.md) && echo identical
identical
```

After this change the file differs from the canonical text by exactly two things, both intentional: the header line documenting the `ormah_` prefix, and item 11, whose mechanism (`/ormah:maintenance`) is genuinely Pi-specific.

## Scope

This fixes the instance. It does not address the structural cause — three hand-maintained copies of the same prose — which is the subject of #201, where the merge sequence above is written up as evidence. The runtime consequence is discussed in #242. This PR stands on its own and is independent of the direction decision on either.

Refs #242, #201


r-spade:
The wording is right — I reproduced your derivation and it's identical. But I don't think this is doc-only, and I'd hold it for one small code change.

**The evidence points at the wrong surface.** `tool_schemas.py` and `mcp_adapter.py` back the MCP tools, which are registered *unprefixed* (`submit_feedback`). The `ormah_` prefix is the Pi plugin only — which is what `pi_instructions.md:5` documents. So this file configures `integrations/pi-plugin/`, and that's where `whisper_log_id` has to be reachable.

**It isn't.** `tools.ts:246` declares three fields — `node_id`, `signal`, `source` — and `client.ts:202` posts exactly those three. Nothing forwards `whisper_log_id`.

Pi *does* receive the ID (the server bakes it into the whisper text, and `whisper.ts:58` passes it through). The inbound half works; it's the return trip that has nowhere to put it. Right diagnosis, one layer short.

**So after merging**, Pi is told to send a field its tool doesn't declare: either it's dropped and the legacy fallback still runs (goal not met, #242 still applies), or strict validation rejects the call (feedback stops entirely). No-op or regression, not neutral.

**Same 32-hour window as the prose.** `git log -S whisper_log_id` shows #101 (Jul 11) as the only commit touching those adapters; the Pi plugin was written Jun 24, before the mechanism existed, and merged Jul 12. Two artifacts, one gap — this closes one.

**To finish it:** add optional `whisper_log_id` to the `tools.ts` schema, thread it through `client.submitFeedback` conditionally (mirroring `mcp_adapter.py:381-382`), and add a test — `integrations/pi-plugin/tests/` has no `submit_feedback` coverage today, which is why this stayed invisible.

Fine as a follow-up if you'd rather keep this minimal, but I'd land the plumbing before the docs so the instructions never describe a parameter Pi can't accept.

r-spade:
@AndreLFSMartins Do you want me to take this on ? Or would you prefer to finish it ?

AndreLFSMartins:
You're right, and I'd rather you take it.

I pointed the evidence at the MCP adapters because that's where `whisper_log_id` actually lives. The `ormah_` prefix should have told me I was reading the Pi plugin instead of the MCP surface. It didn't. Good catch.

I checked what you described: `tools.ts` declares three fields, `client.ts` posts the same three. `grep -rn whisper_log_id integrations/pi-plugin/` comes back empty. So the docs would be telling Pi to send a parameter its own tool doesn't accept, and your ordering call is the right one. Plumbing first.

The honest reason I'm handing this over: I don't run Pi. I can write the schema change and thread it through `client.submitFeedback` blind, but I can't verify the tool actually accepts the field end to end, and that's exactly the gap that stayed invisible since July because nobody exercised it. A unit test I wrote against my own assumption would pass even if I got the shape wrong.

So it's yours. Cleanest path I see: take this branch, add the schema, the threading and the test on top, land both together, so there's never a window where the instructions describe a parameter Pi can't accept. If you'd rather start clean, close this one and lift the prose, it's three paragraphs.

Happy to review whatever you push.

r-spade:
@AndreLFSMartins Will take this on now. Cheers!

YagizEfeGokce:
Taken over — superseded by #262, which lands your prose fix together with the missing plugin plumbing (threading `whisper_log_id` through `ormah_submit_feedback` + first `submit_feedback` test coverage), so the instructions never describe a parameter the Pi tool can't accept. You're credited as co-author for the diagnosis and the verbatim derived wording. Thanks @AndreLFSMartins and @r-spade.

AndreLFSMartins:
Closing this one. The same change landed upstream through #262: `1db813b` (feat(pi): thread whisper_log_id through ormah_submit_feedback) already carries the exact text this PR proposes. I diffed both sides while syncing my tree down and they are byte-identical, same blob pair.

Nothing needed on your end, I am just tidying my own queue.

Issue #192, opened 2026-08-03
bug(background): consolidator summarizes from content[:300] — 85.5% of sources truncated before the LLM sees them, then demoted to archival


r-spade:
## What

`_consolidate_cluster` truncates every source memory to its first 300 characters before showing it to the LLM, then instructs that LLM to "preserve every concrete detail" and produce output that "becomes the PRIMARY representation of this knowledge."

`src/ormah/background/consolidator.py:213`:

```python
items.append(f"- [{title}]: {content[:300]}")
```

Anything past character 300 in a source memory is never presented to the model. It cannot be preserved, because it was never read. The prompt's rules — "Preserve every concrete detail" (line 228), "Preserve all reasoning" (line 230) — are addressed to a model that has been shown a fraction of the input.

Meanwhile `_apply_consolidation` (line 174) demotes every source to `archival`, so the truncated-away content is displaced from the whisper pool by a summary that was written without it.

## Measured impact

On a live 1,843-node store with 130 consolidation events:

```
consolidated originals:                                      297
  longer than 300 chars (truncated before the LLM saw them): 254  (85.5%)
  total characters never shown to the consolidator:      110,519
  worst single source:                  7,943 chars -> LLM saw 300 (3.8%)

median summary/sources length ratio:                        0.56
events where the summary is shorter than its sources:    102/130
```

85.5% of consolidated memories were truncated. The single worst case had 96% of its content withheld from the model that was tasked with representing it.

Note this is not the intended lossiness. Consolidation is *supposed* to shed redundancy — that is the feature. This sheds input the summarizer never saw, which is a different thing and is invisible in the output: the summary reads as complete because the model had no indication anything was missing.

## How it was found

While working through #191 (what "archival" should mean), specifically whether consolidated originals could safely be deleted rather than demoted. They cannot, and this is why — the summary is provably not a superset of its sources. The truncation turned a design question into a data-integrity question.

## Root cause

The 300-char cap appears to be copied from the candidate-generation pattern used elsewhere in `background/` (`conflict_detector.py` uses `[:500]` for pair screening, `_find_conflict_candidates` uses `[:400]`). Those are the right shape for a *screening* call — cheap, high-volume, deciding whether a pair is worth examining. Consolidation is a *generation* call: low-volume (`consolidation_max_clusters_per_run`, daily interval), and its output replaces its input in the whisper pool. It needs the full text, and the cost profile easily allows it.

## Possible fixes

[...]

PR #257, opened 2026-08-24
feat(lifecycle): reversible promotion and the seven-day initial lease (#223)

### LINKED ISSUE #223: feat(lifecycle): add reversible promotion and wire the seven-day initial lease
## Problem

Archival is currently a one-way transition: `TierManager.promote()` has no production callers. New nodes also take `stability=1.0` from the model default even though `fsrs_initial_stability` is configured and validated.

At the default threshold `R=0.3`, `S=1` crosses into decay candidacy after about 29 hours. An archival node deliberately recalled at low stability cannot currently return to the whisper-eligible working tier.

The proposed blanket exclusion for every `derived_from` target is also too broad. `derived_from` is a general relationship and does not necessarily mean that the target was superseded by consolidation.

## Decision from #191

- Set the default exponential initial stability to:

  ```text
  -7 / ln(0.3) ≈ 5.814
  ```

  This gives a mathematical seven-day unused working window.

- Use `fsrs_initial_stability` at node creation and as the promotion floor.
- Promote `archival → working` on confirmed use from `recall_node` or source-qualified positive feedback.
- Compute bounded reinforcement from the old stability first, then apply the floor:

  ```text
  bounded update: 1 → 2
  promotion floor: 2 → 5.814
  ```

- Make confirmed use, reinforcement/floor, and promotion one atomic lifecycle operation.
- Do not rescale existing stability values merely because the default is wired.
- Record explicit consolidation/supersession provenance and block automatic promotion only for those sources—not all `derived_from` targets.

## Acceptance criteria

- New working nodes use the configured initial stability and cross the default decay threshold at seven days, before scheduler cadence.
- Deliberate `recall_node` promotes exactly the requested archival node.
- Qualified positive feedback promotes its archival node; negative and unqualified sources do not.
- Promotion never reduces stability.
- The post-update floor does not amplify the same event into a longer-than-initial lease.
- Concurrent promotion/decay cannot leave markdown and index lifecycle state inconsistent.
- A generic `derived_from` target can promote; an explicitly superseded consolidation source cannot automatically promote.
- Migration, markdown serialization, database schema, tests, and docs cover the new provenance/state.

## Related

- #191 — design decision
- #192 — consolidation currently truncates source content before demoting originals
- #218 — strength calibration is separate; use the source-qualified rule meanwhile
- Depends on the confirmed-use and bounded-reinforcement issues split from #191

AndreLFSMartins:
## Problem

Archival was a one-way transition: `TierManager.promote()` had no production caller, so a node
deliberately recalled after archiving could never return to the whisper-eligible working tier.

New nodes also took `stability=1.0` from the model default, ignoring the configured and validated
`fsrs_initial_stability`. At the default threshold `R=0.3`, `S=1` crosses into decay candidacy
after about 29 hours.

And the exclusion originally proposed for promotion — every `derived_from` target — was too broad.
`derived_from` is a general relationship; it does not mean the target was superseded.

## Solution

Implements the decision recorded in #191.

### Seven-day initial lease

- `fsrs_initial_stability` default becomes `5.814` (`-7 / ln(0.3)`), a mathematical seven-day
  unused working window (`config.py`).
- `remember()` creates nodes at that configured stability instead of the model default
  (`memory_engine.py`).
- `lifecycle.promotion_floor(S, initial) -> max(S, initial)` — `max`, never a sum, so a node
  promoted twice in one cooldown window ends at one lease, not two, and a node already above
  the initial value is never pulled down to it.

### Reversible promotion

`_record_confirmed_use` promotes `archival -> working` on confirmed use, as one atomic lifecycle
operation with the bounded update:

- The bounded update runs on the **old** stability first, then the floor lifts the result
  (`1 -> 2 -> 5.814`), exactly as #191 specifies.
- The floor also runs when the per-day cooldown blocked the numeric update — otherwise a second
  confirmed use in one day promotes with `S=1` and buys a ~29 h lease the next decay run revokes.
- Promotion goes through `TierManager.promote()` rather than assigning `node.tier`, which gives
  the root cause its first production caller and brings the tier-ordering guard along. `updated`
  advances (the tier genuinely changed, and `updated` feeds LWW sync — not advancing it would let
  a stale remote copy silently re-archive the node).
- Tier and `updated` now travel in the same `UPDATE` as the rest of the access tracking, so
  markdown and index cannot disagree under concurrent promotion/decay. The `promote` audit entry
  is written after that transaction, where `update_node` places its own.

### Explicit supersession provenance

- New `superseded_by: str | None` on `MemoryNode`, with markdown (de)serialization, a
  `schema.sql` column, and an additive `ALTER TABLE` migration alongside the other enrichment
  columns; `IndexBuilder` persists it and it survives a reindex.
- The consolidator records supersession via `_mark_superseded` **before** demoting sources.
  Crashing between the two leaves a node working + marked, which is harmless because the marker
  only blocks automatic promotion. The reverse order would leave it archival + unmarked — a
  promotable node.
- `superseded_by` is written outside `UpdateNodeRequest` deliberately: it is policy state, not
  something an agent sets.
- Only an explicitly superseded consolidation source is blocked from automatic promotion. A
  generic `derived_from` target promotes — the narrowing #191 asked for.
- A **dangling** marker does not bury a node forever: the gate re-checks that the consolidation
  node still exists (`file_store.load`), so if the replacement was deleted (the #192 scenario)
  the next confirmed use un-buries the source. The marker itself is never cleared — it is the
  provenance record; only the block is lifted. The extra load runs only for an archival node
  that carries a marker.

## Notes for the reviewer

- **`fsrs_initial_stability` is also the `stability=0` fallback.** `decay_manager` passes it as
  `fallback_stability` and `reinforced_stability` uses it as `effective_stability` when the
  stored value is falsy. Changing the default therefore also moves the lease of `S=0` nodes from
  ~29 h to ~7 days. That is consistent with the intent, but it is a wider blast radius than the
  creation path alone.
- **Soft-deleting a consolidation node un-buries its sources.** Consequence of the dangling-marker
  self-heal above — a deliberate decision, not an oversight: burying a memory behind a replacement
  that no longer exists is the worse failure.
- **Consolidation is not idempotent under retry** — a failure after marking but before demotion
  leaves a working node carrying a marker, and the cluster query filters only on `tier='working'`,
  so a retry can produce a second consolidation. This is **pre-existing behaviour of the
  consolidator**, verified against the base (`git show <base>:src/ormah/background/consolidator.py`),
  not introduced here. Happy to open a follow-up issue for making the marking compare-and-set and
  excluding marked working nodes from new clusters.
- **`lifecycle_model_version` is deliberately not bumped.** The version records which
  reinforcement model wrote a store's stability values; this PR changes a creation default and
  adds a column, not the model. Reasoning is recorded in a comment at the version helper.
- **No stability rescale.** Existing values are left alone, as #191 requires.
- Product docs are intentionally out of this branch, following the precedent in `f8cb685`.

## Testing

10 commits, TDD throughout — 14 test files against 9 source files. New/extended coverage:

- `test_engine/test_reversible_promotion.py` — promotion on `recall_node`, qualified positive
  feedback, negative/unqualified sources not promoting, promotion never reducing stability, the
  floor not amplifying one event past a single lease, the `superseded_by` block, and the
  dangling-marker self-heal.
- `test_engine/test_recall_concurrency.py` — concurrent promotion/decay leaves markdown and index
  consistent.
- `test_index/test_migration_superseded.py`, `test_index/test_builder.py` — migration on an
  existing database and reindex survival.
- `test_store/test_markdown.py` — round-trip of the new front-matter field.
- `test_background/test_consolidator.py` — mark-before-demote ordering and provenance.
- `test_lifecycle.py`, `test_engine/test_reinforcement_cooldown.py`, `test_audit_log.py`,
  `test_tier_manager.py`, `test_decay_manager.py`, `test_lifecycle_model_version.py` — the floor
  helper, the cooldown interaction, the `promote` audit payload, and the unchanged version.

Closes #223.

### DIFF
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -163,7 +163,10 @@ def _apply_consolidation(
                         (new_id,),
                     )
 
-    # Create derived_from edges and demote originals to archival
+    # Create derived_from edges, record supersession, then demote originals to archival.
+    # Mark BEFORE demoting (#223): crashing between the two leaves the node working +
+    # marked, which is harmless because the marker only blocks automatic promotion.
+    # The reverse order would leave it archival + unmarked — a promotable node.
     for node_id in node_ids:
         try:
             engine.connect(ConnectRequest(
@@ -174,6 +177,7 @@ def _apply_consolidation(
             ))
         except Exception:
             pass
+        engine._mark_superseded(node_id, new_id)
         engine.update_node(node_id, UpdateNodeRequest(tier=Tier.archival))
 
     return new_id
--- a/src/ormah/config.py
+++ b/src/ormah/config.py
@@ -88,7 +88,7 @@ class Settings(BaseSettings):
     working_decay_days: int = 14  # Deprecated: superseded by FSRS-based decay
 
     # FSRS spaced repetition decay
-    fsrs_initial_stability: float = 1.0    # days; starting stability for new nodes
+    fsrs_initial_stability: float = 5.814   # days; -7 / ln(0.3) — a seven-day unused lease
     fsrs_decay_threshold: float = 0.3      # R below this = decay candidate
     fsrs_max_stability: float = 365.0      # cap at 1 year
     # Bounded reinforcement (#221/#191): S' = S * (1 + g * S^-w * spacing).
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -1560,6 +1560,32 @@ class MemoryEngine:
         ).fetchall()
         original_edges = [dict(r) for r in edge_rows]
 
+        # Second discovery route for the supersession redirect below, because the
+        # first one reads the index while the promotion gate reads the file, and the
+        # two can diverge: _mark_superseded saves the markdown first and writes the row
+        # second, so a crash in between leaves a marker only in the file. The
+        # consolidator writes the derived_from edge BEFORE calling _mark_superseded, so
+        # inside that window the edge is already there and names the source.
+        #
+        # derived_from on its own proves nothing — it is a general relationship, which
+        # is exactly why #223 narrowed the promotion gate off it. The marker in the file
+        # stays the criterion; the edge only says which files are worth opening, which
+        # keeps this bounded to the consolidation cluster instead of scanning the store.
+        # Read before the transaction: file I/O must not run holding the write lock.
+        marked_in_markdown: set[str] = set()
+        for edge in original_edges:
+            if edge["edge_type"] != EdgeType.derived_from.value:
+                continue
+            if edge["source_id"] != removed.id:
+                continue
+            # Full-id check: FileStore resolves ids through an eight-character filename
+            # prefix and can hand back an unrelated colliding node (#280).
+            candidate = self.file_store.load(edge["target_id"])
+            if candidate is None or candidate.id != edge["target_id"]:
+                continue
+            if candidate.superseded_by == removed.id:
+                marked_in_markdown.add(candidate.id)
+
         # Capture incoming edges for the kept node that aren't in its markdown.
         # index_single calls _remove_node which wipes ALL edges (including
         # incoming ones like self→kept "defines").  We need to restore these.
@@ -1637,18 +1663,26 @@ class MemoryEngine:
             # and after this merge the keeper is what represents them; left pointing at
             # the soft-deleted node it reads as dangling, and the next confirmed use
             # promotes a source back into the whisper-eligible tier (#223, PR #257).
+            #
+            # Union of both routes: the index, and the files named by the derived_from
+            # edges read above. A source found only by the second route has a row whose
+            # marker never landed, so the per-id UPDATE heals that row as it redirects.
             # The keeper is excluded — its own marker was already cleared above.
-            redirected_sources = [
-                r["id"]
-                for r in conn.execute(
-                    "SELECT id FROM nodes WHERE superseded_by = ? AND id != ?",
-                    (removed.id, kept.id),
-                ).fetchall()
-            ]
-            conn.execute(
-                "UPDATE nodes SET superseded_by = ? WHERE superseded_by = ? AND id != ?",
-                (kept.id, removed.id, kept.id),
+            redirected_sources = sorted(
+                {
+                    r["id"]
+                    for r in conn.execute(
+                        "SELECT id FROM nodes WHERE superseded_by = ?", (removed.id,)
+                    ).fetchall()
+                }
+                | marked_in_markdown
+                - {kept.id}
             )
+            for source_id in redirected_sources:
+                conn.execute(
+                    "UPDATE nodes SET superseded_by = ? WHERE id = ?",
+                    (kept.id, source_id),
+                )
 
             # Clean up auto-linker checked pairs:
             # - removed node: delete all (node is gone)
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -2098,11 +2098,21 @@ class MemoryEngine:
         # the next confirmed use un-buries it. The marker itself is NOT cleared —
         # it is the provenance record, and only the block is lifted. The extra load
         # is cheap and rare: it runs only for an archival node that carries a marker.
+        #
+        # The id is compared in full, not merely tested for None: FileStore resolves
+        # an id through the eight-character prefix in the filename and returns the
+        # first match without re-checking what it loaded (#280), so a deleted
+        # replacement whose prefix collides with an unrelated node comes back as that
+        # node. `is not None` would read that as live and bury this memory forever.
         promoted = False
         if node.tier is Tier.archival:
+            replacement = (
+                self.file_store.load(node.superseded_by)
+                if node.superseded_by is not None
+                else None
+            )
             superseded_by_live = (
-                node.superseded_by is not None
-                and self.file_store.load(node.superseded_by) is not None
+                replacement is not None and replacement.id == node.superseded_by
             )
             if not superseded_by_live:
                 node.stability = lifecycle.promotion_floor(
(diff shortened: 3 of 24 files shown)


AndreLFSMartins:
Follow-up issue for the idempotent-consolidation point in the review notes above: #258. Scoped to the pre-existing consolidator behaviour, so it stays out of this PR.

r-spade:
Thanks @AndreLFSMartins  — this is excellent work. The tests around exact-node promotion, cooldown/floor ordering, consolidation provenance and persistence are especially thoughtful.

The final review found two focused integration gaps that I think we should address before merging:

1. If consolidation creates A/B → C and C is later merged into D, `execute_merge()` soft-deletes C but does not redirect `A/B.superseded_by` from C to D. A later confirmed use therefore promotes A or B even though D still represents them. The merge path needs to update the affected Markdown sources and SQLite pointers.

2. The replacement liveness check should verify the complete ID:

   `replacement is not None and replacement.id == node.superseded_by`

   `FileStore.load()` currently trusts an eight-character filename prefix and can return an unrelated colliding node. I reproduced that independently and filed the system-wide root problem as #280.

These look contained, and I’m happy to add a small maintainer commit if that helps. Once they are covered, I’m comfortable approving and merging. Thanks again for carrying this lifecycle work through so carefully.

AndreLFSMartins:
Thanks — both points are covered, each in its own commit.

**1. `execute_merge` now redirects the sources' markers** (`68bca9b`)

Every node marked `superseded_by` the removed node is redirected onto the keeper, in both places that matter: the SQLite row (sync and reindex read it) and the markdown file (the promotion gate reads it). `updated` advances on those files — unlike `_mark_superseded`, nothing else in this path stamps it, and the marker feeds LWW sync, so a stale remote copy would otherwise win and restore the dead pointer.

One case the blind redirect gets wrong, so it is handled explicitly: if the **keeper itself** carries a marker pointing at the node it is absorbing, redirecting writes `kept.superseded_by == kept.id` — a marker that always resolves live, burying the keeper permanently. The keeper is excluded from the `UPDATE` and its own marker is cleared instead, before `save`/`index_single`, so markdown and the row agree from the start.

`undo_merge` is deliberately **not** covered, and I would rather state the reasoning than leave it implicit. Undo restores the removed node but leaves the keeper's merged content in place, so a source still pointing at the keeper is still genuinely represented by it — the block stays correct. Redirecting back would require persisting the redirect set in `merge_history`, i.e. a schema change, which felt like the wrong thing to introduce in a final review round. Happy to file it as a follow-up if you disagree.

**2. The liveness check confirms the complete id** (`f14fc1b`)

```python
replacement = self.file_store.load(node.superseded_by) if node.superseded_by is not None else None
superseded_by_live = replacement is not None and replacement.id == node.superseded_by
```

The same full-id guard is applied to the load inside the new merge redirect, for the same reason. I left the root cause to #280 — this is only the local defence.

**Tests** — five new, all failing before the fix:

- `tests/test_engine/test_supersession_merge_redirect.py` (new): the redirect reaches markdown and SQLite; a redirected source still blocks promotion (this is the one that fails loudest without the fix — the source promotes to `working`); a marker pointing elsewhere is left alone; the keeper never supersedes itself.
- `tests/test_engine/test_reversible_promotion.py::test_a_prefix_collision_is_not_mistaken_for_a_live_replacement`: an unrelated node sharing the first eight characters with a deleted replacement. It opens with a precondition assertion pinning that `load()` really does resolve the prefix to the collider — that assertion is the one to delete once #280 lands.

Full suite on the island: `1993 passed`, 4 pre-existing failures unrelated to this branch (`test_space_detect`, `TestConfigureCodexMcp`) — verified pre-existing by re-running them against `a825a57` with the change stashed.

AndreLFSMartins:
Follow-up to the above — while pinning where the redirect finds its sources, I found the same bug arriving by another door, and closed it in `ca9c849`.

**The redirect was reading the index; the promotion gate reads the file.** `_mark_superseded` saves the markdown first and writes the row second, so a crash between the two leaves a marker only in markdown. `SELECT id FROM nodes WHERE superseded_by = ?` cannot see that source, the merge misses it, and the next confirmed use promotes it back into the whisper-eligible tier — exactly the failure you reported, from a different cause.

The fix is cheap because `run_consolidation` writes the `derived_from` edge one line *before* calling `_mark_superseded`, so inside that window the edge already exists and names the source. The redirect now unions two routes: the index, and the files named by the derived_from edges leaving the removed node — edges `execute_merge` already loads for the remapping, so the cost is one file read per cluster member, not a store scan. The `UPDATE` moved from `WHERE superseded_by = ?` to per-id, so a source found only by the second route gets its row *healed* rather than skipped.

The edge only decides which files are worth opening — the marker in the file is still the whole criterion. `derived_from` on its own means nothing, which is exactly why #223 narrowed the gate off it.

**Twelve tests now, and I checked they are not vacuous:**

- Route 2 disabled (`| marked_in_markdown` → `| set()`): the three recovery tests fail, the other nine pass. So they test route 2 and nothing else.
- Marker check removed (`if candidate.superseded_by == removed.id` → `if True`): the two precision tests fail. This mutation is what caught a bad test of my own — `test_a_derived_from_target_marked_by_someone_else_is_not_redirected` built its edge *before* re-indexing the node, and `index_single` wipes every edge touching it, so the node was unreachable by route 2 and the test passed with the bug present. It now creates the edge last and asserts the edge exists as a precondition.

**One residual case stays open, pinned by `test_a_marker_with_neither_a_row_nor_an_edge_is_still_missed`.** A marker written into a file with neither a row nor an edge is still missed. Nothing in the repository produces that state — the consolidator is the only writer of markers and always writes the edge first — so closing it would mean a full store scan on every merge for a state no code can reach. If you would rather have the scan, say so and I will add it.

Full suite on the island: `2000 passed`, same 4 pre-existing environment failures.

r-spade:
Thanks Andre — I pushed `45b8218` on top to close one final crash-ordering window in the redirect.

SQLite was updated before C was deleted, but the source Markdown was redirected afterward. A crash between deletion and the Markdown write could therefore promote the source despite D still representing it. The commit deletes C last and adds a regression covering that interruption.

109 focused tests and Ruff passed, including after merging current `main`.

r-spade (approved):

Issue #259, opened 2026-08-25
bug(engine): Claude-in-the-loop consolidation summarizes from content[:400] — same defect as #192 on the maintenance route


AndreLFSMartins:
## What

The Claude-in-the-loop maintenance path has the same defect as #192, on a different route.

`MemoryEngine.get_maintenance_batches` (Phase 1 of the `run_maintenance` two-call protocol) hands consolidation clusters to the maintenance agent after normalizing every node to its first 400 characters — `src/ormah/engine/memory_engine.py:1828` (at `90c431e`):

```python
"content": (node.get("content") or "")[:400],
```

Phase 2, `apply_maintenance_results` (line 1902), then feeds the agent's consolidations to the **same** `_apply_consolidation` the background consolidator uses:

```python
for c in results.get("consolidations", []):
    try:
        _apply_consolidation(
            self,
            c["node_ids"],
            c["title"],
            c["content"],
            c.get("type", "fact"),
        )
```

So the sources are demoted to archival — displaced from the whisper pool — by a summary written from a 400-char view of each of them. Exactly the #192 failure: the model cannot preserve what it was never shown, and the output carries no scar, so the summary reads as complete.

## Why this is a separate issue from #192

The fix shape is different. #192's fix gives the background consolidator an explicit prompt budget and splits clusters that don't fit, because there the prompt goes through a provider window we must size and request (`num_ctx` for Ollama). Here the consumer is the Claude maintenance agent, whose context is orders of magnitude larger than any cluster — there is no window of ours to manage, and the `[:400]` protects nothing. The likely fix is simply passing full content for the consolidation batch (the link/conflict/merge batches are *screening* views, where truncation is the right shape — same distinction as #192's root-cause section).

Kept out of the #192 PR deliberately to keep that diff reviewable; this issue is the tracking for the second route.

## Measured context (from #192)

On a live store, 85.5% of consolidated originals exceeded 300 chars; the same distribution applies here — 400 chars does not change the picture (p50 of node content is ~360-390 chars, so roughly half of every cluster's content is still cut).

## Related

- #192 — same defect on the background consolidator route (fix in progress)

Issue #258, opened 2026-08-24
bug(background): consolidation is not idempotent under retry — a partial run creates a second consolidation node and rewrites supersession provenance


AndreLFSMartins:
## Problem

`_apply_consolidation` demotes its sources in a plain loop, with no atomicity and no record
that a given cluster was already consolidated. If the run dies partway through that loop — or
anywhere after the consolidation node is created — the store is left half-converted, and
nothing recovers it.

Verified against `upstream/main` (not a downstream branch):

```
$ git show upstream/main:src/ormah/background/consolidator.py
    # Create derived_from edges and demote originals to archival
    for node_id in node_ids:
        try:
            engine.connect(ConnectRequest(..., edge=EdgeType.derived_from, ...))
        except Exception:
            pass
        engine.update_node(node_id, UpdateNodeRequest(tier=Tier.archival))
```

and the cluster candidates are selected on tier alone:

```
"SELECT id, title, content, space FROM nodes WHERE tier = 'working'"
```

## Failure scenario

1. A cluster of sources `A`, `B`, `C` consolidates into `N1`.
2. `A` is demoted to archival. The process dies (crash, restart, exception in `update_node`)
   before `B` and `C` are demoted.
3. `N1` exists. `A` is archival. `B` and `C` are still `working`.
4. The next `run_consolidation` re-selects `B` and `C` — they match `tier = 'working'` — and
   consolidates them again into a second node `N2`.

The store now holds two consolidation nodes for what was one cluster, and `A`'s sources are
split across them. Nothing detects or repairs this.

## Why this is worth fixing now

This is pre-existing behaviour, not a regression — but #223 makes the damage more visible.
#223 adds `superseded_by` to record which consolidation node replaced a source, and marks
sources *before* demoting them (so a crash leaves a node `working` + marked, which is the safe
half-state rather than `archival` + unmarked). A retry, however, overwrites the existing
`superseded_by` with the new consolidation id, so the provenance of the first consolidation is
silently lost.

This was raised during the #223 PR review and deliberately left out of that PR's scope, since
the root cause predates it.

## Suggested direction

Two independent changes, either of which reduces the blast radius:

1. **Make the marking compare-and-set.** Preserve an existing *live* `superseded_by` instead of
   overwriting it, so a retry cannot rewrite provenance that already points at a surviving
   consolidation node.
2. **Exclude marked working nodes from new clusters.** The candidate query filters on
[...]

Issue #123, opened 2026-07-14
bug(index): reindexing a node deletes its incoming edges and cannot restore them — the connection lives in the source node's markdown


AndreLFSMartins:
Deferred follow-up from #117 (see the closing comment there).

`_remove_node` deletes every edge where the reindexed node is `source_id` **or** `target_id`. But the rebuild that follows only replays *that node's own markdown*, and an incoming edge `A -> B` is stored as a connection in **A's** file, not B's.

So reindexing B deletes the `A -> B` edge and has no way to bring it back: A's markdown is never read during B's rebuild. The edge is gone from the DB while A's file still claims it exists — the file store and the index silently disagree.

This is pre-existing (it predates #117; the diagnosis just surfaced it), but it is a real path to permanent edge loss, and it compounds with `auto_link_checked`: the pair is already marked checked, so the auto_linker will not re-evaluate it and re-create the edge.

PR #260, opened 2026-08-25
fix(consolidator): summarize from full source content, splitting oversized clusters instead of truncating

### LINKED ISSUE #192: bug(background): consolidator summarizes from content[:300] — 85.5% of sources truncated before the LLM sees them, then demoted to archival

AndreLFSMartins:
Fixes #192.

## The bug

`_consolidate_cluster` built its prompt with `content[:300]` per source, then
`_apply_consolidation` demoted every one of those sources to `archival`. The prompt itself
tells the model its output "becomes the PRIMARY representation of this knowledge" and that it
must "preserve every concrete detail" — instructions that are meaningless about text the model
was never shown, and destructive because the full sources stop being read the moment the
summary is written.

## The fix

1. **Full content, never a slice.** `_render_item` renders the whole `content`. `or ""` rather
   than `get(key, "")`, because `nodes.content` is nullable and a present-but-NULL value would
   otherwise render as the literal string `"None"`.
2. **Oversized clusters are split, not truncated.** `_split_cluster_to_fit` packs sources
   greedily in discovery order (seed first, then descending similarity) against
   `consolidation_max_prompt_chars` minus the template's own overhead, which is computed from
   the template so it cannot go stale. A source larger than the whole budget is **left where it
   is** rather than summarized from a partial view: losing a consolidation is recoverable, since
   the node stays `working` and keeps being surfaced; summarizing from a partial view is not.
3. **The per-run cap bounds LLM calls, not discovery.** Splitting can multiply one cluster into
   several, so the excess is deferred to the next run instead of silently costing more.
4. **Ollama gets an explicit input window.** `num_predict` bounds output only; leaving `num_ctx`
   unset inherits a server default we neither control nor version, and a default below the
   payload truncates the prompt silently — HTTP 200, plausible JSON. Only the consolidation
   route opts in; the shared maintenance adapter deliberately passes no `num_ctx`, so
   `auto_linker`, `conflict_detector` and `duplicate_merger` do not pay that KV cache for
   small pair judgements.
5. **Failure became observable.** A consolidation that produces no output now logs the source
   ids and the prompt size, and a successful one logs `source_chars` / `summary_chars`. Before,
   a run where every consolidation failed emitted nothing at all.

`consolidation_max_prompt_chars` defaults to `24000`, sized from the author's own store
(5,923 nodes / 301 real consolidation events): the worst real event builds a 12,961-char prompt
and the largest single node is 5,513 chars. At any value >= 16000 none of those 301 events would
have been split — **the split is a tail safety net, not the common path.** These numbers describe
one store's data, not a universal distribution.

## Known gaps — deliberately not fixed here

These were found while reviewing this branch and are limitations of *this* change, not
pre-existing upstream bugs. Flagging them rather than expanding the PR:

1. **The `litellm` adapter branch discards `num_ctx`.** A `litellm` provider pointing at an
   Ollama-backed model keeps the server's default window, so guarantee (4) above does not hold
   for that configuration. Closing it means either propagating the option through
   `LiteLLMAdapter` or refusing consolidation when the provider cannot promise the capacity.
2. **A character budget cannot bound tokens.** `_CHARS_PER_TOKEN = 2.0` asks for roughly twice
   the tokens Latin-script prose needs (~4 chars/token), but the margin only helps for content
   *less* dense than 2 chars/token. CJK runs ~1 char/token or worse, so a full 24000-char prompt
   would be ~24000 tokens against the 16096 requested. The real fix is tokenizer-based budgeting;
   the constant's comment now states this limit explicitly instead of claiming the opposite.
3. **`consolidation_max_prompt_chars` has a floor (4000) but no ceiling**, and it now sizes a KV
   cache allocation: `200000` asks for `num_ctx=104096`.

## Scope notes

- **#259 is the same defect on the other route.** The Claude-in-the-loop maintenance path
  summarizes from `content[:400]`. It is untouched here — one route per PR.
- **The E2E tests assert on `node_tags WHERE tag='consolidated'`, not on `derived_from` edges**,
  and the docstring says why: #123. `_apply_consolidation` creates the `derived_from` edge and
  the very next line calls `update_node(tier=archival)`, which reaches `IndexBuilder._remove_node`
  and its `DELETE FROM edges WHERE source_id = ? OR target_id = ?` — deleting the edge just
  created. Consolidated nodes on `main` therefore carry no provenance at all. That is #123's
  scope, not this PR's.
- **#258** (consolidation not idempotent under retry) is adjacent and also out of scope.
- **#261** was opened while reviewing this branch: a consolidated node stays `working` and
  re-clusters, so summaries get written from summaries. It is a pre-existing `main` bug that
  this PR neither causes nor fixes — discovery is untouched here.

## Verification

- `4 failed, 1976 passed`. The 4 failures live in `tests/test_setup.py` and
  `tests/test_adapters/test_space_detect.py` — files this branch does not touch — and are
  environment artifacts of running with a clean `HOME`.
- `ruff check src/ tests/` clean.
- Test power was checked by mutation rather than assumed. Two gaps it exposed are now covered by
  named tests: `test_a_cluster_landing_exactly_on_the_budget_stays_whole` pins the splitter's
  `>` / `>=` boundary (a cluster landing on exactly the budget, and one char over), and
  `test_consolidation_pins_the_llm_route` asserts `route == "consolidation"` at the call site —
  without it, deleting that argument left the whole suite green and turned the `num_ctx` commit
  into a no-op.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,3 +127,7 @@ ORMAH_WORKING_DECAY_DAYS=14
 
 # Memory consolidation (requires LLM)
 # ORMAH_CONSOLIDATION_INTERVAL_MINUTES=360
+# Prompt budget for one consolidation, in characters. Also sets the Ollama input window the
+# consolidation route requests (chars/2 + ORMAH_LLM_NUM_PREDICT). A cluster that does not fit is
+# split into smaller consolidations — sources are never truncated.
+# ORMAH_CONSOLIDATION_MAX_PROMPT_CHARS=24000
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -52,9 +52,14 @@ def _prompt_overhead_chars() -> int:
 
 
 def _render_item(node: dict) -> str:
-    """One source as it appears in the items block. FULL content -- never a slice (#192)."""
+    """One source as it appears in the items block. FULL content -- never a slice (#192).
+
+    ``or ""`` rather than ``get(key, "")``: ``nodes.content`` is nullable and the default only
+    applies to a MISSING key, so a present-but-NULL content would render as the literal string
+    "None" and teach the model that is the source's body. Same idiom as the audit log below.
+    """
     title = node.get("title") or "Untitled"
-    content = node.get("content", "")
+    content = node.get("content") or ""
     return f"- [{title}]: {content}"
 
 
@@ -363,6 +368,16 @@ def _consolidate_cluster(engine, cluster: list[dict]) -> None:
 
     raw = llm_generate(engine.settings, prompt, json_mode=True, route="consolidation")
     if raw is None:
+        # Without this the failure is invisible: llm_generate returns None on timeout, connect
+        # error or a disabled provider, the adapter's own warning names neither the job nor the
+        # cluster, and run_consolidation's closing report is guarded by `if consolidated_count:`.
+        # A run where every consolidation failed would emit nothing at all, so the daily job
+        # looks green while the working tier silently stops being curated.
+        logger.warning(
+            "consolidation produced no output for %d source(s) (prompt was %d chars); "
+            "sources left working: %s",
+            len(cluster), len(prompt), ",".join(str(n.get("id")) for n in cluster),
+        )
         return
 
     result = json.loads(extract_json(raw))
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -208,12 +208,17 @@ def _consolidate_cluster(engine, cluster: list[dict]) -> None:
     """Consolidate a single cluster using LLM summarization."""
     from ormah.background.llm_client import extract_json, llm_generate
 
-    # Build prompt
+    # Build prompt from FULL content, never a slice (#192). The prompt below tells the model its
+    # output "becomes the PRIMARY representation of this knowledge" and that it must "preserve
+    # every concrete detail" -- instructions that are meaningless about text the model was never
+    # shown, and destructive here because _apply_consolidation demotes every source to archival
+    # immediately afterwards. A cluster too large for the prompt is split upstream in
+    # run_consolidation; it is never trimmed.
     items = []
     for node in cluster:
         title = node.get("title") or "Untitled"
         content = node.get("content", "")
-        items.append(f"- [{title}]: {content[:300]}")
+        items.append(f"- [{title}]: {content}")
     items_text = "\n".join(items)
 
     prompt = f"""\
@@ -260,4 +265,13 @@ Return a JSON object:
         return
 
     node_ids = [n["id"] for n in cluster]
-    _apply_consolidation(engine, node_ids, title, summary, node_type)
+    new_id = _apply_consolidation(engine, node_ids, title, summary, node_type)
+    # Audit trail (#192): the sources are demoted to archival by the call above, so this summary
+    # becomes what gets read instead of them. Recording both sizes makes a consolidation that
+    # shed too much visible in the logs without re-measuring the store by hand.
+    logger.info(
+        "consolidated %d sources into %s: source_chars=%d summary_chars=%d sources=%s",
+        len(node_ids), new_id,
+        sum(len(n.get("content") or "") for n in cluster), len(summary),
+        ",".join(node_ids),
+    )
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -361,7 +361,7 @@ def _consolidate_cluster(engine, cluster: list[dict]) -> None:
 
     prompt = _CONSOLIDATE_PROMPT.format(items_text=items_text)
 
-    raw = llm_generate(engine.settings, prompt, json_mode=True)
+    raw = llm_generate(engine.settings, prompt, json_mode=True, route="consolidation")
     if raw is None:
         return
 
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -294,16 +294,57 @@ def run_consolidation(engine) -> None:
     if not clusters:
         return
 
-    consolidated_count = 0
+    # A cluster whose sources do not fit the prompt is SPLIT, never truncated (#192).
+    budget = settings.consolidation_max_prompt_chars - _prompt_overhead_chars()
+    if budget <= 0:
+        # Checked ONCE per run, not once per cluster: _split_cluster_to_fit would emit the same
+        # line for every one of the (up to consolidation_max_clusters_per_run) clusters, turning
+        # a single operator misconfiguration into a wall of identical WARNINGs.
+        logger.warning(
+            "consolidation prompt budget is %d chars once the template's own overhead is "
+            "subtracted from ORMAH_CONSOLIDATION_MAX_PROMPT_CHARS (%d); nothing can be "
+            "consolidated this run -- raise the setting",
+            budget, settings.consolidation_max_prompt_chars,
+        )
+        return
+
+    min_size = settings.consolidation_min_cluster_size
+    queue: list[list[dict]] = []
+    dropped_nodes = 0
+
     for cluster in clusters:
+        parts = _split_cluster_to_fit(cluster, budget)
+        kept = [p for p in parts if len(p) >= min_size]
+        dropped_nodes += len(cluster) - sum(len(p) for p in kept)
+        queue.extend(kept)
+
+    # The cap bounds LLM CALLS, not discovery. Splitting can multiply one cluster into several,
+    # and a daily job silently costing 2.5x more is not what this setting promises. The excess is
+    # simply not processed, so it is rediscovered next run.
+    capped = queue[: settings.consolidation_max_clusters_per_run]
+    if len(queue) > len(capped):
+        logger.info(
+            "consolidation queue held %d sub-cluster(s) over the per-run cap; deferring to the "
+            "next run", len(queue) - len(capped),
+        )
+    if dropped_nodes:
+        logger.info(
+            "%d source(s) left working: too large for the prompt budget, or alone in a "
+            "sub-cluster after the split", dropped_nodes,
+        )
+
+    consolidated_count = 0
+    for sub in capped:
         try:
-            _consolidate_cluster(engine, cluster)
+            _consolidate_cluster(engine, sub)
             consolidated_count += 1
         except Exception as e:
             logger.warning("Failed to consolidate cluster: %s", e)
 
     if consolidated_count:
-        logger.info("Consolidated %d cluster(s)", consolidated_count)
+        logger.info(
+            "Consolidated %d sub-cluster(s) from %d cluster(s)", consolidated_count, len(clusters)
+        )
 
 
 def _consolidate_cluster(engine, cluster: list[dict]) -> None:
--- a/src/ormah/background/consolidator.py
+++ b/src/ormah/background/consolidator.py
@@ -9,6 +9,108 @@ from ormah.background.memory_lock import serialized_memory_job
 
 logger = logging.getLogger(__name__)
 
+_CONSOLIDATE_PROMPT = """\
+You are consolidating a cluster of semantically similar memories into a single, richer memory. The consolidated memory will be stored as a new working-tier node, and the originals will be demoted to archival (still searchable but deprioritized). This means your output becomes the PRIMARY representation of this knowledge — it must be complete.
+
+Memories to consolidate:
+{items_text}
+
+## Your task
+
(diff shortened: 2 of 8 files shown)


r-spade:
Thanks Andre — one thing I think we need to change in the split behaviour.

If a cluster like `[Seed, A, B, C]` is too large, it is safe to consolidate a smaller group such as `[Seed, A]`.

But `B` and `C` should stay working for a later pass. They were each selected because they are similar to `Seed`; Ormah has not proved that `B` and `C` are related to each other.

The current split can consolidate `[B, C]`, create a summary, and archive both memories. That could combine two unrelated memories.

Could we only consolidate one smaller group that includes the seed, and leave everything else working for rediscovery? A regression test for this would be useful too.

r-spade:
One more safety gap: a character budget does not guarantee that the model can fit the prompt.

A 24,000-character CJK or token-dense source can require far more tokens than the 16,096-token Ollama window requested here. Ollama can silently trim the prompt, return a plausible summary, and then the full sources are archived.

Could we use the configured model’s tokenizer, or fail closed when full capacity cannot be guaranteed? This needs a regression with token-dense input.

r-spade:
The same capacity guarantee does not currently reach the LiteLLM path.

`num_ctx` reaches direct Ollama, but LiteLLM drops it. A supported `llm_provider=litellm` setup pointing at Ollama can therefore use the server default, silently truncate a full source prompt, and archive the originals.

Could we propagate the context setting for Ollama-backed LiteLLM models, or skip oversized consolidation when the provider cannot guarantee capacity? A mocked LiteLLM regression would pin this.

r-spade:
Small observability issue: `run_consolidation` increments its success count even when the model returns no output or an empty summary.

That can log a successful consolidation even though no new memory was created and no source changed.

Could `_consolidate_cluster` return whether it actually applied a consolidation, and count only true results?

r-spade:
Small prompt-budget boundary: `_item_chars` charges one newline for every source, but the actual joined prompt has only `n - 1` separators.

A cluster that exactly fits can be unnecessarily split, and the resulting smaller groups may be skipped.

Could we measure the rendered joined block directly, or charge separators only between items?

r-spade:
Finally, this branch now conflicts with current `main` in the consolidator tests.

Could you rebase once the fixes are in, preserving the #257 consolidation provenance/lifecycle changes, then rerun fresh CI?

AndreLFSMartins:
Thanks for the thorough pass. I went after all six before touching any code, and I landed somewhere slightly different on three of them. Short version: 4, 5 and 6 I'll just fix. 2 and 3 are real, but the trigger isn't what either of us assumed. On 1 I agree with the diagnosis and I'd like to push back on the fix, because I think it starves.

On how I checked, since the method matters for how much you should trust the numbers: I ran each point as its own investigation against the PR HEAD, then ran a second, adversarial round whose only job was to knock the first round down. The first round came back 5-for-5 "confirmed" and two of those were wrong in the details. The second round is what caught it. Everything below is the second round's numbers, all measured on my machine against this branch.

## 4. Success count

You're right, and I'll fix it the way you described: `_consolidate_cluster` returns whether it actually applied, and only true results count.

One note for the record. This one predates the PR. At the merge base (`90c431e`), `run_consolidation` already did `_consolidate_cluster(cluster)` followed by an unconditional `consolidated_count += 1`, and both silent returns were already in place. This branch actually improved half of it by adding the WARNING on the `raw is None` path. The `not summary` path is still completely silent, which is the one that worries me more. I'll fix it here since we're already in the file, but it isn't something this PR introduced.

## 5. `_item_chars` off-by-one

Right, and it fires more often than I would have guessed. The delta is a constant 1 char, not n, because it doesn't accumulate: at every prefix k the modeled cost sits exactly 1 above `len("\n".join(...))`. I checked k=1..20 with random sizes.

But the budget gets compared inside the greedy loop, once per item, so every item is a fresh chance for that 1 char to flip the decision. Monte Carlo over 20,000 random clusters (n=1..8, budget uniform 20-6000): 37 splits change, about 1 in 540. Probing near item boundaries specifically: 2064 of 10005, about 20%. Direction is the safe one, it charges too much and never overflows, so nothing here can blow the window. It just splits early and then `kept = [p for p in parts if len(p) >= min_size]` can drop the remainder, exactly as you said.

Also worth flagging: `test_a_cluster_landing_exactly_on_the_budget_stays_whole` derives its own budget from `_item_chars`, so it's self-referential. I reran its body verbatim with the `+1` removed and it still passes. Fixing this doesn't break it, which also means it was never really testing the boundary.

## 2. Token budget

The hole is real, and the failure mode is nastier than a trim. But CJK isn't the example that proves it, and I want to get that on the record before we design around it.

Measured with the model's own tokenizer, slope-differenced across two prompt lengths with a nonce to defeat prefix caching, cross-checked against a second tokenizer family:

| content | chars/token (Gemma3) | chars/token (XLM-R) |
|---|---|---|
| English prose | 5.69 | 4.98 |
| Chinese | 1.76 | 1.66 |
| minified JSON | 1.75 | |
| emoji | 0.99 | |
| hex/hash blob | 1.11 | |
| base64 | 1.43 | |

So 24,000 chars of natural CJK come out around 13,600 tokens and fit inside the 16,096 we ask for. What actually overflows is emoji (24,022 tokens, 1.49x), hash blobs (21,623, 1.34x) and base64, marginally. The 2.0 assumption is still wrong, it's just wrong for different content than we thought.

The truncation itself is worse than "silently trims", and this is the part I'd build the fix around. Ollama 0.33.2 discards deterministically: `prompt_eval_count = num_ctx/2 + num_keep - 1`, verified at five window sizes, and it's llama.cpp's context-shift discard rather than anything mysterious. It throws away the head and the middle and keeps the tail. Needle test with markers at start, middle and end of an 81k-char prompt: full window returns all three, `num_ctx=4096` returns only the last one. HTTP 200, `error: null`, `done_reason: stop` every time. `num_predict` buys no margin at all, I tried 1, 512 and 4096 against the same window and got identical counts.

On your fix suggestion: `/api/tokenize` is a 404 on 0.33.2 and I have no matching tokenizer cached locally, so a pre-flight count in the way you probably meant isn't available. The cheaper route is that `prompt_eval_count` already comes back in the very response `ollama_adapter.py:75` throws away. Plumb it out, compare against `num_ctx`, and refuse to archive when it doesn't match. Zero extra cost, exact model tokenizer, works offline. I'd rather do that than ship a chars-per-token constant tuned for a case I can't measure everywhere.

## 3. LiteLLM

Confirmed, and I got a bit further than reading docs. litellm isn't installed in this worktree, but there were eleven unpacked wheels sitting in the uv cache, so I built a scratch venv and actually ran 1.91.4 against a mock transport.

`num_ctx` does reach Ollama's `options` for both `ollama/` and `ollama_chat/` model strings, so the propagation you're asking for genuinely works. For a non-Ollama model it doesn't raise and it doesn't get dropped: it lands in `extra_body` and goes out on the wire (I dumped the body: `{"messages":..., "model":"gpt-4o", "num_ctx":8192}`). OpenAI rejects unknown top-level params, and `litellm.drop_params` doesn't help because it governs OpenAI-spec params, not passthrough kwargs.

Which matters because `setup.py` makes litellm the provider for four of six wizard options, including the one labelled "(recommended)". Three of those are Anthropic/OpenAI/Gemini with 128k-1M windows, where dropping `num_ctx` is the correct behaviour. So I'd gate the propagation on the model string resolving to an Ollama provider, rather than propagating unconditionally.

I'd skip your second option (refuse oversized consolidation when the provider can't guarantee capacity). That turns the job off for every litellm user, including the three where capacity was never in question.

One more data point on ranking, not on whether to fix: measured Ollama default here is 32,768 (`OLLAMA_CONTEXT_LENGTH`), and a 16,387-token prompt came back untruncated. The 2048/4096 default is a couple of versions back. The gap is real, the precondition is narrower than it looks.

## 1. The split

No argument on the diagnosis. I read the whole path again and there is nothing pairwise anywhere: `vec_store.search` is kNN against the seed vector, line 182 is the only similarity test and it's always against the seed, and the only filter after the split is `len(p) >= min_size`. For two members that both clear threshold t, the worst case for `sim(B,C)` is `2t²-1`, which at my configured 0.52 is -0.46. No floor at all.

I also measured how often it can happen in my own store: 489 clusters, worst one costs 9,510 chars against a 21,562 budget, zero of them split. Biggest working node is 3,439 chars and the five biggest together come to 16,381, so with `max_cluster_nodes=5` the split can't currently fire for me at all. That's a note on priority, not a defence. Nothing in the model caps content length, so it's reachable.

Where I'd like to push back is the fix. "Only consolidate one smaller group that includes the seed" breaks on the case where the seed fits alone but seed+smallest-match doesn't: `parts[0]` comes out as `[seed]`, `min_size=2` filters it, and nothing consolidates. Next cycle the same seed claims the same matches and drops them again, and the working-tier scan has no `ORDER BY`, so it's stable. That repeats forever. Right now `parts[1]` at least makes progress.

That's also #268, which I opened and is still open, and its suggested fix points the other way: reject the seed alone and let the eligible matches form a cluster of their own on the same pass. Related case I hadn't written down there: when the seed alone exceeds the entire budget, `parts[0]` already comes out without the seed, so the seedless group isn't only a tail effect of the greedy loop.

My current guess at the rule is something like: keep one group, it must contain the seed, and if that group can't reach `min_size`, drop the seed and let the matches cluster among themselves. But I'm honestly not sure that's right, and it feels like it belongs to #268 as much as to this PR. If you'd rather make it a short design discussion, I'm up for that, and I'd hold the rebase until it's settled.

## 6. Rebase

Will do, last. The conflict is only in `test_consolidator.py` and I'll preserve the #257 provenance/lifecycle changes, but rebasing before 1 is settled means doing it twice.

So: tell me where you want to land on 1 and I'll start on the rest.

PR #262, opened 2026-08-25
feat(pi): thread whisper_log_id through ormah_submit_feedback

YagizEfeGokce:
Picking this one up as requested — @AndreLFSMartins's #247 diagnosed the doc drift correctly one layer short; this PR lands **prose + plumbing together** so the Pi instructions never describe a parameter the Pi tool can't accept.

## What's broken

Post-#101, implicit feedback must include `whisper_log_id` to attach to the exact whisper/recall event. In the Pi plugin the **inbound half works** (server bakes the ID into whisper text, `whisper.ts` passes it through), but the **return trip had nowhere to put it**: `tools.ts` declared only `node_id`/`signal`/`source`, and `client.ts` posted the same three. Every Pi feedback call silently resolved through the legacy latest-event fallback (#242).

Same 32-hour merge window as the prose: the plugin was written Jun 24, before #101's mechanism existed, and merged Jul 12.

## Changes

- **`integrations/pi-plugin/src/tools.ts`** — optional `whisper_log_id` integer field on `ormah_submit_feedback` (description mirror of upstream `tool_schemas.py:207-213`); tool description now explains exact-event attribution, as upstream's does.
- **`integrations/pi-plugin/src/client.ts`** — `submitFeedback` gains optional `whisperLogId`; forwarded **only when present**, mirroring `mcp_adapter.py`'s conditional forward. Omitted → key absent from the POST body → legacy fallback behavior unchanged for older outputs.
- **`src/ormah/pi_instructions.md`** — item 12 updated to the post-#101 wording with the `ormah_` prefix (Andre's verbatim derivation from `instructions.md:29`, so no drift).
- **`integrations/pi-plugin/tests/client-feedback.test.ts`** — the first `submit_feedback` coverage for the Pi plugin (this gap is why the omission stayed invisible). Pins both body shapes: `whisper_log_id` present in the POST body when given, key absent entirely when not.
- **`integrations/pi-plugin/tsconfig.json`** — `allowImportingTsExtensions` so test files can import `../src/*.ts` under tsx + tsc.

## Validation

- `npm test` (tsx): **21/21 pass** (17 pre-existing + 4 new).
- `tsc --noEmit`: clean.
- `pytest`: 1949 passed; the 3 `test_config.py` failures reproduce on a clean `upstream/main` worktree (local `~/.config/ormah/.env` leaks provider config), unrelated to this change.
- **End-to-end against a live server:** `POST /agent/feedback` accepted with and without `whisper_log_id`; verified the ID actually lands — affinity row stores `whisper_log_id=1827` after the exact-attribution call.

## Notes

- Superseeding #247 as agreed — @AndreLFSMartins is credited as co-author for the prose + diagnosis. Closes this instance; the structural three-copies-of-same-prose cause remains #201.
- Release note for after merge: the plugin change only reaches users once `ormah-pi` is published — suggest bumping to `0.1.1` and `npm publish` so pi.dev/packages re-indexes.

Refs #242, #201


r-spade (approved):
Nice work — this is the right fix. Mirroring the conditional forward from `mcp_adapter.py` rather than just bolting the field on was the correct instinct, and the test pinning the key as absent rather than null is the one that matters. Approving.

One question, not blocking. The `tsconfig.json` change is only needed because the test imports `../src/client.ts` — the other tests in that folder use `.js` (`../src/config.js`). I swapped those three lines locally and everything passes without touching tsconfig. It's harmless as written since `noEmit` is on; I'd just rather not loosen a project-wide setting for one file. Any reason you went with `.ts`?

On the publish: `package.json` is already 0.1.1 and npm has that published, so this one needs 0.1.2. You've shipped both releases so far — do you want to take this one as well, or would you rather hand it over? Genuinely happy either way, I just don't want to step on it if you're already set up for it.

Separately, CI doesn't actually run the plugin tests — `ci.yml` runs `npm test` in `ui/`, so the green check here didn't include yours. That one's on us; I'll open an issue.

r-spade:
Pushed the import tweak myself as 05ad324 rather than leave this sitting — three imports to `.js`, tsconfig line dropped. Purely consistency with the other tests in `tests/`; nothing was wrong with what you had. Typecheck clean, 21/21 pass.

Before merging I also closed the coverage gap I mentioned and checked the wiring properly: the tool hands `whisper_log_id` to the client as the 4th arg, omitting it yields a genuine `undefined` (worth knowing — `null` there would have posted `whisper_log_id: null` straight through the `!== undefined` guard), and against a live server the affinity row came back with `whisper_log_id` stored. Works end to end.

Taking the 0.1.2 publish this round so it isn't waiting on you — say the word if you'd rather keep doing them and I'll hand it back.

One thing you might hit later: `npm install` in the plugin doesn't pull the peer deps (`typebox`, the two `@earendil-works` packages), so anything importing `src/tools.ts` fails to resolve until they're installed by hand. Probably why coverage stopped at `client.ts`. I'll fold that into the CI issue.

Thanks for this one — the diagnosis and the fix were both right.

r-spade:
`ormah-pi@0.1.2` is on npm — this is live now. `latest` points at it, 15 files, shasum `80d88b82`.

Thanks again for the fix.

Issue #272, opened 2026-08-26
bug(feedback): the heuristic usage detector never claims confirmed use — 0 of 1,629 positive whisper references reinforce a memory


AndreLFSMartins:
## Summary

The heuristic usage detector in `session_watcher` records a positive signal and an affinity row,
but **never claims confirmed use**. Only the LLM-judge branch calls `_claim_confirmed_use`. And
because a whisper the heuristic marked as referenced is *excluded* from the judge, the most
confident detector is the only one whose positive verdict cannot advance a memory's lifecycle —
and it also blocks the one path that would.

Measured on a live store: **0 of 1,629** positive heuristic pairs took a confirmed-use claim.

## Root cause

Two lines, both in `src/ormah/background/session_watcher.py` at `upstream/main` (90c431e).

**1. The heuristic commit block writes the signal and the affinity row, then stops** (L499-521).
There is no `_claim_confirmed_use` call anywhere in it:

```python
for record in heuristic_records:
    row = record["row"]
    recorded += _insert_usage_signal(
        conn, row, transcript,
        signal_type=record["signal_type"],
        polarity=record["polarity"],
        strength=record["strength"],
        source=_HEURISTIC_SOURCE,
        evidence=record["evidence"],
        created=now_iso,
    )
    if record["polarity"] == 1:
        _insert_affinity(
            conn, row, signal=1,
            source=_HEURISTIC_AFFINITY_SOURCE,
            confirmed_at=now_iso,
        )
```

The judge block does take the claim (L591-598), with the `#220` at-most-once comment explaining
why the claim — not a polarity check — is the right gate:

```python
if engine._claim_confirmed_use(
    conn, row["id"], row["node_id"],
    signal=record["polarity"],
    source=_LLM_JUDGE_AFFINITY_SOURCE,
):
    confirmed_node_ids.append(row["node_id"])
```

**2. A heuristic hit suppresses the judge** (L495):

```python
if llm_judge_enabled and not has_llm_judge and not referenced:
    llm_groups.setdefault((prompt_text, response), []).append(row)
```

`not referenced` means: when the heuristic *does* find the memory referenced in the assistant's
response — a node_id, title, sentence or token-overlap match — that whisper never reaches the
judge. So the positive path is: heuristic says "used" -> signal + affinity written -> judge
skipped -> no claim -> `_record_confirmed_use` never runs.

Net effect: `whisper_referenced` is a dead-end for the lifecycle. The only positive verdicts that
[...]

PR #273, opened 2026-08-26
fix(signals): signals.strength becomes an ordinal evidence scale (#218)

### LINKED ISSUE #218: bug(feedback): signals.strength has no variance in any channel — a tier enum stored as a float, blocking strength-based promotion gating
## What

`signals.strength` is documented and shaped as a per-event confidence score. In **two of its three producing channels it is a constant** — a tier label stored as a float. The third channel varies, which makes a cross-channel threshold not merely useless but actively inverted.

Every strength value ormah can currently emit:

| channel | value | why |
|---|---|---|
| `submit_feedback` (all sources: explicit, implicit) | **always 1.0** | hardcoded literal |
| session watcher — `node_id` match | always 1.00 | fixed return |
| session watcher — `title` match | always 0.95 | fixed return |
| session watcher — `sentence` match | always 0.90 | fixed return |
| session watcher — `token_overlap` match | **always 0.85** | formula saturates before its own entry gate |
| session watcher — `auto_llm_judge` | **varies** — the model's own confidence | `strength=confidence`, `session_watcher.py:549` |

### 1. `submit_feedback` hardcodes strength

`src/ormah/engine/memory_engine.py:2590` — the `signals` insert declares `strength` in its column list at `:2579` and then passes a literal:

```python
VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, datetime('now'))
"""
(
    whisper_log_id,
    resolved_node_id,
    "feedback_submitted",
    signal,
    1.0,               # <-- strength, hardcoded
    source,
    ...
```

So an explicit user confirmation and an implicit agent self-assessment are stored with identical strength. The `source` column is the only thing that distinguishes them.

### 2. `token_overlap` saturates before it can qualify

`src/ormah/background/session_watcher.py:146-147`:

```python
if len(overlap) >= 4 and overlap_ratio >= 0.5:
    return True, min(0.85, 0.45 + overlap_ratio), {...}
```

The entry gate requires `overlap_ratio >= 0.5`. The `min()` saturates at `overlap_ratio >= 0.40`. **Every match that reaches this branch is already at the ceiling.**

```
overlap_ratio 0.50 -> 0.45 + 0.50 = 0.95 -> min(0.85, ...) = 0.85
overlap_ratio 0.60 -> 1.05 -> 0.85
overlap_ratio 1.80 -> 2.25 -> 0.85
```

A barely-passing match and a ratio-1.8 match report identically.

## Measured

@AndreLFSMartins measured the distribution on a 36,729-node store with the session watcher enabled ([#191 comment](https://github.com/r-spade/ormah/issues/191#issuecomment-5268022245)):

```
node_id         1.0        30
title           0.95        2
sentence        0.9        13
token_overlap   0.85     1,520
```

97% of events land in the one tier that is pinned at its ceiling, and all 1,520 carry the identical value.

## Why it matters

This blocked a design decision in #191. The proposal there was:

> Promotion out of archival requires a `+1` above a strength threshold, whatever the source.

That is a strictly better rule than whitelisting sources — one condition instead of a list to maintain, and a new signal channel could earn promotion by being precise rather than by being added to an allowlist.

It is currently unsound, in two distinct ways.

**Within the heuristic channel the threshold is decorative.** Each tier is pinned to a constant, so any threshold placed between two tiers excludes exactly the set a tier rule would exclude.

**Across channels the threshold is inverted.** `submit_feedback` is pinned at `1.0` — the *maximum* — while `auto_llm_judge` reports the model's genuine confidence, typically below 1.0 and floored at `feedback_llm_judge_min_confidence = 0.75` (`config.py:255`). So any threshold at or below 1.0 admits **every** implicit agent self-assessment while filtering out honest retroactive judgements. That is backwards: a judge reading the finished response is stronger evidence than a model's in-flight claim about its own retrieval, and the threshold ranks them the other way round.

#191 therefore uses source-based filtering, which has a known cost: excluding `auto_heuristic` wholesale discards the 45 `node_id`/`title`/`sentence` verbatim matches — the strongest evidence of use in the system — because they share a source label with the 1,520 weak ones. The `evidence.match` field distinguishes them, so a detector-class rule could recover those without needing `strength` at all.

## Three things to fix, not two

1. **`submit_feedback` should record a real strength.** At minimum, explicit and implicit should differ. `affinity_implicit_weight = 0.8` (`config.py:252`) already encodes that judgment for the affinity boost; the signals path ignores it.

2. **`token_overlap` should vary with `overlap_ratio`.** Either widen the output range or drop the `min()` so the ratio actually moves the number. The current formula cannot distinguish its own best case from its worst.

3. **Strength must become comparable across channels.** This is the part that is easy to miss: fixing the two formulas independently is not sufficient. Strength today is confidence *within a detection channel*, not a calibrated probability — a 0.7 from the transcript watcher and a 0.7 from `submit_feedback` would still not mean the same thing. A threshold rule requires cross-channel comparability, which is a design decision, not a formula change.

Until (3) is answered, a threshold over mixed sources remains unsound even with (1) and (2) fixed.

## Related

- #191 — the design debate this affects; uses source-based filtering instead
- `strength` is written in `signals` only, never in `affinity` (`index/db.py:363` vs `:450`), so any promotion rule using it needs a join or a new column

## How it was found

Cause 1 surfaced in an adversarial review of the #191 implementation proposal. Cause 2 was found independently by @AndreLFSMartins while measuring the tier breakdown requested in that thread. Neither is visible from the other channel alone.

AndreLFSMartins:
Closes #218.

## The problem

`signals.strength` is documented as a per-event confidence score. In practice it is a tier label stored as a float:

- `submit_feedback` declares `strength` in its INSERT column list and then passes the literal `1.0`, so an explicit user confirmation and an implicit agent self-assessment are stored identically — `source` is the only discriminator.
- Three heuristic match kinds return fixed constants.
- `token_overlap` computes `min(0.85, 0.45 + ratio)` behind an entry gate of `ratio >= 0.5`. The `min()` saturates at `ratio >= 0.40`, so **every match that reaches the branch is already at the ceiling**.

Measured on a live store (read-only snapshot, 6,355 nodes), positive heuristic pairs on injected whispers:

| match | pairs | share |
|---|---|---|
| `token_overlap` | 1,587 | 97.4% |
| `node_id` | 29 | 1.8% |
| `sentence` | 13 | 0.8% |
| `title` | 0 | 0% |

`overlap_ratio` across those 1,587 rows ranges **0.50 to 7.583** (mean 1.452, median 1.167, with 38% at ≥ 1.50) — and all 1,587 report exactly `0.85`. Three distinct strength values across the whole positive heuristic channel.

## The design decision to review

**`strength` becomes the strength of the evidence backing a row's polarity, on a single ordinal scale, comparable in rank across channels.** It is explicitly *not* a calibrated probability, and the module docstring says so.

The load-bearing assertion — and the thing worth attacking first — is that **the channel dominates within-channel confidence**. A verbatim match outranks any LLM judgment however confident; an LLM judgment outranks any agent self-report. This is #218's own argument (*"a judge reading the finished response is stronger evidence than a model's in-flight claim about its own retrieval"*) carried to its conclusion. Bands are therefore disjoint per channel:

| band | channel | mapping |
|---|---|---|
| `1.00` | `explicit` | constant |
| `0.98` | heuristic `node_id` | constant |
| `0.94` | heuristic `title` | constant |
| `0.92` | heuristic `sentence` | constant |
| `0.82–0.90` | `auto_llm_judge` | affine over `[min_confidence, 1.0]` |
| `0.80` | `implicit` | constant |
| `0.40–0.78` | heuristic `token_overlap` | asymptotic in `overlap_ratio` |
| `0.00` | any row with `polarity = 0` | convention |

If you reject that ordering rule, the design collapses back to per-source thresholds — so it is the one premise to push on.

Two notes on the mapping:

- **`token_overlap` is asymptotic, not clamped.** `overlap_ratio = len(overlap) / min(len(candidate_tokens), 12)` is unbounded above, so any linear map needs an arbitrary clamp and every clamp recreates the saturation defect. A clamp at 1.50 would tie 38% of observed rows. `0.40 + 0.38·(1 − e^−(r−0.5))` gives variance across the whole domain; the supremum 0.78 sits below `implicit = 0.80` by construction. On real data this takes the channel from **3 distinct values to 59** — injective over the 57 distinct stored ratios.
- **`polarity = 0` stores `0.00`.** The judge previously wrote its confidence on rows that assert nothing. Its raw confidence still lives in `signals.evidence`. This is safe because the judge assigns non-zero polarity only when `confidence >= min_confidence`, so no scoring row can fall below its band floor.

## Backfill

Historical rows would otherwise keep pre-ladder values, and the 1,587 `token_overlap` rows pinned at `0.85` land **inside the new judge band** — the column would go on lying, now about which channel produced the row.

The recompute is exact, not estimated: no `signals` row in the measured store lacks `evidence`, heuristic rows carry `match` and `overlap_ratio`, and judge rows carry the `min_confidence` in force **when they were written**, so a row normalises to what it would have stored had the ladder existed then. Only `strength` is written; `evidence` and `polarity` are untouched, which makes a later ladder revision re-runnable from the same source.

## What no reader means

`signals.strength` has **no readers**. A sweep of `src/`, `ui/`, `schema.sql` and `db.py` finds only write paths and unrelated homonyms (`belief strength` in `tool_schemas.py`, a force-simulation local in `GraphView.tsx`). This PR therefore produces **no observable behaviour change**, and `tests/test_engine/test_confirmed_use_contract.py` staying green is the intended proof. If you find a reader I missed, that invalidates the risk assessment.

## Review history

Two adversarial review rounds ran before this PR, and both found real defects:

1. A one-time migration stamp could not repair rows an older binary writes *after* the stamp — a rollback then re-upgrade, or the second unmanaged server process this project already documents. The migration now runs at every startup, bounded by a cutoff on `signals.id`. Rescanning is a repair rather than a drift only because `strength_from_evidence` is pure and every write site stores exactly what it returns: recomputing over all 5,354 rows of a real store changes **0** rows.
2. Fixing (1) surfaced a second issue: `json.loads` accepts bare `NaN`/`Infinity` and returns an unbounded `int` for a large integer literal, `float()` of which raises `OverflowError` rather than `ValueError`. SQLite stores a NaN REAL as NULL and `signals.strength` is `NOT NULL` — so with an every-boot migration, one malformed historical row would stop the server from **ever** starting again. `_as_float` now catches `OverflowError` and rejects non-finite results.

A third defect was found in between: the detector computed `token_overlap` strength from the full-precision ratio while recording `round(ratio, 3)` in `evidence`, so the migration's own recompute disagreed with what the write site stored (`0.45833694` vs `0.45844415`). It now rounds once and uses that value for both. **The admission gate still reads the raw ratio**, so which responses count as referenced is deliberately unchanged.

## Testing

39+12 = 51 new tests. Local baseline on an untouched `upstream/main` checkout is **1948 passed, 4 failed**; with this branch, **2003 passed, the same 4 failed**. Those 4 are pre-existing and unrelated to this change — three in `tests/test_setup.py::TestConfigureCodexMcp` (the test patches `shutil.which`, but `_find_binary` falls past it to hardcoded paths including `/opt/homebrew/bin/codex`, which exists on my machine) and one `test_detect_returns_none_for_home` artifact of running with an isolated `HOME`. Both look worth their own issue; neither is touched here.

`ruff check src/ tests/` is clean.

## Out of scope

Admitting `auto_heuristic` into `_CONFIRMED_USE_SOURCES`, calling `_claim_confirmed_use` from the heuristic commit block, and the judge suppression at `if llm_judge_enabled and not has_llm_judge and not referenced`. Those are **#272**, which this PR unblocks — the #220 design deferred `auto_heuristic` admission explicitly "until #218 provides signal calibration", and Contract 9 still pins the exclusion.

### DIFF
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -550,7 +550,9 @@ def _record_whisper_usage_signals(
                 "row": row,
                 "signal_type": signal_type,
                 "polarity": polarity,
-                "strength": confidence,
+                "strength": signal_strength.judge_strength(
+                    confidence, min_confidence, polarity
+                ),
                 "evidence": {
                     "detector": _LLM_JUDGE_SOURCE,
                     "verdict": verdict,
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -148,10 +148,16 @@ def _node_usage_evidence(row, response_text: str) -> tuple[bool, float, dict]:
     denominator = min(len(candidate_tokens), 12)
     overlap_ratio = (len(overlap) / denominator) if denominator else 0.0
     if len(overlap) >= 4 and overlap_ratio >= signal_strength.OVERLAP_GATE:
-        return True, signal_strength.token_overlap_strength(overlap_ratio), {
+        # Round ONCE, then use that same value for both the strength and the evidence.
+        # The #218 backfill recomputes strength from evidence, so a strength derived
+        # from a more precise ratio than the one recorded makes the recompute perturb
+        # a correct row instead of confirming it. The gate above still reads the raw
+        # ratio, so which responses count as referenced is unchanged.
+        recorded_ratio = round(overlap_ratio, 3)
+        return True, signal_strength.token_overlap_strength(recorded_ratio), {
             "match": "token_overlap",
             "overlap": overlap[:12],
-            "overlap_ratio": round(overlap_ratio, 3),
+            "overlap_ratio": recorded_ratio,
         }
 
     return False, 0.0, {
--- a/src/ormah/background/session_watcher.py
+++ b/src/ormah/background/session_watcher.py
@@ -16,6 +16,7 @@ from threading import Event, Lock, Thread, Timer
 from watchdog.events import FileSystemEventHandler
 from watchdog.observers import Observer
 
+from ormah import signal_strength
 from ormah.engine.memory_engine import MemoryEngine
 from ormah.text.tokens import distinctive_tokens
 from ormah.transcript.parser import (
@@ -37,8 +38,8 @@ class IngestResult(Enum):
 
 _STATE_FILENAME = ".session_watcher_state"
 MAX_RECONCILE_RETRIES = 3
-_HEURISTIC_SOURCE = "transcript_watcher_heuristic"
-_LLM_JUDGE_SOURCE = "transcript_watcher_llm_judge"
+_HEURISTIC_SOURCE = signal_strength.HEURISTIC_SOURCE
+_LLM_JUDGE_SOURCE = signal_strength.LLM_JUDGE_SOURCE
 _HEURISTIC_AFFINITY_SOURCE = "auto_heuristic"
 _LLM_JUDGE_AFFINITY_SOURCE = "auto_llm_judge"
 _FENCE_RE = re.compile(r"```(?:json)?\s*\n(.*?)```", re.DOTALL)
@@ -116,13 +117,13 @@ def _node_usage_evidence(row, response_text: str) -> tuple[bool, float, dict]:
     node_id = row["node_id"]
     short_id = node_id[:8] if node_id else ""
     if short_id and short_id.lower() in response_text.lower():
-        return True, 1.0, {"match": "node_id", "short_id": short_id}
+        return True, signal_strength.VERBATIM_NODE_ID, {"match": "node_id", "short_id": short_id}
 
     title = row["title"] or ""
     title_tokens = distinctive_tokens(title, extra_stop_words={"memory", "ormah"})
     title_norm = _normalise_text(title)
     if len(title_tokens) >= 2 and len(title_norm) >= 12 and title_norm in response_norm:
-        return True, 0.95, {"match": "title", "title": title}
+        return True, signal_strength.VERBATIM_TITLE, {"match": "title", "title": title}
 
     content = row["content"] or ""
     for sentence in re.split(r"[\n.!?]+", content):
@@ -132,7 +133,10 @@ def _node_usage_evidence(row, response_text: str) -> tuple[bool, float, dict]:
         sentence_tokens = distinctive_tokens(sentence, extra_stop_words={"memory", "ormah"})
         sentence_norm = _normalise_text(sentence)
         if len(sentence_tokens) >= 4 and sentence_norm in response_norm:
-            return True, 0.9, {"match": "sentence", "text": sentence[:160]}
+            return True, signal_strength.VERBATIM_SENTENCE, {
+                "match": "sentence",
+                "text": sentence[:160],
+            }
 
     node_tokens = distinctive_tokens(
         f"{title} {content}",
@@ -143,8 +147,8 @@ def _node_usage_evidence(row, response_text: str) -> tuple[bool, float, dict]:
     overlap = sorted(candidate_tokens & response_tokens)
     denominator = min(len(candidate_tokens), 12)
     overlap_ratio = (len(overlap) / denominator) if denominator else 0.0
-    if len(overlap) >= 4 and overlap_ratio >= 0.5:
-        return True, min(0.85, 0.45 + overlap_ratio), {
+    if len(overlap) >= 4 and overlap_ratio >= signal_strength.OVERLAP_GATE:
+        return True, signal_strength.token_overlap_strength(overlap_ratio), {
             "match": "token_overlap",
             "overlap": overlap[:12],
             "overlap_ratio": round(overlap_ratio, 3),
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import Any
 
 from ormah import lifecycle
+from ormah import signal_strength
 from ormah.config import Settings
 from ormah.embeddings.text import embedding_text as _embedding_text
 from ormah.engine.context_builder import ContextBuilder
@@ -2800,7 +2801,7 @@ class MemoryEngine:
                     resolved_node_id,
                     "feedback_submitted",
                     signal,
-                    1.0,
+                    signal_strength.feedback_strength(source, signal),
                     source,
                     session_id,
                     "submit_feedback",
--- a/src/ormah/engine/memory_engine.py
+++ b/src/ormah/engine/memory_engine.py
@@ -172,32 +172,60 @@ class MemoryEngine:
         self._warmup_embedder()
         self._warmup_reranker()
 
-    def _migrate_signal_strength(self) -> None:
-        """Recompute signals.strength onto the #218 ordinal evidence ladder.
-
-        Exact rather than estimated: every row carries the evidence its own channel
-        needs, and the judge stamps the min_confidence in force when the row was
-        written -- so a row recomputes to what it would have stored had the ladder
-        existed then, not to what today's settings would produce.
-
-        Re-runnable by construction: evidence is never written, so a later revision
-        of the ladder recomputes from the same untouched source. Only strength moves.
-
-        No file_store call, so this does not take db-lock before memory-lock and
-        stays outside the ordering hazard documented on _record_confirmed_use.
-        """
-        stamp = self.db.conn.execute(
-            "SELECT value FROM meta WHERE key = 'signal_strength_ladder_version'"
-        ).fetchone()
+    def _meta_int(self, key: str) -> int:
+        """Read an integer meta value, treating absent or unparseable as 0."""
+        row = self.db.conn.execute("SELECT value FROM meta WHERE key = ?", (key,)).fetchone()
         try:
-            version = int(stamp["value"]) if stamp else 0
+            return int(row["value"]) if row else 0
         except (TypeError, ValueError):
-            version = 0
-        if version >= SIGNAL_STRENGTH_LADDER_VERSION:
-            return
+            return 0
+
+    def _migrate_signal_strength(self) -> None:
+        """Normalise signals.strength onto the #218 ordinal evidence ladder.
+
+        Runs on every startup, not once. The first pass covers the whole table; every
+        later pass covers only rows that appeared since the previous one, tracked by a
+        cutoff on signals.id (AUTOINCREMENT, so ids are monotonic and never reused).
+
+        The rescan exists because a one-time stamp cannot repair what an OLD binary
+        writes AFTER the stamp is set -- a rollback then re-upgrade, or the second
+        unmanaged server process this project already knows about (#238). Those rows
+        would carry pre-ladder values forever on a table the stamp calls migrated, and
+        the column would silently mix incomparable scales.
+
+        Rescanning already-correct rows is safe because strength_from_evidence is a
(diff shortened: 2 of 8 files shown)


r-spade:
Thanks Andre — I want to revisit the load-bearing ordering decision before we continue.

Ormah is an agent-native memory layer. `implicit` is not weak, vague feedback: it is a deliberate agent confirmation tied to a specific injected memory event.

In our active store, we have 342 positive `implicit` signals and zero positive human/explicit signals. Humans are not expected to sit and curate memory feedback; agents are the practical learning loop.

For that reason, placing `implicit` at `0.80` below the LLM judge and automatic match types does not fit the product. I think deliberate agent confirmation should be a high band, close to explicit confirmation. Automatic heuristics and judges are useful corroboration, but should not outrank a deliberate confirmation by default.

Since `strength` has no reader today, could we pause and rework this ordering policy before locking in the backfill/migration? We have room to get the semantics right before #272 consumes it. If you think broader input would help, I’d be happy to make this a short design discussion rather than force the decision into this PR.

AndreLFSMartins:
Thanks for stopping this before the migration, and yes to the design discussion. I went to measure before answering, and I think your case is stronger than the way you put it.

`origin/main`, `memory_engine.py:2803`:

```python
                    "feedback_submitted",
                    signal,
                    1.0,
                    source,
```

`implicit` is at `1.0` today. Tied with `explicit`, and nothing has ever outranked it. So this PR was never declining to promote it, it takes it from the top rung to the seventh of eight. I never framed it that way for you because I had not seen it myself, and it is the part of your objection I would have answered first if I had.

Three things I checked that hold up your side.

I tried to falsify the premise that `implicit` is a deliberate act, since your argument inverts if it turns out to come from the judge or a matcher. It does not. `tool_schemas.py:176` → `mcp_adapter.py:375` → `routes_agent.py:196` → `memory_engine.py:2790`, all on `origin/main`, and `grep -n implicit src/ormah/background/session_watcher.py` exits 1. The watcher writes `transcript_watcher_heuristic` and `transcript_watcher_llm_judge`, never `implicit`. No UI path, no CLI, no hook.

"`strength` has no reader today" is literally correct on your repo. `signal_strength.py` and `HEURISTIC_CONFIRM_FLOOR` do not exist on `origin/main` at all. I had been reasoning from my own tree, where #272 is already merged locally, so the window you describe is real and I was arguing from a state you cannot see.

The product premise holds too. `human` has no producer anywhere in `src`. Zero feedback surfaces in `ui/src` and in the CLI. `review_log.answered` is 0 across all 243 rows on my install, over the store's full 21 days. On a stock config the judge is off, which leaves `implicit` as the only deliberate positive channel that exists at all. Ranking the only one seventh is hard to defend.

One correction, and it makes the disagreement smaller than it looks:

```
token_overlap  1923   max strength 0.77968
node_id          33
sentence         15
```

`token_overlap` is 97.6% of the heuristic volume here, and its ceiling sits at `0.7797`, hard under the `implicit` rung at `0.80`. It never outranked `implicit`. Only the judge did, by `0.02`, plus three verbatim match types that recompute from stored evidence. So the automatic channel worth worrying about was already underneath it.

On #218: the ordering I shipped came from your own analysis there. "a judge reading the finished response is stronger evidence than a model's in-flight claim about its own retrieval" is your sentence, and the `0.80` came from the `affinity_implicit_weight` precedent you cited a few paragraphs down. I implemented it literally. My own design note flagged the ranking as assumed and untested and wrote "check whether the maintainer accepts channel-dominates-confidence" right next to it, and then I never checked. That one is mine, and it is a good argument for having this conversation now rather than after the backfill.

**What I would propose.** `IMPLICIT` to `0.99`, second only to `explicit` and above everything automatic. The number is forced rather than picked: `test_signal_strength.py:11` in this PR pins the bands disjoint, `EXPLICIT` is `1.00`, `VERBATIM_NODE_ID` is `0.98`, so `(0.98, 1.00)` is the only room left for "high band, close to explicit".

One thing would have to move with it, and it is the only place I would push back. #272 pins `HEURISTIC_CONFIRM_FLOOR = signal_strength.IMPLICIT` (`memory_engine.py:64` on that branch, three call sites: `session_watcher.py:499`, `memory_engine.py:360` and `:3021`). The rung and the admission gate are the same constant, so raising one raises the other and nothing tells us:

| floor | heuristic rows admitted | nodes |
|---|---|---|
| 0.80 (today) | 48 | 23 |
| 0.92 | 48 | 23 |
| 0.94 / 0.96 | 33 | 14 |
| 0.99 | 0 | 0 |

At `0.99` the gate admits nothing, `_record_confirmed_use` never fires, and that is the only `archival → working` transition in the system. The job still reports `ok`. It fails green.

Pointing the gate at `VERBATIM_SENTENCE` instead costs nothing, and not as an argument: the heuristic channel has no signals at all between `0.78` and `0.92`, so both floors admit the same 48 rows. It also matches what the constant's own comment says it meant, and what its tests describe, verbatim in and token_overlap out. One consequence worth putting up front: on the #274 branch, `tests/test_engine/test_confirmed_use_contract.py:353` parametrizes the floor with the literal `0.80`, so that expectation moves to `0.92` in the same commit. I would rather flag that here than have it read later as a test bent green.

**Two problems with the channel itself,** which I think are worth separate issues, and both argue for tightening it rather than keeping it low.

Every `implicit` row on my store carries the same evidence blob:

```json
{"source": "implicit"}
```

All 169 of them, byte-identical. The heuristic channel stores four to five keys, the judge stores seven, and both recompute. `implicit` records that it happened and nothing else. It is a claim with no receipt, and promoting it does not fix that, it raises the price of it. Neither of us can audit one of those rows today, including for this conversation.

The second: 69 of those 169 (40.8%), 24 of them positive, are anchored to whisper events with `was_injected = 0`. The FK is nullable, there is a most-recent-event fallback with no filter (`origin/main`, `memory_engine.py:2620`, whose docstring at `:2733` already admits "that fallback is not exact"), and `_REVIEW_FRAMING` tells the agent to use `source="implicit"` on a non-injected event. The consuming SQL gates on `was_injected = 1`, so nothing is broken today. That is the reason I would rather fix it before the rung goes up than after.

There is a third next door if you want it filed: `affinity.py:102` hands the same `implicit_w` to `implicit`, `auto_heuristic` and `auto_llm_judge`, so that path already flattens the distinction this ladder is trying to draw.

On the 342: I cannot reproduce it here, and I am not disputing it. Mine returns `169 / 105`:

```sql
SELECT COUNT(*) AS total, SUM(polarity = 1) AS positive
FROM signals WHERE source = 'implicit';
```

105 positives across 21 days on an 8,606-node store, so about five a day. At that rate 342 is just a store a few times older than mine, which is the boring explanation and probably the right one. Is that roughly the shape of yours?

One caveat over all of it: my Beta runs `ORMAH_FEEDBACK_LLM_JUDGE_ENABLED=true`, which is not the default, so the judge rung weighs more in my numbers than it would on a stock install. If yours is stock, this reordering moves less than it moves here.

Nothing is pushed, and I am happy to keep this at the discussion stage as long as it is useful. If the shape works for you, I will land it as two commits, the rung here and the gate in #274, on a clean island with CI rerun. And if you would rather set a different number or take the ordering somewhere broader first, that is genuinely fine by me. It is your call to make, and I would rather you made it than have me guess at it twice.

PR #275, opened 2026-08-31
fix(db): retire per-thread SQLite connections on thread death (fd leak, #83)

### LINKED ISSUE #83: bug(db): per-thread SQLite connections never released → fd leak → [Errno 24] Too many open files under long-running server

AndreLFSMartins:
Fixes #83.

## What

`Database._new_connection` registers every per-thread connection in `_all_conns` and never closes it when the owning thread dies — the only close path is `Database.close()` at shutdown. Ephemeral threads (the maintenance phases spawn a fresh daemon thread per cycle) each leak a connection and its ~3 WAL fds for the whole process lifetime, until `[Errno 24] Too many open files` wedges the server (2641 fds on `index.db` observed in a live incident — details in #83).

## Fix

Register a `weakref.finalize` on the owning thread object at connection creation, so the connection is closed and removed from `_all_conns` when the thread is garbage-collected:

```python
weakref.finalize(threading.current_thread(), self._retire_connection, conn)
```

Root-cause and driver-agnostic — no need to hunt every ephemeral-thread call site. `Database.close()` still handles any still-live threads' connections at shutdown, and `_retire_connection` is idempotent with it.

## Testing

- New `tests/test_index/test_db_fd_lifecycle.py`: spawns ephemeral threads that each touch `db.conn`, joins them, forces GC, and asserts the connections are retired (fails on `main` today, passes with the fix).
- Full suite on this branch: 1949 passed; the 4 failures (`test_space_detect`/`TestConfigureCodexMcp`) also fail on unmodified `upstream/main` in the same environment — pre-existing, unrelated.

This fix has also been running continuously in my downstream beta since 2026-07-05 with no recurrence of the fd growth.

cc @r-spade — flagging because the failure mode takes down a long-running server.

### DIFF
--- a/src/ormah/index/db.py
+++ b/src/ormah/index/db.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import sqlite3
 import threading
+import weakref
 from contextlib import contextmanager
 from pathlib import Path
 
@@ -48,6 +49,11 @@ class Database:
             pass  # sqlite-vec not installed — vector search disabled
         with self._conns_lock:
             self._all_conns.append(conn)
+        # Auto-close this connection when its OWNING thread is garbage-collected. Without this,
+        # an ephemeral thread (e.g. a maintenance phase thread) leaks its connection — and its
+        # ~3 WAL fds — for the whole process lifetime, until [Errno 24]. finalize fires at most
+        # once; close() (shutdown) also handles any still-live threads' connections.
+        weakref.finalize(threading.current_thread(), self._retire_connection, conn)
         return conn
 
     @property
@@ -542,6 +548,18 @@ class Database:
                 )
         except ImportError:
             pass  # sqlite-vec not available, vector search disabled
+
+    def _retire_connection(self, conn: sqlite3.Connection) -> None:
+        """Drop a dead thread's connection from the registry and close it."""
+        with self._conns_lock:
+            try:
+                self._all_conns.remove(conn)
+            except ValueError:
+                pass  # already retired by close()
+        try:
+            conn.close()
+        except Exception:  # noqa: BLE001
+            pass
 
     def close(self) -> None:
         with self._conns_lock:
--- a/src/ormah/index/db.py
+++ b/src/ormah/index/db.py
@@ -22,6 +22,7 @@ class Database:
         self.db_path.parent.mkdir(parents=True, exist_ok=True)
         self._local = threading.local()
         self._all_conns: list[sqlite3.Connection] = []
+        self._finalizers: list[weakref.finalize] = []
         self._conns_lock = threading.Lock()
         self._lock = threading.RLock()  # serializes write transactions across threads
 
@@ -47,13 +48,19 @@ class Database:
             conn.enable_load_extension(False)
         except ImportError:
             pass  # sqlite-vec not installed — vector search disabled
-        with self._conns_lock:
-            self._all_conns.append(conn)
         # Auto-close this connection when its OWNING thread is garbage-collected. Without this,
         # an ephemeral thread (e.g. a maintenance phase thread) leaks its connection — and its
         # ~3 WAL fds — for the whole process lifetime, until [Errno 24]. finalize fires at most
         # once; close() (shutdown) also handles any still-live threads' connections.
-        weakref.finalize(threading.current_thread(), self._retire_connection, conn)
+        finalizer = weakref.finalize(threading.current_thread(), self._retire_connection, conn)
+        with self._conns_lock:
+            self._all_conns.append(conn)
+            # A finalizer keeps its callback and args alive, so this bound `_retire_connection`
+            # pins the Database until the OWNING thread dies — never, on a long-lived one.
+            # close() detaches them; drop the ones that already fired so the list stays bounded
+            # by live threads instead of growing once per thread ever seen.
+            self._finalizers = [f for f in self._finalizers if f.alive]
+            self._finalizers.append(finalizer)
         return conn
 
     @property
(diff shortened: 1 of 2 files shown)


r-spade:
Thanks Andre. I confirmed the core fix bounds the FD growth.

One lifecycle question before merge: `weakref.finalize` retains the bound `_retire_connection` callback and `conn` until the thread object is collected. In a small repro, a `Database` remained alive after `close()`, deletion, and GC because its finalizer was still attached to a long-lived thread.

This is much lower severity than #83, but do you think we should track and detach the finalizers in `Database.close()` to avoid retaining closed temporary databases?

AndreLFSMartins:
Good catch. I reproduced it, and it behaves exactly as you describe:

```
Database created on the main thread      -> alive after close+del+gc? True
Database created on an ephemeral thread  -> alive after close+del+gc? False
```

Two things narrow the blast radius, and both are worth stating before the fix. No fds are involved: `close()` had already closed the retained connection, so what stays resident is the `Database` object plus a dead connection object, not descriptors. And the #83 case does not regress. The ephemeral thread dies, the finalizer fires, nothing is held (the `False` above).

Pushed two commits.

`a504a32` is what you asked for. `close()` now detaches the tracked finalizers, and `_new_connection` prunes the ones that already fired, so the list stays bounded by live threads instead of growing once per thread ever seen. `close()` also does its closing outside the registry lock, since nothing there needs it.

`b282458` is a second finding I ran into while writing the test for the first one. Drop it if you would rather keep this PR narrow. CPython runs weakref callbacks from the generational GC, which triggers on allocation, including an allocation inside `_new_connection`'s own critical section. If a dead thread gets collected right there, `_retire_connection` re-enters `_conns_lock` on the very thread that holds it. With a plain `Lock` that is a hard deadlock, and it does not end with the test: my first version of the test deadlocked for real and hung the interpreter at exit, because the finalizers still registered run in weakref's atexit hook and block on the same lock. So the commit makes `_conns_lock` an `RLock`, and the test probes reentrancy non-blockingly, precisely because a real deadlock takes the whole test process down with it.

I honestly do not know how likely that GC window is in production. It is narrow. But the fix is one word and the failure mode is a hang, so for me it is worth paying now.

Full suite on a clean island cut from `upstream/main`: 1951 passed, 4 failed. The 4 are `test_space_detect` and the three `TestConfigureCodexMcp`, all of them pre-existing on `upstream/main` at 90c431e with the same names.

r-spade:
I added a small test-cleanup commit on top. The lifecycle and re-entrancy tests now call `Database.close()` in `finally`, so they release their connection/finalizer even if an assertion fails. Verified the lifecycle tests, full index suite, and Ruff. Thanks for addressing the retention concern so thoroughly.

AndreLFSMartins:
Thanks, and the `try/finally` is the right call: a failing assertion was leaving behind exactly the object this PR stops retaining. I pulled it and ran it here, `tests/test_index/` is 40 passed.

One heads-up, and it is about the repo rules rather than the code. Now that `9a2e687` was pushed to the branch, the PR reads `mergeStateStatus: BLOCKED` / `reviewDecision: REVIEW_REQUIRED` while still `MERGEABLE`. The ruleset asks for an extra approval on unattributed changes, and the pusher's own approval does not count toward it. I hit the same thing on an earlier PR here and it sat for a while before anyone noticed, so I would rather flag it now.

I do not know what the cleanest way around it is on your side, so I will leave that to you.

r-spade:
> Thanks, and the `try/finally` is the right call: a failing assertion was leaving behind exactly the object this PR stops retaining. I pulled it and ran it here, `tests/test_index/` is 40 passed.
> 
> One heads-up, and it is about the repo rules rather than the code. Now that `9a2e687` was pushed to the branch, the PR reads `mergeStateStatus: BLOCKED` / `reviewDecision: REVIEW_REQUIRED` while still `MERGEABLE`. The ruleset asks for an extra approval on unattributed changes, and the pusher's own approval does not count toward it. I hit the same thing on an earlier PR here and it sat for a while before anyone noticed, so I would rather flag it now.
> 
> I do not know what the cleanest way around it is on your side, so I will leave that to you.

The PR rules, I have been meaning to sort that out soon. Will get to it and clean it up to make things easier. Thanks for flagging it.

PR #283, opened 2026-09-03
fix(whisper): PT-BR temporal keywords narrow the search window too (#135)

### LINKED ISSUE #135: bug(whisper): temporal keyword regexes are English-only — PT-BR "ontem"/"semana passada" get no window narrowing
## Summary

`_TIME_KEYWORDS` (`src/ormah/engine/prompt_classifier.py:81`) is 100% English regex. A PT-BR temporal query ("ontem", "semana passada", "últimos 3 dias") is correctly classified as `temporal` by the embedding-based intent classifier (see below), but the keyword regex layer that narrows the search window (e.g. 48h→24h for "yesterday"-equivalents) and strips temporal noise tokens from the query never fires for PT-BR text — the query falls back to the default 3-day window and keeps the temporal phrase as search noise.

## What's NOT the bug

The archetype-based intent classifier is not the bottleneck. Verified live with the production encoder (bge-m3 @ Ollama, 2026-07-15): the current English archetypes already classify PT-BR prompts cross-lingually — 14/16 correct (temporal 0.78–0.96, identity 0.79–0.94). Adding PT-BR archetypes only widens margins (15/16; temporal → 1.00; identity's one miss fixed 0.56 → 0.95) — an upgrade, not a rescue. Production's 93% `general` intent rate is explained by real prompt mix + synthetic traffic (see companion issue on synthetic-prompt filtering), not classifier failure.

## Proof

`tests/test_whisper_claims_investigation.py` §1 (unit) + `test_live_bge_m3_classifies_ptbr_against_en_archetypes` (integration, requires live Ollama).

## Proposed fix

Add PT-BR patterns to `_TIME_KEYWORDS` mirroring the EN set (ontem, hoje, semana passada, últimos N dias, etc.).

## Source

`docs/investigation-2026-07-15-whisper-pipeline.md`, item [1].

AndreLFSMartins:
Closes #135.

`_TIME_KEYWORDS` and `_NUMERIC_TIME_RE` in `engine/prompt_classifier.py` were
English-only, so a PT-BR temporal prompt reached the retrieval layer with two
things missing: the narrowed window (it fell through to the 3-day default) and
the phrase stripping (the temporal words stayed in the query as search noise).
The intent classifier itself was never the problem — #135 shows it already
classifies PT-BR cross-lingually.

## What this adds

PT-BR patterns mirroring the existing English set, one for one:

| EN | PT-BR | window |
|----|-------|--------|
| `today` | `hoje` | 24h ago → now |
| `yesterday` | `ontem` | 48h ago → 24h ago |
| `last week` | `semana passada` | 14d ago → 7d ago |
| `this week` | `esta/essa/nesta/nessa semana` | 7d ago → now |
| `last month` | `mês passado` | 60d ago → 30d ago |
| `recently`, `lately` | `recentemente`, `ultimamente` | 3d ago → now |
| `last/past N <unit>` | `últimos/últimas N <unidade>` | same rolling rule |

The same patterns go into `_TEMPORAL_STRIP_PATTERNS`, so the phrase is stripped
from the search query exactly as the English one is.

## Two details worth reviewing

`_UNIT_ALIASES` folds the PT-BR unit onto the English key the tables are written
in. Without it `_UNIT_TO_DAYS.get(unit, 1)` falls back to 1 day silently, so
"últimas 2 semanas" would have meant 2 days, and the rolling-window branch —
which tests for `"week"`/`"month"` — would never have fired.

`extract_time_params` now lowercases before stripping the plural `s`, instead of
after. `"DIAS".rstrip("s")` is `"DIAS"`, which matched nothing; `"DIAS".lower()`
then `.rstrip("s")` is `"dia"`, which the alias table resolves.

## Tests

18 new cases in `tests/test_engine/test_prompt_classifier.py`. Every asserted
window is deliberately different from the 3d→now default, so a passing test
proves the keyword actually matched rather than falling through to the fallback.
The two keywords that do map to the default window (`recentemente`,
`ultimamente`) are covered through `has_temporal_phrases`, which skips the
default entirely and can tell a match from a fallback.

`pytest tests/ -q` → 2022 passed. Four failures are pre-existing on `main` and
unrelated to this change: `test_detect_returns_none_for_home` (fails identically
on an untouched tree) and three `TestConfigureCodexMcp` cases. `ruff check
src/ tests/` → clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,3 +127,12 @@ ORMAH_WORKING_DECAY_DAYS=14
 
 # Memory consolidation (requires LLM)
 # ORMAH_CONSOLIDATION_INTERVAL_MINUTES=360
+
+# ── Temporal locale packs ────────────────────────────────────────────────────
+#
+# Which language packs are consulted when a time reference is parsed out of a
+# prompt ("last week", "na semana passada"). Comma-separated, order kept.
+#
+# Built-in codes: en, pt-BR (case-sensitive; an unknown code fails startup)
+#
+# ORMAH_TEMPORAL_LOCALES=en,pt-BR
--- a/docs/12 - Configuration Reference.md
+++ b/docs/12 - Configuration Reference.md
@@ -296,6 +296,17 @@ backup taken before the upgrade.
 | `claude_maintenance_interval_hours` | `24` |
 | `claude_maintenance_batch_size` | `25` |
 
+## Temporal Locales
+
+| Setting | Default |
+|---|---|
+| `temporal_locales` | `en,pt-BR` |
+
+The language packs consulted when a time reference is parsed out of a prompt.
+Comma-separated and order-preserving; the built-in codes are `en` and `pt-BR`,
+and they are case-sensitive. An unknown code is rejected at startup. The
+setting selects packs only — it never carries grammar.
+
 ## Code Anchor
 
 - `src/ormah/config.py`
--- a/src/ormah/config.py
+++ b/src/ormah/config.py
@@ -8,6 +8,8 @@ from pathlib import Path
 from pydantic import field_validator
 from pydantic_settings import BaseSettings
 
+from ormah.engine.temporal import parse_locale_codes
+
 
 _ENV_FILES = [
     Path.home() / ".config" / "ormah" / ".env",  # Fixed global config
@@ -291,6 +293,14 @@ class Settings(BaseSettings):
     claude_maintenance_enabled: bool = False
     claude_maintenance_interval_hours: int = 24  # hours between maintenance runs
     claude_maintenance_batch_size: int = 25  # candidates per type per run
+
+    # Temporal locale packs consulted when parsing a time reference out of a
+    # prompt. Declared as a plain ``str`` and parsed by the validator below,
+    # not as a ``list[str]``: pydantic-settings JSON-decodes complex types from
+    # the environment and would reject the comma-separated form the ``.env``
+    # file uses everywhere else. The setting selects packs; it never carries
+    # grammar.
+    temporal_locales: str = "en,pt-BR"
 
     # --- Validators ---
 
@@ -645,6 +655,17 @@ class Settings(BaseSettings):
         if v < 1:
             raise ValueError(f"interval must be >= 1 minute, got {v}")
         return v
+
+    @field_validator("temporal_locales")
+    @classmethod
+    def _temporal_locales_known(cls, v: str) -> str:
+        parse_locale_codes(v)  # raises on an empty result or an unregistered code
+        return v
+
+    @property
+    def temporal_locale_codes(self) -> tuple[str, ...]:
+        """The enabled locale codes, ordered, trimmed and de-duplicated."""
+        return parse_locale_codes(self.temporal_locales)
 
     @property
     def llm_enabled(self) -> bool:
--- a/src/ormah/engine/prompt_classifier.py
+++ b/src/ormah/engine/prompt_classifier.py
@@ -5,11 +5,13 @@ from __future__ import annotations
 import logging
 import re
 from dataclasses import dataclass, field
-from datetime import datetime, timedelta, timezone
+from functools import lru_cache
 
 import numpy as np
 
+from ormah.config import Settings
 from ormah.embeddings.base import EmbeddingAdapter
+from ormah.engine.temporal import TemporalParser, resolve_locales
 
 logger = logging.getLogger(__name__)
 
@@ -74,71 +76,6 @@ ARCHETYPES: dict[str, list[str]] = {
     ],
 }
 
-# Maps time-reference keywords to (days_start, days_end | None).
-# days_start = how far back the window starts (created_after = now - days_start).
-# days_end   = how far back the window ends   (created_before = now - days_end).
-#              None means "now" (i.e. the window extends to the present).
-_TIME_KEYWORDS: list[tuple[re.Pattern, int, int | None]] = [
-    (re.compile(r"\btoday\b", re.IGNORECASE), 1, None),           # 24h ago → now
-    (re.compile(r"\bhoje\b", re.IGNORECASE), 1, None),            # 24h ago → now
-    (re.compile(r"\byesterday\b", re.IGNORECASE), 2, 1),          # 48h ago → 24h ago
-    (re.compile(r"\bontem\b", re.IGNORECASE), 2, 1),              # 48h ago → 24h ago
-    (re.compile(r"\blast\s+week\b", re.IGNORECASE), 14, 7),       # 14d ago → 7d ago
-    (re.compile(r"\bsemana\s+passada\b", re.IGNORECASE), 14, 7),  # 14d ago → 7d ago
-    (re.compile(r"\bthis\s+week\b", re.IGNORECASE), 7, None),     # 7d ago → now
-    (re.compile(r"\b(?:esta|essa|nesta|nessa)\s+semana\b", re.IGNORECASE), 7, None),
-    (re.compile(r"\blast\s+month\b", re.IGNORECASE), 60, 30),     # 60d ago → 30d ago
-    (re.compile(r"\bm[êe]s\s+passado\b", re.IGNORECASE), 60, 30),  # 60d ago → 30d ago
-    (re.compile(r"\brecently\b|\blately\b", re.IGNORECASE), 3, None),  # 3d ago → now
-    (re.compile(r"\brecentemente\b|\bultimamente\b", re.IGNORECASE), 3, None),  # 3d → now
-]
-
-_NUMERIC_TIME_RE = re.compile(
-    r"\b(?:last|past|[úu]ltim[oa]s?)\s+(\d+)\s+"
-    r"(hours?|days?|weeks?|months?|horas?|dias?|semanas?|meses|m[êe]s)\b",
-    re.IGNORECASE,
-)
-
-_UNIT_TO_DAYS: dict[str, float] = {"hour": 1 / 24, "day": 1, "week": 7, "month": 30}
-
-# PT-BR unit -> canonical English key, applied after the caller has lowercased
-# and stripped the plural "s". Without this, _UNIT_TO_DAYS.get(unit, 1) silently
-# falls back to 1 day, so "últimas 2 semanas" would mean 2 days, and the
-# rolling-window branch (which tests for "week"/"month") would never fire.
-_UNIT_ALIASES: dict[str, str] = {
-    "hora": "hour",
-    "dia": "day",
-    "semana": "week",
-    "mese": "month",  # "meses" -> "mese"
-    "mê": "month",  # "mês" -> "mê"
-    "me": "month",  # "mes" -> "me"
-}
-
-_DEFAULT_TEMPORAL_DAYS = 3
-
-# Regex to match all temporal phrases for stripping from search queries.
-# Combines _TIME_KEYWORDS patterns + _NUMERIC_TIME_RE into one list.
-_TEMPORAL_STRIP_PATTERNS: list[re.Pattern] = [
-    re.compile(r"\btoday\b", re.IGNORECASE),
-    re.compile(r"\bhoje\b", re.IGNORECASE),
-    re.compile(r"\byesterday\b", re.IGNORECASE),
-    re.compile(r"\bontem\b", re.IGNORECASE),
-    re.compile(r"\blast\s+week\b", re.IGNORECASE),
-    re.compile(r"\bsemana\s+passada\b", re.IGNORECASE),
-    re.compile(r"\bthis\s+week\b", re.IGNORECASE),
-    re.compile(r"\b(?:esta|essa|nesta|nessa)\s+semana\b", re.IGNORECASE),
-    re.compile(r"\blast\s+month\b", re.IGNORECASE),
-    re.compile(r"\bm[êe]s\s+passado\b", re.IGNORECASE),
-    re.compile(r"\brecently\b|\blately\b|\brecent\b", re.IGNORECASE),
-    re.compile(r"\brecentemente\b|\bultimamente\b", re.IGNORECASE),
-    _NUMERIC_TIME_RE,
-]
-
-# Dangling prepositions left after temporal phrase removal.
-_DANGLING_PREP_RE = re.compile(
-    r"\b(?:in|during|from|over|for)\s+(?:the\s+)?(?=\s*$|\s*,)", re.IGNORECASE
-)
-
 # Acknowledgements are distinct from conversational turns: when they follow a
 # useful answer, embedding similarity can otherwise label them as a
 # continuation and cause the prior question to be searched again. Keep this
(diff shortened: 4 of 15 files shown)


r-spade:
Thanks Andre — the PT-BR coverage and tests make sense, and I agree that #135 needs fixing.

Before we merge, could we shape this as locale support rather than extend the shared English regex tables? `_TIME_KEYWORDS`, `_TEMPORAL_STRIP_PATTERNS`, `_NUMERIC_TIME_RE`, and `_UNIT_ALIASES` would otherwise need each language added in multiple places.

I think the clean boundary is:

- the generic temporal parser owns matching, window calculation, and rolling-period behavior;
- a `TemporalLocale` data model supplies static phrases, numeric patterns/unit mappings, and post-strip cleanup (so PT-BR can also remove leftover `na` / `no` / `nos`);
- a registry loads built-in `en` and `pt-BR` packs. A setting can select enabled packs, e.g. `ORMAH_TEMPORAL_LOCALES=en,pt-BR`; grammar itself should live in tested locale packs, not raw user regex config.

Could you move the PT-BR rules and tests into that structure? It preserves this feature, makes a new language a self-contained locale module, and removes the need to keep duplicate canonical tables in sync.

AndreLFSMartins:
Reshaped as locale packs, as asked. The PT-BR grammar no longer touches the shared tables.

**Structure**

- `TemporalLocale` carries a language's grammar as data: static phrases (pattern, optional window,
  `priority`, and a `probe` used by the tests), the pack's own numeric pattern plus unit aliases,
  and its post-strip cleanup patterns.
- `TemporalParser` owns everything language-agnostic — match ordering, window arithmetic, the
  unit→days map, the rolling previous-period rule and the default window.
- A registry loads the built-in `en` and `pt-BR` packs; `ORMAH_TEMPORAL_LOCALES` (default
  `en,pt-BR`) selects them. Grammar lives in the packs, never in configuration.
- The keyword table and the strip table are derived from one declaration, so a phrase cannot be
  recognised for the window and forgotten for the strip.
- The three public functions keep their names, signatures and behaviour; `memory_engine` is
  untouched.

**Things worth your attention, rather than buried in the diff**

1. **This changes English behaviour in one case, and it is a bug fix.** Uppercase numeric units now
   normalise: `"what happened LAST 2 WEEKS"` used to yield a 2-day window, because `rstrip("s")` ran
   before `.lower()`, so `"WEEKS"` kept its plural and fell through the unit map to 1 day. It now
   yields the intended 28→14 day rolling window. Required for the PT alias map to work at all.
2. **Two tables disagreed today and one had to win.** `\brecent\b` is in the strip list but not the
   keyword list. It is declared strip-only: removed from the query, never selecting a window, never
   making `has_temporal_phrases` true. That last part matters — `memory_engine` gates `created_after`
   on that function, so promoting it would start date-filtering "show me recent changes".
3. **Cleanup is anchored to the span that was removed, with the owning pack's patterns**, and the
   end-of-residue/comma boundary is evaluated in the *final* residue. Running every pack's cleanup
   would turn `"please say no last 3 days"` into `"please say"`. The cost is one asymmetry:
   `"work from ontem"` returns `"work from"` where it used to return `"work"` — removing that `from`
   needs cross-pack cleanup, which is exactly what breaks the previous example. Pinned by a test
   with the reason, so it does not get "fixed" back.
4. **Removals stay a sequential re-search of the current residue.** Collecting spans against the
   original prompt and deleting them afterwards looks equivalent and is not: `semana passada` and
   `(esta|essa|nesta|nessa) semana` overlap on "semana", and right-to-left deletion turns
   `"nesta semana passada discutimos autenticação"` into `"timos autenticação"`.
5. **Each pack owns its numeric pattern** rather than sharing one. A shared pattern makes every pack
   claim every numeric match, which makes cleanup scoping untestable. One consequence: the
   cross-language accidents of the old combined regex (`"last 2 semanas"`, `"últimas 2 weeks"`) no
   longer match. No test asserted them.
6. **Match order is a property of the entries, not of pack registration.** Static entries merge by
   `priority` — today's `_TIME_KEYWORDS` interleaves the languages, so `ontem` sits before
   `last month` and `"compare ontem with last month"` picks the yesterday window; concatenating packs
   by setting order would flip it. The numeric winner is the leftmost match across packs, not the
   first pack that matches.
7. **Minor, disclosed rather than hidden:** cleanup no longer fires away from a vacated gap
   (`"what did we do last month over ,"` keeps its `over`), which needs a space before a comma to
   reach; `config.py` imports the locale registry so an unknown code fails at boot rather than at
   first use; `probe` and `registered_codes()` are consumed only by tests.

8. **Stripping went from linear to quadratic in the number of removals, and that is the price of
   per-pack cleanup.** `parser.py:127` re-offsets the whole gap list on every removal, so `_remove`
   is O(removals²). Measured end to end against this branch's merge-base, same output on both sides
   (identical SHA-1 of the residue in all three rows):

   | input | base | this branch |
   |---|---|---|
   | normal prompt, 3 temporal phrases | 4.7 µs | 9.6 µs |
   | 64 KB document, 300 temporal words | 3.45 ms | 7.13 ms |
   | 64 KB document, 4000 temporal words | 2.85 ms | 167 ms |

   Worth separating the two regimes, because they have different causes: at 300 removals line 127 is
   only 15% of the cost and the ~2× is just two packs doing more regex work; at 4000 removals it is
   92% and the curve is quadratic. The base is flat in removal count (`pat.sub("", result)`), so the
   growth is new.

   The gap list exists only so cleanup can run the owning pack's patterns at the span that was
   vacated — point 3. If that is not worth the complexity, the packs work with `cleanup_patterns=()`
   and the old global `_DANGLING_PREP_RE`: the same locale-pack shape you asked for, linear again,
   and the only behavioural loss is a dangling PT-BR preposition (`"mudanças na API nos"` instead of
   `"mudanças na API"`). Tell me which you prefer and I'll push that instead.

**Verification**

Full suite green (the four pre-existing failures on a clean `main` checkout are unrelated: three
`TestConfigureCodexMcp` and one `test_detect_returns_none_for_home`), ruff clean. Every PT-BR
assertion from the first pass moved across unchanged. Beyond that: a 201k-prompt differential of the
new parser against the current implementation found zero differences in detection and none in
English windows except the uppercase-unit fix above; the rules that are easy to get subtly wrong —
cleanup boundary, span anchoring, numeric leftmost, priority ordering, sequential removal — were each
checked by deliberately breaking them and confirming a test goes red.

r-spade:
Thanks for the detailed implementation and write-up, Andre. I took another pass through the locale internals. The locale-pack direction is right, and the implementation is well tested, but I think we may be preserving more historical and mixed-language behavior than the product needs.

Before doing another implementation pass, I’d like your opinion on a smaller design:

1. **Load locale modules from the configured codes.** Instead of importing and registering every built-in pack in `temporal/__init__.py`, validate each code, map something like `pt-BR` to `locales.pt_br`, import that module from the fixed locale package, and read its exported `LOCALE`. Adding a packaged language would then be self-contained and would not require editing the central registry.

2. **Enable only English by default.** PT-BR should remain bundled and fully tested, but users would opt into it with `ORMAH_TEMPORAL_LOCALES=en,pt-BR` or select it alone with `pt-BR`. Shipping a locale and enabling it globally feel like separate decisions.

3. **Choose the leftmost temporal phrase in the prompt.** Could we use the earliest match position, with a deterministic tie-break when two matches begin at the same place, instead of exposing the old shared-table order as cross-locale integer `priority` values? For example, `compare last month with ontem` would follow what the user wrote first. This would also let numeric and static matches use one understandable selection rule.

4. **Keep test probes in the tests.** `probe` and `registered_codes()` appear to exist mainly so tests can walk the registry. I think the test prompts can live in fixtures, leaving the production `StaticPhrase` and locale API limited to runtime grammar.

5. **Let strip patterns consume their own prepositions.** A locale phrase can optionally include its leading preposition in the match—for example, PT-BR can remove all of `na semana passada` or `nos últimos 3 dias`. That should remove the need for owner-aware gaps, repeated offset adjustment, and the O(removals²) cleanup path. The grammar becomes slightly more explicit, but each pack owns the complete text it removes.

My concern is less about current correctness—the implementation works and the test coverage is strong—and more about making the abstraction small enough that the third and fourth locales stay easy to add.

What do you think of this shape? I may be underweighting a real constraint behind the priority or owner-aware cleanup machinery, so please push back if those solve production cases that the simpler rules would mishandle.

AndreLFSMartins:
Thanks. I think this shape is better, and I'd go with all five points. Before replying I checked two things against real data, and one of them corrects something I wrote in the PR description.

**Real traffic doesn't settle priority vs leftmost**

You asked whether priority ordering and owner-aware cleanup solve real production cases. `retrieval_events` and `whisper_decisions` on my own install, from 2026-08-13 to 2026-09-14, hold 4,404 retrievals (4,147 whisper, 257 `recall_search`) and 2,538 distinct prompts. I ran all of them through this PR's parser and through a throwaway prototype of your design: leftmost match, longest match on a tie, each strip pattern with an optional leading preposition (`in|during|from|over|for (the)` for en, `na|no|nos|nas|em` for pt-BR), and no gap list.

Detection differed on 0 prompts and the stripped query on 0. The window differed on 2, which turn out to be the same pasted email sent twice. 64 prompts contain a temporal phrase, but none of them is a deliberate time question. They are pasted text, agent messages and "hoje" used to mean "currently", and all 64 were classified `general`. `recall_search` had no temporal phrase at all.

So the log neither supports nor refutes the machinery. It has no case where the two designs could disagree in a way that matters. My agreement rests on the constructed cases below, not on this count.

**Correction: PT-BR intent depends on the embedding model**

The PR description says the intent classifier already classifies PT-BR cross-lingually. That is only true with a multilingual encoder. I ran `PromptClassifier.classify` on this branch against eight direct PT-BR time questions ("o que fizemos ontem", "mudanças na API nos últimos 3 dias", and so on) at the default 0.65 threshold:

| encoder | PT-BR questions classified `temporal` |
|---|---|
| `bge-m3` (what my install runs) | 8 of 8 |
| `BAAI/bge-base-en-v1.5` (the default) | 1 of 8 |

On a default install, then, the whisper path never applies a PT-BR window, because `extract_time_params` only runs when intent includes `temporal`. `recall_search` is not gated on intent, so the PT-BR grammar does work there. The English archetypes are the gate, not the grammar.

**Point by point**

1. Agreed. One guard: check the code against a fixed shape before `importlib` and import only from the fixed locale package, so the setting can never name an arbitrary module. An unknown code should still fail at settings validation, as it does now.

2. Agreed, and the table above is a second reason. PT-BR only fully works with a multilingual embedding model, so opting in belongs next to that choice. I'd say so in the setting's docs. Adding PT-BR archetypes to the classifier feels like a separate issue to me, but tell me if you want it here.

3. Agreed, with the longest match breaking a tie at the same offset, then pack order. The one case I know where leftmost reads worse is "nesta semana passada" ("this past week"). Leftmost picks `nesta semana` (7d → now), while this PR picks `semana passada` (14d → 7d). The pack can fix it by declaring `(?:nesta|nessa) semana passada` as its own phrase: with the longest-match tie-break it wins and gives 14d → 7d. Also, numeric no longer outranks static, so "last week and last 2 weeks" now picks `last week`. I think that is the right reading.

4. Agreed. The probe prompts move into a test fixture.

5. Agreed. One constraint from my earlier comment survives: removal still re-searches the current residue, because `semana passada` and `(esta|essa|nesta|nessa) semana` overlap on "semana". That is plain `pattern.sub` in declaration order, so it stays linear with no gap list. The combined phrase from point 3 has to come first in that order, or "nesta semana passada" still leaves "nesta" behind, as it does in this PR today. It also fixes a mid-sentence case the current cleanup misses: "what did we do in the last 3 days on auth" becomes "what did we do on auth" instead of "what did we do in the on auth".

Cost on my side: the tests that pin priority, cleanup scoping, probes and the registry get rewritten or deleted, about 17 by test name. The PT-BR window assertions stay as they are.

If this matches what you had in mind, I'll rework the PR along these lines and fix the description.

r-spade:
Thanks for taking the time to test this so carefully, Andre — this is really helpful, and I appreciate the correction on the classifier behaviour. Yes, this matches what I had in mind, and I think the smaller shape will give us a much cleaner foundation for future locales.

Please go ahead with all five changes. The leftmost match with the longest-match tie-break, the explicit combined `nesta/nessa semana passada` phrase, and declaration-order substitution all sound good to me. The real-traffic comparison is reassuring too, especially with its limitations called out as clearly as you have.

For the PT-BR classifier finding, I agree that it is a separate concern rather than something to fold into this parser refactor. Could you update the PR description, and ideally leave a note on #135, to clarify that the original result was with `bge-m3`, while the default `BAAI/bge-base-en-v1.5` does not reliably recognise PT-BR temporal intent?

English-only by default feels right in light of that. It would be helpful for the setting docs to explain that enabling PT-BR grammar does not by itself make the whole whisper/search stack multilingual, and that a multilingual embedding model is recommended for PT-BR use.

I would also avoid adding unconditional PT-BR archetypes as a quick follow-up: if the PT-BR pack is disabled, they could still classify a prompt as temporal and leave the parser applying its generic fallback window. Locale-aware multilingual intent support feels worth treating as its own design problem.

Thanks again for digging into the real behaviour instead of just reworking the code mechanically. Please push the simpler version when ready and I’ll take another pass.

AndreLFSMartins:
Thanks. The simpler version is pushed, with all five changes, and I rewrote the description for the new shape. The description now says the PT-BR intent result depends on the encoder. I left the same note on #135.

One correction to my comment of 2026-09-15. I wrote that on a default install the whisper path never applies a PT-BR window. That was wrong. `build_whisper_context` calls `recall_search_structured` with automatic temporal parsing on, so with the PT-BR pack enabled the window and strip apply whatever the encoder. What the encoder controls is the whisper's temporal treatment: the relaxed score floor, no reranker or injection gate, and recency order. The conclusions stand: English-only by default, a multilingual encoder recommended for PT-BR, and no unconditional PT-BR archetypes.

Three things came up while I tested it that the five points did not cover. The description lists them under "Design" and "Behaviour changes in English":

- **Verb particles.** With a preposition in every strip pattern, "fail over last week" lost "over" and "log in yesterday" lost "in". `in`, `over` and `for` are now only stripped when `the` follows. `during` and `from` are stripped as before. "over" also joins the recall stop words, so "what did we work on over last week" still takes the recency path.
- **Leftmost and a narrower later phrase.** "what changed today in the last 2 hours" now gives 24h, because `today` comes first. I kept that as the documented rule rather than adding an exception for narrower windows. Tell me if you would rather have one.
- **Engine settings.** `MemoryEngine` and the classifier in `ContextBuilder` now build their parser from the `Settings` they hold. Before, they read the process environment, so a `Settings` with a different `temporal_locales` was silently ignored.

I also added `desta/dessa` next to `esta/essa/nesta/nessa` in the PT-BR pack.

Two gaps I did not close, because they belong to the multilingual-intent problem. The whisper's temporal treatment needs the `temporal` intent, which the default encoder rarely gives a PT-BR prompt. And recall decides whether a query is time-only with English stop words, so "o que fizemos ontem" gets the PT-BR window but not the recency path. The description says both. I can open a separate issue for them, if you want.

Ready for another pass.

r-spade:
Thanks so much for sticking with this, Andre. I know I asked you to revisit the design a few times, and I really appreciate your patience with that. The way you tested the ideas against real usage, pushed back where it mattered, and came back with a simpler implementation made this much easier to reason about. I learned a fair bit from the back-and-forth too.

The re-review looks good. I’ve made the small documentation correction about reranking still running, and I’m getting this merged. Really glad to have your help on Ormah — thank you 🙂

r-spade (approved):
Verified the five requested simplifications and the parser, settings, recall, and whisper paths. 437 focused tests and three additional verification checks passed, with Ruff clean. The final commit corrects documentation only. Approved for merge once fresh CI passes.

r-spade (approved):

r-spade:
@AndreLFSMartins Also, I believe I have fixed the permission issues. So you should be able to hit the merge pr button there. :)

PR #284, opened 2026-09-03
fix: make fresh-install recovery reachable

### LINKED ISSUE #251: Fresh install cannot reach recovery-kit import: restore needs the store id the kit provides
## Severity

Blocker for the new-machine scenario. Reproduced end-to-end on a wiped Mac with desktop 0.3.24 / ormah 0.14.11, signing in to an account that had cloud backups.

## What happens

Install the app on a machine with no prior Ormah state, sign in to an account that already has cloud backups. The Protection panel shows `local_only`, offers **Protect this memory**, and offers no route to the existing backups. The Cloud column reads **"unavailable"**.

The backend is not confused. `GET /admin/cloud/protection/remote` returns:

```json
{"snapshot_id": null, "created_at": null,
 "error": "Cloud store id is missing; import a recovery kit first."}
```

That message is correct and actionable. It never reaches the user.

## Why it cannot be escaped from the UI

The import control exists — **"Choose recovery kit"**, `ui/src/components/ProtectionPanel.tsx:992`. It is unreachable on a new machine:

1. The button renders only when `view === "restore_key"`, a step inside the restore flow.
2. Restore only appears when `restorable = Boolean(remote?.snapshot_id)` (`ui/src/productBridge.ts`, `protectionActions`).
3. `remote.snapshot_id` is null because `newest_cloud_snapshot` cannot list blobs without a store id (`src/ormah/cloud/restore.py:100`).
4. The store id comes from the recovery kit.

**To import the kit you must start a restore; to start a restore the app needs what the kit provides.**

The second door is shut too: `recoveryKitSectionVisible(status)` is `Boolean(status?.enabled)` (`ui/src/productBridge.ts:119`), so the recovery-kit section is hidden whenever protection is not already enabled — i.e. on every new machine.

## Why the store id is missing at all

`.store_id` lives in the memory directory. `ormah uninstall` deletes that directory while deliberately preserving `cloud.key` and `ormah-recovery-kit.md` in `~/.config/ormah`. The preserved kit *does* contain both the key and the store id, so the material is sufficient — `_cmd_cloud_init` even installs the store id before the key, with a comment naming this exact hazard:

> The kit's store id is the remote namespace — it must be installed too, or the restored machine would point at a brand new store and orphan every existing backup.

Recovery is fully implemented. It is only unreachable through the product.

## Consequence

Without a terminal, a returning user concludes their memories are gone. The visible next step is **Protect this memory**, which creates a new store id and orphans every existing backup — a data-loss-shaped trap in precisely the "I got a new laptop" scenario cloud backup exists to serve.

## Workaround

```
ormah cloud init --import-key ~/.config/ormah/ormah-recovery-kit.md
```
Then restart the daemon. Note this currently exits non-zero when a key already exists even though it installs the store id — see the companion issue.

## Suggested fix

The machinery all exists; this is a routing problem.

1. Surface `remote.error` instead of collapsing it to "unavailable" at `ProtectionPanel.tsx:1069`. The comment there already insists that failing to read the listing "must never be shown as" the cloud being empty — that principle is right and is not currently honoured for this case.
2. Offer a **Recover from a backup** entry point from `local_only` when `remote.error` indicates a missing store id, routing straight to the `restore_key` view.
3. Alternatively, make `recoveryKitSectionVisible` true when protection is disabled but a kit could establish identity.

## Tests worth adding

- `protectionActions` offers a recovery route for `local_only` when `remote.error` reports a missing store id.
- The `restore_key` view is reachable without a readable remote listing.
- A regression test for the loop itself: no state should require a snapshot in order to import the kit that makes snapshots visible.

r-spade:
## Problem

A signed-in fresh device has no local store identity, so remote discovery returns `key_missing` with no snapshot. The panel previously skipped discovery in that state and only exposed recovery-kit import from an unreachable restore flow; choosing Protect could instead create a new store.

## Flow

- `/admin/cloud/protection/remote` now returns the stable, redacted `reason_code` alongside its human explanation.
- A signed-in `local_only` device with `reason_code: key_missing` shows **Recover from a backup** beside **Protect this memory**.
- Recovery opens the existing kit chooser directly, without prepare/list first.
- After import, the panel refreshes status and remote discovery before the existing restore preparation flow. Cancelling returns to summary.

## Screenshots

Not included: this is a desktop-native file chooser flow and the UI changes are covered by component/action tests.

## Tests

- `npm --prefix ui test -- --run`
- `npm --prefix ui run build`
- `ORMAH_LLM_PROVIDER=none uv run pytest tests/test_api/test_protection_routes.py -v`
- `uv run ruff check src/ormah/api/routes_protection.py src/ormah/cloud/restore.py tests/test_api/test_protection_routes.py`
- `git diff --check`

`cargo test --manifest-path desktop/src-tauri/Cargo.toml product_bridge::tests` is blocked in this checkout because the required `desktop/src-tauri/binaries/uv-x86_64-unknown-linux-gnu` sidecar resource is absent.

Closes #251

### DIFF
--- a/desktop/src-tauri/src/product_bridge.rs
+++ b/desktop/src-tauri/src/product_bridge.rs
@@ -1062,7 +1062,8 @@ mod tests {
     fn protection_payload_rejects_secret_bearing_fields() {
         assert!(reject_forbidden_response_fields(&json!({
             "state": "protected",
-            "operation": { "id": "safe" }
+            "operation": { "id": "safe" },
+            "reason_code": "key_missing"
         }))
         .is_ok());
         for key in [
--- a/desktop/src-tauri/src/product_bridge.rs
+++ b/desktop/src-tauri/src/product_bridge.rs
@@ -1062,8 +1062,7 @@ mod tests {
     fn protection_payload_rejects_secret_bearing_fields() {
         assert!(reject_forbidden_response_fields(&json!({
             "state": "protected",
-            "operation": { "id": "safe" },
-            "reason_code": "key_missing"
+            "operation": { "id": "safe" }
         }))
         .is_ok());
         for key in [
--- a/src/ormah/api/routes_protection.py
+++ b/src/ormah/api/routes_protection.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 from pathlib import Path
 import uuid
 
@@ -24,11 +25,25 @@ from ormah.cloud.state import (
 )
 from ormah.cloud.store_lock import StoreLockTimeout
 
+logger = logging.getLogger(__name__)
+
 router = APIRouter(
     prefix="/admin/cloud/protection",
     tags=["cloud-protection"],
     dependencies=[Depends(require_local_admin)],
 )
+
+_REMOTE_LISTING_MESSAGES = {
+    "key_missing": "This device is not connected to a cloud memory store.",
+    "sign_in_required": "Sign in again to check cloud backups.",
+}
+
+
+def _remote_listing_message(reason_code: str) -> str:
+    return _REMOTE_LISTING_MESSAGES.get(
+        reason_code,
+        "Cloud backups could not be checked.",
+    )
 
 
 class EmptyRequest(BaseModel):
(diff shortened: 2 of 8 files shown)


r-spade:
Maintainer follow-up `1ec3600` rebases this PR onto merged #285 and addresses the agreed review findings:

- derives fresh-device choices from local `store_id === null` instead of an impossible pre-identity cloud listing
- presents **Protect this memory** and **Recover existing memory** as neutral primary/secondary choices for both new and returning users
- does not call remote discovery until a store identity exists
- maps stable reason codes to fixed product copy and never renders backend exception text
- keeps the recovery screen open when file selection is canceled
- separates successful import from refresh/preparation failures, avoiding contradictory success feedback
- renders the actual policy output in the component test and tests import → refresh → preparation outcomes
- exercises the real missing-store guard in the Python route test

Per maintainer decision, the suggested bridge-v2 compatibility guard was intentionally not added; supported released clients already use bridge v2+, and no legacy cloud user base exists.

Local verification: 57 UI tests passed, production UI build passed, Ruff/diff checks passed. The local FastAPI TestClient suite hit the known sandbox stall without a test failure; clean GitHub CI is the authoritative route-suite run.

PR #285, opened 2026-09-03
fix(cloud): make recovery imports idempotent

### LINKED ISSUE #252: ormah cloud init --import-key exits non-zero after successfully installing the store id
## What happens

On a machine that already has `cloud.key` but no `.store_id` (the state left by `ormah uninstall` + reinstall):

```
$ ormah cloud init --import-key ~/.config/ormah/ormah-recovery-kit.md
[xx] A cloud key already exists at ~/.config/ormah/cloud.key; refusing to overwrite.
     Move it aside first if you really mean to replace it.
```

Reads as a total failure. It is not — the store id **was** installed:

```
$ cat ~/.local/share/ormah/memory/.store_id
4792b0e3-31d2-4a82-a059-2c0ec4d67aac
```

## Why

`_cmd_cloud_init` (`src/ormah/cli.py`) installs the store id before importing key material, on purpose:

```python
# Installed first so a store-id conflict aborts before any key material is written.
imported_store_id = extract_store_id(import_source)
if imported_store_id:
    install_store_id(settings.memory_dir, imported_store_id)
import_key(import_source)
```

The ordering is correct and worth keeping. But when `import_key` then refuses because a key exists, the command reports failure and says nothing about the store id it just wrote.

## Why it matters

This is the documented workaround for #251 — the only way a returning user recovers their memories on a new machine. Hitting an error message that says "refusing to overwrite" makes it look like the recovery did not happen, so the natural next move is to abandon it, or to "move the key aside" and overwrite good key material for no reason.

The case is not exotic. It is exactly what `ormah uninstall` leaves behind: it preserves `cloud.key` and the kit while deleting the memory directory that held `.store_id`. Key present, store id absent, is the *normal* post-uninstall state.

## Suggested fix

Treat an already-present, matching key as success rather than conflict:

- If the kit's identities are already a subset of the existing `cloud.key`, there is nothing to write — report success and state what changed (`store id installed; key already present and matches`).
- Only refuse when the kit carries identities that genuinely conflict with the existing key.
- Either way, report the store-id outcome separately from the key outcome so a partial success is never rendered as a failure.

## Test

Given an existing `cloud.key` whose identities match the kit and a missing `.store_id`, importing the kit exits zero, installs the store id, and leaves the key byte-for-byte unchanged.

r-spade:
## Problem

`ormah cloud init --import-key` installed a recovery kit store ID before refusing a preserved matching `cloud.key`, leaving a partial mutation after normal uninstall/reinstall.

## Invariants

- Preflight reads/parses the source once and validates kit identities, store ID, existing store ID, and existing keyring before writes.
- Matching preserved keyrings (including extra rotated identities) are retained byte-for-byte.
- Store/key conflicts abort with no resource changes; new key material is never merged implicitly.
- Fresh keys still use atomic 0600 writes; genuine post-write filesystem failures report separate resource outcomes.

## Tests

- Preserved key + missing store, full idempotent no-op, rotated keyring, store/key conflicts, file/raw/stdin inputs, permissions, output statuses, and partial-write reporting.
- `ORMAH_LLM_PROVIDER=none uv run pytest tests/test_cloud_keys.py tests/test_cloud_cli.py -v`
- `uv run ruff check src/ormah/cloud/keys.py src/ormah/cli.py tests/test_cloud_keys.py tests/test_cloud_cli.py`
- `git diff --check`

Closes #252

### DIFF
--- a/docs/15 - Cloud Backup, Verification, and Restore.md
+++ b/docs/15 - Cloud Backup, Verification, and Restore.md
@@ -487,6 +487,19 @@ ormah server stop
 ormah backup restore --cloud --yes
 ```
 
+Recovery-kit import is preflighted before either local resource changes. Ormah reads the kit once,
+validates its store ID and every identity, then compares them with any installed `.store_id` and
+`cloud.key`. The normal uninstall/reinstall state has a preserved matching key but no store ID; in
+that case Ormah installs only the missing store ID, leaves the key file byte-for-byte unchanged,
+regenerates the canonical recovery kit from the complete installed keyring, and exits successfully.
+Repeating an already-complete import is also a successful no-op. A different store ID or an identity
+that is not already present in an existing keyring fails before either resource is written.
+
+Human and JSON output report the store and key outcomes independently. JSON callers receive
+`store_id_status` and `key_status`; neither output includes private identity material. Existing
+keyrings with additional retained rotation identities are valid, and the refreshed recovery kit
+contains the full current-first keyring rather than preserving a stale or unrelated kit file.
+
 The restore command:
 
 1. reads the imported store ID instead of generating a new remote namespace;
--- a/src/ormah/cli.py
+++ b/src/ormah/cli.py
@@ -515,7 +515,6 @@ def _cmd_cloud_init(args):
     from ormah.cloud.crypto import CloudCryptoError
     from ormah.cloud.keys import (
         KEY_PATH,
-        RECOVERY_KIT_PATH,
         MAX_RECOVERY_KIT_BYTES,
         CloudKeyError,
         apply_recovery_import,
@@ -523,6 +522,7 @@ def _cmd_cloud_init(args):
         init_key,
         load_identity_strings,
         plan_recovery_import,
+        validate_recovery_kit_destination,
         write_recovery_kit,
     )
     from ormah.config import settings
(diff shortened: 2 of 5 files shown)


r-spade:
Maintainer follow-up `ac88d47` addresses the adversarial review findings:

- removes blind canonical-kit preservation and always regenerates it from the authoritative installed keyring
- preflights a newer-format kit destination before changing key/store state
- fixes `already_present` and zero-mutation filesystem-failure reporting
- consolidates path/text reading and guards source-read-once
- rejects duplicate/malformed identities, symlinked resources, and group/world-readable adopted keys
- writes imported store IDs atomically with owner-only permissions
- restores human identity-count feedback and adds safe conflict remediation
- adds direct plan/apply and missing fail-closed tests plus recovery documentation

Verification: 70 focused tests passed; 247 setup tests passed; repository-wide Ruff and diff checks passed. The broader cloud suite was 350 passed with one `.env.lock` failure reproduced unchanged on base `b38311a`.

r-spade:
Final maintainer safety pass `fa19163` prevents the opposite failure mode: canonical-kit refresh now checks that an existing kit belongs to the planned store and contains only identities explained by the resulting installed keyring. A stale subset is safely refreshed; unrelated store/key recovery material fails before writes; an empty damaged kit self-heals. Bare-key import creates a new store only when no canonical kit exists.

Final local verification: 73 focused tests passed, 247 setup tests passed, all Ruff/diff checks passed, and the broader cloud suite was 354 passed plus the one base-reproduced `.env.lock` failure.

PR #299, opened 2026-09-06
fix(setup): plugin mode ships one maintenance surface — remove the CLI agent and command, bind the plugin agent to its own tool

AndreLFSMartins:
## Problem

On an install that has the Claude Code plugin, `ormah setup` leaves two maintenance surfaces that both call `mcp__ormah__run_maintenance` — the tool name of the CLI-registered `ormah` MCP server:

1. `~/.claude/agents/ormah-maintenance.md` and `~/.claude/commands/ormah-maintenance.md`. The plugin branch of `_claude_code_wire()` keeps installing them while removing that very server in the same run. #147 kept them on the premise that the plugin namespaces them, so they are not duplicate registrations. The namespacing is real, but the kept agent names a server that no longer exists.
2. `integrations/claude-plugin/agents/ormah-maintenance.md`, the plugin's own agent, although the plugin's server is exposed as `mcp__plugin_ormah_ormah__*` (the plugin's `.mcp.json` names it `ormah`, and Claude Code scopes plugin servers as `plugin_<plugin>_<server>`).

Claude Code resolves `~/.claude/agents/` ahead of plugin agents, so the stale CLI copy shadows the plugin's: `/ormah:maintenance` dispatches `subagent_type="ormah-maintenance"` and lands on an agent whose named tool does not exist. Neither agent declares `tools:`, so the model recovers the live name through ToolSearch and the defect stays invisible — maintenance kept running in every transcript I checked, through the wrong file, with the whole parent toolset.

## Change

Each install channel ships exactly one maintenance surface.

- **Plugin present:** `_claude_code_wire()` removes the CLI agent and slash command alongside the CLI hooks and MCP entry. What remains is the plugin's `/ormah:maintenance` and `ormah:ormah-maintenance`. `_claude_code_is_wired()` never reads the agent file, so the removal cannot cause a re-wire loop.
- **No plugin:** unchanged. The CLI server, `~/.claude/agents/ormah-maintenance.md` and `/ormah-maintenance` are installed as before, and `src/ormah/agents/ormah-maintenance.md` keeps `mcp__ormah__run_maintenance`, which is the right name for that channel.
- **Plugin agent:** calls `mcp__plugin_ormah_ormah__run_maintenance` and declares it as its only tool — the scoped form the plugins reference documents for plugin-provided MCP servers in a subagent's `tools:`.

## Evidence (Claude Code 2.1.263)

- A subagent declared with `tools: [mcp__plugin_ormah_ormah__run_maintenance]` and asked to list its tools without calling any replied with exactly that one name.
- With the user-level agent and command moved out of the way, `Agent(subagent_type="ormah-maintenance")` — bare, the way `/ormah:maintenance` calls it — dispatched the plugin agent, which answered `PONG` with `tool_uses: 0`. Removing the CLI copies does not break the plugin command.

## Tests

- `test_working_plugin_strips_hooks_and_mcp_and_writes_no_wiring` now seeds the two stale CLI files and asserts they are gone after wiring. Fails on `main`.
- `test_maintenance_agent_binds_the_plugin_scoped_tool_only` checks the plugin agent's frontmatter and body. Fails on `main`.
- `test_cli_channel_agent_keeps_the_cli_tool_name` guards the deliberate difference between the two agent files, so nobody "syncs" them later.
- Suite: 2042 passed. The 4 failures are environment-only and identical on pristine `main` in this environment (`TestConfigureCodexMcp` sees a real `codex` binary on PATH; `test_detect_returns_none_for_home` sees a temporary `HOME`). `ruff check src/ tests/` is clean.

### DIFF
--- a/integrations/claude-plugin/agents/ormah-maintenance.md
+++ b/integrations/claude-plugin/agents/ormah-maintenance.md
@@ -2,20 +2,21 @@
 name: ormah-maintenance
 description: Ormah memory graph maintenance. Runs the run_maintenance two-call protocol to link memories, detect conflicts, merge duplicates, and consolidate clusters.
 model: sonnet
+tools: mcp__plugin_ormah_ormah__run_maintenance
 ---
 
 You are the ormah memory maintenance agent. Your only job is to run the two-call
-run_maintenance protocol using the mcp__ormah__run_maintenance tool.
+run_maintenance protocol using the mcp__plugin_ormah_ormah__run_maintenance tool.
 
 ## Protocol
 
-**Phase 1** — call `mcp__ormah__run_maintenance` with no arguments. You will receive four batches:
+**Phase 1** — call `mcp__plugin_ormah_ormah__run_maintenance` with no arguments. You will receive four batches:
 - `link_candidates`: pairs to classify with a relationship
 - `conflict_candidates`: belief pairs to check for contradictions or evolution
 - `merge_candidates`: near-duplicate pairs to merge
 - `consolidation_clusters`: groups of similar memories to synthesize into one
 
-**Phase 2** — analyze all batches, then call `mcp__ormah__run_maintenance` again with a `results` object:
+**Phase 2** — analyze all batches, then call `mcp__plugin_ormah_ormah__run_maintenance` again with a `results` object:
 
 ```json
 {
--- a/src/ormah/setup.py
+++ b/src/ormah/setup.py
@@ -2824,25 +2824,29 @@ def _claude_code_wire() -> None:
     # The plugin registers the same UserPromptSubmit/PreCompact/SessionEnd hooks
     # and the same MCP server. Wiring them again in ~/.claude/settings.json runs
     # both copies: the whisper fires twice per human turn, and no merge can dedupe
-    # across the two files. The agent and slash command are namespaced by the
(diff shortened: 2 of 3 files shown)


AndreLFSMartins:
Second commit, 627a9ae, closes a gap an independent review of 5e36a39 found: nothing asserted that the non-plugin branch of `_claude_code_wire()` still installs the CLI agent and slash command — deleting those two calls left the suite green. `test_enabled_but_uninstalled_plugin_wires_normally` now captures both mocks and asserts each was called once. Verified by mutation: with the two lines removed it fails on `install_agents.assert_called_once()` (called 0 times); restored, the class passes.

One more probe on the frontmatter form of `tools:`, since the earlier probe used the `--agents` JSON form. A project-level agent file with exactly `tools: mcp__plugin_ormah_ormah__run_maintenance` spawned and listed that one tool. When the tool is unavailable to the session (I first ran the probe with that tool under a deny rule), the spawn refuses loudly: `Agent 'probe-fm' would be spawned with zero tools — refusing. Its tools list resolved to nothing: unrecognized [mcp__plugin_ormah_ormah__run_maintenance]`. So a session where the plugin MCP server is not connected gets a named error instead of a silent no-op — which is the right failure for a maintenance run that could not have reached the store anyway.

r-spade (approved):
LGTM

Issue #280, opened 2026-09-01
bug(store): 8-character UUID lookup can return, overwrite, or delete the wrong node


r-spade:
## Problem

Memory files end with MemoryNode.short_id, the first UUID group (8 hexadecimal characters / 32 bits):

```text
fact_example_deadbeef.md
```

FileStore._find_file() extracts that prefix, globs `*_<short_id>.md`, and immediately returns `matches[0]` without parsing the file and confirming that its full frontmatter ID equals the requested ID.

FileStore._path_for() also trusts that lookup when deciding whether to reuse an existing path.

## Reproduction

Using two controlled IDs with the same first group:

```text
requested: deadbeef-0000-4000-8000-000000000001
live node: deadbeef-0000-4000-8000-000000000002
```

1. Save the first node.
2. Soft-delete it.
3. Save the unrelated second node.
4. Call `FileStore.load()` with the first full ID.

Observed:

```text
requested=deadbeef-...0001
returned=deadbeef-...0002
wrong_node=True
```

If two colliding nodes are live when the second is saved, `_path_for()` can treat the first file as belonging to the second and overwrite/reuse it.

## Blast radius

This is system-wide because load, save, delete, soft_delete, and touch_access all rely on `_find_file()`. Higher-level engine, lifecycle, merge, backup/restore, watcher, and sync paths inherit those operations.

Possible outcomes include:

- reading or mutating the wrong memory;
- deleting the wrong memory;
- silently overwriting a live memory;
- making lifecycle decisions using an unrelated node.

The bug was discovered while reviewing PR #257: its `superseded_by` liveness check can mistake an unrelated same-prefix node for a deleted consolidation replacement and keep the source archived.

## Scale

For uniformly generated UUIDs, the birthday-bound probability of at least one shared 32-bit prefix is approximately:

- 1,843 nodes: 0.04%
- 9,300 nodes: 1%
- 36,729 nodes: 14.5%
- 77,000 nodes: 50%

The exact failure scenario depends on operation order, but collision probability increases rapidly with store history.

## Root cause

The short-ID filename/glob optimization existed in the original FileStore commit. It provides readable filenames and a fast lookup before the full ID cache is built, but assumes the 32-bit prefix is unique and never verifies the full ID stored in frontmatter.

## Suggested fix

- Never return or cache a glob match until its parsed full node ID exactly matches the requested ID.
- Never reuse an existing path in `_path_for()` unless the file contains the same full ID.
[...]

PR #301, opened 2026-09-09
fix(store): a Node reference resolves only when the node's Full id confirms it

AndreLFSMartins:
## Problem

`FileStore._find_file` extracted the 8-character Short id from a reference, globbed `*_<short_id>.md` and returned `matches[0]` without opening the file to confirm that its frontmatter id equals the requested one. Two nodes sharing a 32-bit prefix therefore resolved to each other, and since load, save, delete, soft_delete and touch_access all go through that lookup, any of them could act on a stranger's node. The unverified path was also written into the id cache, so the wrong answer outlived the call and persisted until restart. Reported in #280.

## What changes

- `50ff0599` The glob narrows but never decides: each candidate is parsed and accepted only when its Full id confirms the reference (a Full id must match exactly; a bare Short id must equal the candidate's `short_id`). An ambiguous Short id, one prefix with two live nodes, resolves to `None` with a warning-level log. The id cache is keyed by Full id only, written only after that confirmation, and never on a miss. A cache hit stays validated by file existence alone.
- `dc663800` Cache eviction on delete and soft_delete goes by path (`_forget(path)`), so it no longer depends on reading the file about to be removed. Previously a swallowed read error left a stale entry pointing at a path that `_path_for` would deterministically hand to the next node with the same type, title and Short id.
- `e734ac5f` An `OSError` while confirming a candidate propagates instead of resolving as a confirmed absence: a caller that reads `None` as absence (for example `MemoryEngine.delete_node`) would go on to mutate index state the file still contradicts. An unparseable file still confirms nothing and is skipped, matching what `list_all` and `_build_cache` already do.

## Design decisions, open to review

- The accepted prefix width stays at 8. Filenames end with the 8-character Short id, so any other width would force a directory scan per lookup.
- Ambiguity returns "not found" plus a warning rather than raising. The load contract is nullable and the background jobs branch on `None`, so raising would trade silent corruption for a crashed sleep cycle.
- No public signature changes.

## Tests

- `tests/test_store/test_prefix_resolution.py` (11 tests): colliding pairs across load, save, delete, soft_delete and cache rebuild; a Short id that stops resolving once a collider is born and resolves again once only one node is left; the ambiguous-id warning; the accepted existence-only cache hit.
- `tests/test_engine/test_short_id_reference.py`: an update addressed by a Short id, end to end through the engine.
- One adjustment in `tests/test_engine/test_reversible_promotion.py`.
- Both halves of the third commit are pinned: reverting the `raise` fails the read-failure test; widening it to every exception fails the corrupt-file test.

## Scope against #280

Covers bullet 1 of the suggested fix (never return or cache a glob match until the parsed Full id matches) and regression tests for load, save, delete, soft_delete and cache rebuild.

Does not cover bullet 2 (`_path_for` reuse), bullets 3 and 4 (collision-safe filenames and deterministic behaviour for legacy files, both of which need a decision about existing vaults), or the reindex, restore and sync/import parts of bullet 5. Hence Refs rather than a closing keyword.

## Verification on this branch

Suite run from the branch's own venv: 2076 passed, 4 failed, 1 deselected. The 4 failures reproduce on pristine `main` (`9d8c48c3`) on macOS and are untouched by this PR:

- `tests/test_adapters/test_space_detect.py::test_detect_returns_none_for_home` (`/var` vs `/private/var` realpath under a temporary `HOME`)
- `tests/test_setup.py::TestConfigureCodexMcp` x3 (`_find_binary` finds a real `/opt/homebrew/bin/codex` despite the `shutil.which` patch)

`ruff check src/ tests/` clean.

Refs #280
Fork ticket: AndreLFSMartins/ormah#29

### DIFF
--- a/src/ormah/store/file_store.py
+++ b/src/ormah/store/file_store.py
@@ -31,16 +31,18 @@ def _serialized_store_operation(method):
 class FileStore:
     """Manages memory node files on disk.
 
-    Maintains an in-memory ``id → Path`` cache so that lookups are O(1)
+    Maintains an in-memory ``Full id → Path`` cache so that lookups are O(1)
     after the first scan, instead of falling back to an O(N) grep over
-    every markdown file.
+    every markdown file. Only a Full id is ever a key, and only once the file
+    at that path has confirmed it: a Short id lookup re-resolves on each call,
+    and a miss caches nothing.
     """
 
     def __init__(self, nodes_dir: Path, operation_lock=None) -> None:
         self.nodes_dir = nodes_dir
         self._operation_lock = operation_lock or threading.RLock()
         self.nodes_dir.mkdir(parents=True, exist_ok=True)
-        # id -> Path cache, populated lazily on first miss
+        # Full id -> Path cache, populated lazily on first miss
         self._id_cache: dict[str, Path] = {}
         self._cache_built = False
 
@@ -90,8 +92,8 @@ class FileStore:
         path = self._find_file(node_id)
         if path is None:
             return False
+        self._forget(path)  # while the file is still there to name its Full id
         path.unlink()
-        self._id_cache.pop(node_id, None)
         return True
 
     @_serialized_store_operation
@@ -116,8 +118,8 @@ class FileStore:
         deleted_dir = self.nodes_dir.parent / "deleted"
         deleted_dir.mkdir(parents=True, exist_ok=True)
         dest = deleted_dir / path.name
+        self._forget(path)  # while the file is still there to name its Full id
         path.rename(dest)
-        self._id_cache.pop(node_id, None)
         return True
 
     @_serialized_store_operation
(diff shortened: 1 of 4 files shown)


r-spade:
Thanks Andre — the full-ID verification, path-based cache eviction and propagation of read errors are useful improvements. I agree with keeping #280 open for the wider work.

I tested the current head (`e734ac5`) with two different UUIDs sharing `deadbeef`, the same type and the same title. `_path_for()` still generates the same filename for both. Saving B overwrites A, and the warm cache then makes `load(A_full_uuid)` return B. This happens through normal `FileStore.save()`, without an external file replacement.

Before merging, I think we should close that overwrite path and ensure unresolved references cannot lead callers into a successful mutation:

- Give new files collision-safe names and never replace an existing destination belonging to another full ID. Existing files can keep their names and be located through their stored full UUID.
- Distinguish ambiguity/read failure from confirmed absence and handle that outcome in callers. In particular, `delete_node` can currently fall back from the store's ambiguous `None` to `GraphIndex.get_node`, whose `LIKE ... LIMIT 1` picks a collider, then report success without deleting it. A unique short-ID delete also needs to use the resolved full UUID for index removal and audit.
- Add regressions for same-prefix/type/title saves, full-ID reads with a warm cache and after restart, and unique/ambiguous short-ID deletion with file/index consistency checks. A malformed matching file must not be treated as permission to mutate index state either.

For the broader design, the direction we're leaning toward is full UUIDs in agent-facing output and internal operations, with optional short prefixes accepted only when uniquely resolvable. Existing short filenames don't need a bulk rename: a full-ID-to-path lookup can support them. We'll track the shared resolver, remaining callers, agent output, restore/import coverage and lookup performance under #280.

Could you extend this PR with the storage and deletion safety fixes above? We can keep the wider agent-output change separate and merge this once those safety cases pass.

AndreLFSMartins:
Thanks for the review, and for the repro. You were right: the save path could still overwrite a colliding node, and `delete_node` could still act on the wrong one.

I extended the PR for your three points. The description now covers each change.

- **Collision-safe names.** `_path_for` widens the filename with more of the UUID when the path is taken, and `_publish` claims a new name with `os.link`. A save never replaces a file that holds another full ID. Existing files keep their names.
- **Ambiguity is not absence.** `delete_node` resolves the reference first, then uses the resolved full UUID for the Self guard, the index removal and the audit entry. An ambiguous or unreadable reference deletes nothing. The index fallback accepts only an exact full-ID match, and a malformed file never authorises a mutation.
- **Regressions** for same prefix/type/title saves, full-ID reads with a warm cache and after a restart, and unique, ambiguous, too-short, Self and malformed-file deletes.

About the force-push: the three commits you reviewed keep the same code. I only rewrote their messages. `git diff e734ac5f b5f1c739` is empty, and `git range-diff e734ac5f...d74b5616` shows it. The eleven commits after them are new.

Two known limits remain, both listed in the description. They need a second process writing the same nodes directory. I think both belong with the wider work in #280, and I will write them up there.

AndreLFSMartins:
Pushed `cbc72d55` for the deleted-storage case from your tracker. Before it, `soft_delete` renamed the file into `deleted/` under its live name without checking the destination. So when two colliding nodes were deleted one after the other, the second tombstone replaced the first. Now the move claims its name with `os.link`, like a save does, and widens past a tombstone that belongs to another node. A new regression test covers the two-node sequence.

r-spade (approved):
Thanks Andre, really appreciate the care you've put into this and the time you've spent working through the review. Following the collision all the way into the deleted files was a good catch, and the extra tests and clear explanation of the remaining limits made this much easier to review.

I've now run the full test suite, the real-engine integration test and some extra collision cases through creation, updates, deletion, merge/undo and backup/restore. The fixes we asked for are holding up well, so I'm happy to merge this.

I've updated #280 with what's covered here and the remaining work, including the shared-store cache and write races, consistent full-ID handling and the compatibility checks. It's marked urgent and important, and we'll keep it open after this merges. When you get a chance, could you have a look at that checklist and let me know if I've missed anything or if you'd approach any of it differently?

Thanks again for sticking with this — it's an important improvement to a part of Ormah we really need to be able to trust.

Issue #306, opened 2026-09-19
Publish LongMemEval-S and LoCoMo benchmark results for Ormah


r-spade:
## Goal

Publish honest, reproducible Ormah numbers on the two standard long-term memory benchmarks, LongMemEval-S and LoCoMo, so Ormah can be compared with Mem0, Synap and others.

Today Ormah only has private, local-only whisper and recall evals. Competitors publish headline numbers (Mem0 reports 92.5 on LoCoMo and 94.4 on LongMemEval on its managed platform; Synap reports 93.2 and 92 without raw artifacts).

## What a published result must include

- Reproducible config and exact commands
- Raw per-question artifacts: question, gold answer, retrieved memories, answer, judge verdict
- Methodology and differences from Mem0 and the official LongMemEval setup
- Model versions, token usage, cost and latency
- Honest limitations

## Plan

Design: [`eval/bench/DESIGN.md`](https://github.com/r-spade/ormah/blob/feat/memory-benchmarks/eval/bench/DESIGN.md)
Agent brief used for implementation: [`eval/bench/AGENT_BRIEF.md`](https://github.com/r-spade/ormah/blob/feat/memory-benchmarks/eval/bench/AGENT_BRIEF.md)

- [x] Harness: dataset loaders, raw and extract stores, retrieval, answer and judge providers, metrics, reports, resumable runs
- [x] Free retrieval-only smoke runs (LoCoMo full, LongMemEval first 50)
- [ ] Fix subscription CLI provider auth (`claude -p --bare` disables OAuth login)
- [ ] 10-question LLM smoke run per dataset
- [ ] Pin answer, extraction and judge model ids
- [ ] Full LoCoMo run, extract mode, answer and judge
- [ ] Full LongMemEval-S run, raw mode, all 500 questions and all six question types
- [ ] Publish results and methodology

## Budget

First published run should need no API credits. Extraction and answering use `claude -p` on the Claude subscription, judging uses `codex exec` on the Codex subscription. The Anthropic API provider is optional and capped with `--max-usd`.

LongMemEval-S in extract mode is out of scope for the first run: about 79M input tokens across 19,829 unique sessions.

PR #307, opened 2026-09-19
feat(bench): public LongMemEval/LoCoMo benchmark harness

r-spade:
Part of #306.

**Draft, not ready to merge.** The harness works end to end for the free retrieval phases. The LLM answer and judge phases have not produced a single scored answer yet because of the auth blocker below.

Design: `eval/bench/DESIGN.md`. The exact brief given to the implementation agent is in `eval/bench/AGENT_BRIEF.md` for review.

## Summary

Adds a source-checkout public benchmark harness for LongMemEval-S and LoCoMo. Raw store/retrieve/report runs are free by default; extraction, answering and judging require explicit phases. The implementation follows `eval/bench/DESIGN.md`, with subscription CLI providers and an optional, budget-gated Anthropic provider tested exclusively with fakes.

The harness streams LongMemEval haystacks, caches extraction and embeddings, journals per-question progress for resumption, preserves historical session dates without FSRS expiry, and reports retrieval metrics, judge coverage, latency, token usage and API costs. Answer prompts adapt Mem0; judge prompts use the upstream dataset rules. Dataset content and raw artifacts stay ignored; checked-in results are explicitly smoke validation summaries.

Real validation uncovered a SQLite/sqlite-vec compatibility problem that silently forced lexical fallback. Using the explicit KNN `k` constraint restores vector retrieval on this node's SQLite 3.40.1. Regression tests cover this and structured-search tag hydration. Five existing setup tests now mock binary discovery completely so installed CLIs do not invalidate their no-CLI fixtures.

Smoke validation: full LoCoMo raw retrieval recall@30 **0.738223** on 1,982 labeled questions (1,986 total), p50/p95 **94.5/138.0 ms**, command wall **15.03 min**. LongMemEval's first 50 questions have session recall@30 **0.98**, nDCG@30 **0.900266**, p50/p95 **79.7/116.3 ms**, and **78.41 min** completed phase wall; all 50 are single-session-user. Linear 500-question phase projection is **13.07 h**, with cache/resume/sample caveats documented in results.

Both requested ten-question subscription smokes were attempted but scored **0/10** due to `claude --bare` authentication failure; accuracy is **N/A**. LoCoMo made six failed extract attempts; LongMemEval made ten failed answer attempts; no judge inference ran and API spend was **$0**. These are explicitly smoke summaries, not published accuracy claims.

Validation passed: `uv run pytest -q` — **2,120 passed, 1 integration test deselected**; `uv run pytest tests/test_eval_bench -q` — **47 passed**; existing recall/whisper evals — **135 passed, 1 deselected**; required Ruff command — **clean**. No log-directory override was needed. See `eval/bench/results/README.md` for commands, category/type tables, caveats and token projections.

Open decisions before full LLM runs:
- Installed Claude Code 2.1.220 disables subscription OAuth with the required `--bare` flag. The smoke attempts fail with `Not logged in`; an owner-approved subscription-compatible invocation is needed. No paid API was called.
- Pin concrete answer/extraction/judge model IDs before publishing. The node's configured Codex model is gpt-5.6-sol; the requested Claude alias is sonnet.
- Exact LoCoMo turn recall is unavailable for extracted facts because the shipped extractor does not return source turn IDs. Raw turn recall remains measurable.
- The first 50 LongMemEval questions are all single-session-user and cannot stand in for the other five question types.

## Notes from the orchestrator

- The agent's sandbox had a read-only `.git`, so it could not commit. The work was committed and pushed from the node afterwards as 030baff, unchanged.
- The `--bare` auth failure is confirmed outside the sandbox: `claude -p --bare` returns an error with no login, and the same call without `--bare` succeeds on the subscription. The fix is to drop `--bare` in `eval/bench/providers.py` and disable tools through the tools flags instead.
- Two production changes need review on their own: `VectorStore.search` now uses the explicit sqlite-vec KNN `k` constraint, and `llm_client` gains a `set_adapter()` helper. The first one may affect normal recall on SQLite 3.40, so it deserves its own check.
- Five setup tests were changed to patch binary discovery. Worth confirming they still test what they meant to.


r-spade:
## LLM smoke runs now pass (run on the Mac after 16d6457)

The `--bare` blocker is fixed. `claude -p` now runs with `--safe-mode`, which keeps subscription login but still disables CLAUDE.md, hooks, plugins, skills and MCP, so the owner's own Ormah memories cannot reach benchmark answers. Verified directly: the isolated call reports no Ormah context.

Both 10-question smokes ran with zero API spend. Sonnet 5 answered through `claude -p` and GPT-5.6 Terra judged through `codex exec`.

| Run | Questions | Accuracy | Errors |
|---|---|---|---|
| LongMemEval-S, raw, first 10 (all single-session-user) | 10 | 10/10 | 0 |
| LoCoMo, extract mode, conversation 0, first 10 | 10 | 9/10 | 0 |

The LoCoMo miss is a real error, not a harness bug: the answer placed a picnic in the week before 6 July instead of the week before 9 June.

Measured per-call latency:

| Phase | Provider | Median | Max |
|---|---|---|---|
| Extract, per session | claude-cli / Sonnet 5 | 13.0 s | 16.4 s |
| Answer | claude-cli / Sonnet 5 | 2.5 to 2.7 s | 5.5 s |
| Judge | codex / GPT-5.6 Terra | 4.5 s | 6.8 s |

Notes for the full runs:
- Codex judge calls carry about 14k to 17k input tokens each, mostly Codex's own overhead. That draws on the Codex subscription, not credits.
- Claude Code adds the account email as context in safe mode. It is irrelevant to answers but is worth listing as a limitation.
- These are smoke checks on 10 questions each and are not publishable accuracy numbers.

